PR #1606, opened 2014-12-09
Adding altitude component to the Geospatial Coordinate Encoder

### LINKED ISSUE #1461: GeospatialCoordinateEncoder should allow optional altitude
### Introduction to the GeospatialCoordinateEncoder

[See a video of @chetan51 describing the encoder here.](https://www.youtube.com/watch?v=KxxHo-FtKRo)

The [GeospatialCoordinateEncoder](https://github.com/numenta/nupic/blob/master/nupic/encoders/geospatial_coordinate.py#L36) (GCE) is a subclass of the [CoordinateEncoder](https://github.com/numenta/nupic/blob/master/nupic/encoders/coordinate.py#L29). Given a coordinate in an N-dimensional space, and a radius around that coordinate, the Coordinate Encoder returns an SDR representation of that position.

The Coordinate Encoder uses an N-dimensional integer coordinate space. For example, a valid coordinate in this space is (150, -49, 58), whereas an invalid coordinate would be (55.4, -5, 85.8475).

It uses the following algorithm:

1. Find all the coordinates around the input coordinate, within the specified radius.
2. For each coordinate, use a uniform hash function to deterministically map it to a real number between 0 and 1. This is the "order" of the coordinate.
3. Of these coordinates, pick the top W by order, where W is the number of active bits desired in the SDR.
4. For each of these W coordinates, use a uniform hash function to deterministically map it to one of the bits in the SDR. Make this bit active.
5. This results in a final SDR with exactly W bits active (barring chance hash collisions).

The GCE is given latitude, longitude, and speed. It calculates the radius for the CoordinateEncoder based upon the speed. Altitude is currently not a part of the GCE. 

### The Enhancement

This task is to add an optional altitude component to the GCE.

If the altitude is specified, the GCE should use a 3D coordinate when interfacing with the CoordinateEncoder. If the altitude is not specified, it should use a 2D coordinate. Ensure that the radius calculation based on speed of the CoordinateEncoder works in the 3D case (with altitude specified), and that it continues to adhere to the intention of `radiusForSpeed`, specifically in the amount of overlap it tries to achieve (as documented in the function currently).

### Tasks to be completed

- [ ] Capture NuPIC output from current GCE example code from sample route given in the [Geospatial Tracking App](https://github.com/numenta/nupic.geospatial). You'll want to save the anomaly scores for each point to compare results after you've made changes. 
- [ ] Before changing any production code, create unit tests that test input/output to the GCE, based upon the values you retrieved from the sample app above. Ensure they pass without changing the GCE.
- [ ] Update the GCE input tuple. Move speed to position 0, and add optional altitude parameter to the end. If tuple does not contain 4th parameter for altitude, it should use a 2D coordinate instead of 3D. Update all consumers of the GCE to adhere to the new format. Update documentation.
- [ ] If necessary, update the radius calculation to work for the 3D case as well.
- [ ] Ensure unit tests continue to pass.
- [ ] Create new unit tests to test that radius properly increases and decreases with changes in speed, with altitude specified. 
- [ ] Ensure new unit tests pass.
- [ ] Test end-to-end with the [Geospatial Tracking App](https://github.com/numenta/nupic.geospatial) codebase as-is.

### GPS Data

Sample data from @c4oe can be downloaded from [Google Drive](https://drive.google.com/file/d/0B38sjwW-2zVFRWhvbFlCWnRKM2c/view?usp=sharing).

The field labels are:

1. Message type: may be interrogation, squitter, or actual ADSB position message.
2. Status
3. Flight
4. Callsign
5. Mode-S Hex ID
6. Registration
7. Message date commence
8. Message time commence
9. Message date end
10. Message time end
11. Callsign 
12. Altitude
13. Speed
14. Heading
15. Latitude
16. Longitude
17. Vertical Speed
18. F-Flag
19. T-Flag
20. Vertical Speed Source
21. Mode 3

Most of these can be ignored. The important ones are `12. Altitude`, `13. Speed`, `15. Latitude`, & `16. Longitude`. Please note that you will need to pre-process this data to extract individual tracks for different aircraft. 

### Bounty

A [$500 bounty](https://www.bountysource.com/issues/5519417-geospatialcoordinateencoder-should-allow-optional-altitude) has been posted by @c4oe. See [more details on our mailing list](http://lists.numenta.org/pipermail/nupic_lists.numenta.org/2014-October/009803.html).

hernan-erasmo:
Fixes #1461 

Since the Coordinate Encoder doesn't have an upper limit on the number of dimensions of the coordinates, I think that using a 3D geocentric coordinate system to encode longitude, latitude and altitude should work.

This is the solution I propose:
- If no altitude is given, then continue to use the spherical mercator projection to calculate x and y coordinates from latitude and longitude.
- If altitude is given, then calculate the spherical mercator projection and transform the x and y values to geocentric 3d coordinates using WGS84 datum and the given altitude.


hernan-erasmo:
@rhyolight Sorry, but I just saw [the message you sent to the mailing list](http://lists.numenta.org/pipermail/nupic-hackers_lists.numenta.org/2014-December/003005.html) where you say that it is no longer needed to update the CHANGELOG manually. Do I still need to update it?

If it's needed then I have no problem in doing it, but I wanted to check with you first since you wrote that mail after your comment. 

Thank you.

Issue #1516, opened 2014-11-13
Upgrade to numpy 1.9.2


rhyolight:
This message has been a [pain](http://lists.numenta.org/pipermail/nupic_lists.numenta.org/2014-July/004344.html) since we open-sourced NuPIC:

    RuntimeError: module compiled against API version 9 but this version of numpy is 7

New users run into this, and need to uninstall their system version of `numpy` to fix it. It needs to be fixed once and for all on our end, and I think upgrading is the best way to fix it. 

* * *

- [ ] [Coordinate encoder should use integer seeds for random number generation](https://github.com/numenta/nupic/issues/1517)
- [x] [Upgrade to numpy 1.9.2 (linux64)](https://github.com/numenta/nupic-linux64/issues/32)
- [x] [Upgrade to numpy 1.9.2 (darwin64)](https://github.com/numenta/nupic-darwin64/issues/15)

flavius:
@rhyolight  Any news about this? I am using archlinux and I'd like to use nupic within sagemath, however installing nupic makes sagemath kaputt because sagemath was linked against the latest version of numpy.

rhyolight:
https://github.com/numenta/nupic/blob/e60704bb0f20f08594dd37c83ebb2e5a6b0cb1ca/requirements.txt#L21

We are already there!

flavius:
@rhyolight Well then sage is compiled with numpy 1.11, because I get ```RuntimeError: module compiled against API version 0xa but this version
of numpy is 0x9```

In general, I don't think it's a good idea for the nupic.bindings to mess with my system's packages when installing:

```
localhost# pip2 install https://s3-us-west-2.amazonaws.com/artifacts.numenta.org/numenta/nupic.core/releases/nupic.bindings/nupic.bindings-0.4.5-cp27-none-linux_x86_64.whl
Collecting nupic.bindings==0.4.5 from https://s3-us-west-2.amazonaws.com/artifacts.numenta.org/numenta/nupic.core/releases/nupic.bindings/nupic.bindings-0.4.5-cp27-none-linux_x86_64.whl
  Downloading https://s3-us-west-2.amazonaws.com/artifacts.numenta.org/numenta/nupic.core/releases/nupic.bindings/nupic.bindings-0.4.5-cp27-none-linux_x86_64.whl (4.1MB)
    100% |████████████████████████████████| 4.1MB 231kB/s 
[...]

PR #1659, opened 2014-12-27
pip requirements allow numpy>=1.7

breznak:
as a step to upgrading to numpy 1.9
part of #1516

### DIFF
--- a/external/common/requirements.txt
+++ b/external/common/requirements.txt
@@ -18,7 +18,7 @@ unittest2==0.5.1
 validictory==0.9.1
 PyMySQL==0.6.2
 DBUtils==1.1
-numpy==1.7.1
+numpy>=1.7.1
 tweepy==2.1
 pyproj==1.9.3
 prettytable==0.7.2


scottpurdy:
This change actually does upgrade numpy to the latest version and will mean that the version used is whatever the latest version is. This deviates from our current policy of fixing versions to ensure that we get deterministic builds.

@rhyolight my suggestion is that we explicitly change the version to the latest and close #1516.

rhyolight:
I thought that this change would NOT install a newer version of numpy if version 1.7 or higher already existed on the user system? So it only affects new installations of numpy with nupic.

scottpurdy:
That's true but it still means we have no idea what version they will have. And Travis, etc will be using whatever the latest is.

rhyolight:
Ok, but I didn't think upgrading to 1.9.1 was a simple task, that's why #1516 is open. I would love to do exactly what you said:

> we explicitly change the version to the latest and close #1516.

So we agree there. But I'm not sure that simply updating the requirements string is sufficient to close #1516.

PR #1663, opened 2014-12-27
Swig optimizations

### LINKED ISSUE #1671: Speedup swig cpp-py transitions by using proper compile flags

breznak:
with recent upgrade to the newest swig binary, add available `swig/python` optimization flags. 
edit:
Fixes #1671 

EDIT: descriptions of swig flags added

```
 - Add the -safecstrings/-nosafecstrings options to
	    enable/disable the use of safe conversions from PyString
	    to char *. Python requires you to never change the internal
	    buffer directly, and hence 'safectrings' warranties that
	    but returning a copy of the internal python string buffer.

	    The default, as in previous releases, is to return a
	    pointer to the buffer (nosafecstrings), so, it is the user's
	    responsibility to avoid its modification.
```


setup.py
```diff
@@ -18,7 +18,7 @@
 #   python setup.py install make_options="-j3"
 # which will add "-j3" option to Make commandline
 cmakeOptions = ""
-makeOptions = "install"
+makeOptions = "install -j4"
```
scottpurdy: We shouldn't include this by default because it can cause issues on some systems. When I build on small linux vps systems I have seen build errors that occur pretty reliably when using this option but don't show up without it.

CMakeLists.txt
```diff
@@ -366,6 +366,7 @@ include(${CMAKE_ROOT}/Modules/UseSWIG.cmake)
 
 include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 
+# swig options from: https://github.com/swig/swig/blob/master/Source/Modules/python.cxx#L111
```
scottpurdy: `swig` -> `SWIG`

CMakeLists.txt
```diff
@@ -376,11 +377,16 @@ set(CMAKE_SWIG_FLAGS
...
     -fastunpack
-    -nofastproxy
+    -nofastproxy #TODO remove this for optimization -O
     -fastquery
     -outputtuple
     -castmode
-    -w402
+    -fastdispatch
+    -nosafecstrings
+    -fastinit
```
scottpurdy: `-fastinit` is enabled by default

scottpurdy:
I don't know what the impact of a lot of these options are. @breznak - Can you please update the commit message or PR description or #1671 with an explanation for why you think each addition is a good thing?

setup.py
breznak: good point

breznak:
@scottpurdy i reviewed the flags with [swig changelog](http://www.swig.org/Release/CHANGES) and the code, reduced almost all of them - there were some small possible issues.

rhyolight:
@breznak Test failure.

Issue #1669, opened 2014-12-30
Encoders should use the "best" encoder for numbers


breznak:
As part of #1479 and other discussions, `RandomDistributedScalarEncoder` seems to be "encoder of choice" with performance better than `ScalarEncoder` (we keep ScalarEncoder for its simplicity and nice output for demonstration). 

However, most of the encoders internally use (extend or instantiate) the simple scalar encoder.

scottpurdy:
Thanks for the explanation. Really interesting results. I would have expected the scalar encoder to be faster. I will have to take a look and see what it is doing.

breznak:
> Really interesting results. I would have expected the scalar encoder to be faster.

@scottpurdy sorry, I have to take the speed numbers shamefully back :) , unfortunately I git clean'ed the file, so I can't see what I did wrong. Now ScalarEnc is about an order of magnitude faster than RDSE, and both are minimal compared to SP/TP speeds.

scottpurdy:
@breznak - thanks for following up!

breznak:
@scottpurdy continuing the discussion from PR here.. 

> I wouldn't assume that we want to switch encoders that subclass ScalarEncoder over to subclassing RDSE. The category encoder, for instance, wants the exact behavior of the ScalarEncoder, not the RDSE. Feel free to think through those and propose specific ones to switch over but that would be a separate effort.

This almost cuts the issue to nil, so what do we want? 
* leave the encoders' unittests as is?
* change integration and other unit tests that do not touch encoder's inner state (output of encodeIntoArray) to use RDSE instead of ScalarEncoder? 
* update the template files (.tpl) to have RDSE as default? 
* which encoders should swith to using RDSE? *) 

*) you said not `Category`, `Date` probably would not get much benefit (because of small range of values (1-31, 1-12, ..)), `LogEnc` should, so would `Delta` if it switched from extending AdaptiveScalar, `AdaptiveScalar` is supposted to be removed.

breznak:
an observation on RDSE vs ScalarEncoder and SP speed: 
SP speed depends O(n) on inputDimension range (not the dimension of inputDimension, but the value: [1000] vs [10000]). 
This input space must not be less than encoding produced by the encoder. 
With same resolution and input values range, RDSE produces much lower encoding (minval=0,maxval=10k,resolution=1,w=21: RDSE.n=400 (almost always const this value), Scalar.n=10k)

```
[...]

PR #1675, opened 2014-12-31
make RDSE the default encoder for scalars

breznak:
aims to transform (most of) encoders that use SimpleScalarEncoder internally to use RDSE instead.

Today's `ScalarEncoder` is renamed to `SimpleScalarEncoder`.

TODO:
- [ ] what to do about all the tests, that fail due to the different inner-workings of RDSE/(Simple)ScalarEncoder ? 
- [x] handle w,n params from SimpleScalar->RDSE init() -- resolution: do not create compatibility class
- [ ] set offset param from minval,maxval in RDSE init()
- [ ] fix tests

Fixes: #1669


breznak:
@chetan51 
> We have to fix the tests before we can merge this. What do you propose @breznak?

Either we, 
* `drop` the tests that touch inner-workings (bitarray representation) of an encoder - it'd be easier & more future-proof, but not so cool for spotting bugs

Or, 
* `bootstrap` the encoders, say "now it works" and replace each hard-coded scalarencoder array with output that RDSE would give for the same input. That is, running all the tests and accepting all the `assertEqual(expected, got)`

nupic/encoders/scalar.py
```diff
@@ -156,561 +62,17 @@ def __init__(self,
...
+    "get resolution by constructing a dummy SimpleScalar encoder and asking its resolution"
+    dummySimpleScalar = SimpleScalarEncoder(w, minval, maxval, periodic=periodic,
+                                            n=n, radius=radius, resolution=resolution,
+                                            name=name, verbosity=verbosity,
+                                            clipInput=clipInput, forced=forced)
+    res = dummySimpleScalar.resolution
+    super(ScalarEncoder, self).__init__(res,
+                                   #     w=w, #TODO should we: a) add 'forced' to RDSE and allow given w,n; or
+                                              #                b) pass only resolution and not w,n from former SimpleScalar to RDSE ?
+                                   #     n=n,
```
breznak: @chetan51 before any work on the tests, we need to decide how to handle the wrapper constructor from old ScalarEncoder: 
`RDSE(resolution, n=opt, w=opt)` vs `Scalar(w, minval, maxval, n=opt, resolution=opt)` - what to do with the optional params to RDSE? 

* We could ignore them completely (and put warning in the init() so users dont swarm over them!:) )
* use just `w` and set `n` to n=11*w which is said to be optimal value for RDSE.

breznak: Also, reading about `offset` param for RDSE, would it be useful to set it to (maxval-minval)/2 ?

chetan51: @scottpurdy Any thoughts on this?

breznak:
> Instead we should just make RDSE the default by convention 

@scottpurdy ok, that's a solution too, but will require more manual work to fix the constructors. I will revert the changes. Still a couple of questions: 

* is `from .. import RandomDistributedScalarEncoder as ScalarEncoder` ok? So I need to change only the params? 
* how to transform the tests? Bootstrap? (see comment above)
* what to do with the `n, w, (offset=)(maxval-minval)/2` params which are there for ScalarEnc and could be used to construct a RDSE?

scottpurdy:
@breznak -

- `import RandomDistributedScalarEncoder as ScalarEncoder` seems confusing to me but you could use a short abbreviation like `RDSE`
- which tests are you worried about? The tests for ScalarEncoder and RandomDistributedScalarEncoder don't need to be changed, right?
- My understanding from [the documentation](https://github.com/numenta/nupic/blob/9dc7548bfd241eec6cba0ea6e4b7b860f801f60d/nupic/encoders/random_distributed_scalar.py#L79) is that each bucket in RDSE represents a value range equal to `resolution`. The value range for buckets in the ScalarEncoder is `(maxVal-minVal)/(n-w+1)`. So give a ScalarEncoder min, max, n, and w, the RDSE resolution can be computed as `resolution=(maxVal-minVal)/(n-w+1)`.

breznak:
@scottpurdy 
> which tests are you worried about? The tests for ScalarEncoder and RandomDistributedScalarEncoder don't need to be changed, right?

`multi, data, category, logenc` use ScalarEncoder internally, some of their tests check exact hard-coded output-field which will be changed and not so elegant when switching to RDSE. 

I think the formula may not be 100% correct: `resolution=(maxVal-minVal)/(n-w+1)` this way, the more values, the finer resolution, and the more bits for the encoder, the worse res. Maybeit's just switched? 
Anyway, such nice formula leads to idea of creating alternative constructor (maybe in RDSE then?) that would just compute the resolution and create RDSE from it.

scottpurdy:
@breznak - I wouldn't assume that we want to switch encoders that subclass ScalarEncoder over to subclassing RDSE. The category encoder, for instance, wants the exact behavior of the ScalarEncoder, not the RDSE. Feel free to think through those and propose specific ones to switch over but that would be a separate effort.

scottpurdy:
Oh and regarding the conversion equation, see the link I put to the doc string for `resolution`. I think it says that `resolution` is the value range of each bucket, which may be reverse of the actual English definition of "resolution" but is what I implemented in the equation.

And you could propose that we always specify `resolution` (perhaps change name to `valueRange` or something more accurate) or always specify `minVal` and `maxVal` but again, there is no reason to try to make different encoders have the same parameters solely for that purpose.

breznak:
Closing as we decided to take another approach to the issue. Issue is still open.

Issue #1712, opened 2015-01-04
Allow predictions for multiple fields per model


scottpurdy:
It would be nice if you could specify multiple fields to make predictions for.

This will require a change to the CLAModel to support multiple classifier regions. It currently supports only one with the region name "Classifier":
https://github.com/numenta/nupic/blob/0cc904e5835a45e541eb0cac92901271b8f91108/nupic/frameworks/opf/clamodel.py#L1117

Additionally, the OPF and description.py format will need to be updated to support multiple fields to predict. It require some changes to metrics or swarming so it knows which field (or fields?) to use for optimization.

There are two implementations of the classifier, so we will need a task in nupic.core for the C++ implementation.

subutai:
OPF is setup in an OO way to support different model features. Is it possible to implement this feature as a subclass of CLAModel or perhaps as a separate independent model type?  This would avoid disrupting or complicating the (already overly complex) clamodel.py.  The factory would be used to instantiate the correct class type and after that everything else should just work.

rhyolight:
@salexashenko You going to tackle this?

salexashenko:
Yes

JonnoFTW:
Any progress on this so far?

rhyolight:
Nothing yet.

yyccR:
So, any progress on this till now?

rhyolight:
Nothin yet. To be clear, Numenta has not prioritized this task. If any work is going to get done here, it is going to be community-driven.

barnettjv:
Guys, its 2020. Has anyone made any progress on multi field predictions? I'm hoping so.

breznak:
> Guys, its 2020. Has anyone made any progress on multi field predictions? I'm hoping so. 

Have a look at actively developed 
https://github.com/htm-community/htm.core

We don't use OPF anymore but it's easy to setup multiple `Classifiers` to give you `multi-label` predictions. Cheers

barnettjv:
Thank you so much!!!! My Ph D. dissertation now stands a chance!
s/
Jeff Barnett

barnettjv:
[...]

Issue #1710, opened 2015-01-03
new (Multi)Metric that does linear combination of existing metrics


breznak:
sort of like MultiEncoder, use existing metrics, and weight them, eg
`0.1*speedBasedMetric+0.9*altMAPE` 

Relevant: #1363 

edit: 
suggested approach: construct instances of desired (sub)metrics and manually assign them to the multimetric in description.py, this avoids complicated parsing of "definition-strings". Ie.:
`metric100=Metric(type='altMAPE', windowSize=100, ..)
metric1000=Metric(type='altMAPE', windowSize=1000,...)
multi=Metric(type='multi', weights=[0.2 0.8], metrics=[metric100 metric1000])`

edit: 
- [ ] still blocked by #1848

breznak:
blocked by: #1848

PR #1726, opened 2015-01-06
speed metric

### LINKED ISSUE #1711: new metric that optimizes for speed
Reference: #1363

breznak:
the PR aims to implement a new metric that minimizes execution time of CLA model.

Fixes #1711


breznak:
@scottpurdy can you please make an early review on this? Esp. the data structures, how to pass iterationTime from `CLAModel.run()` to a new `Metric(AggregateMetric)`

nupic/utils.py
```diff
@@ -76,10 +97,55 @@ def compute(slidingWindow, total, newVal, windowSize):
...
 
 
   def getSlidingWindow(self):
-    return self.slidingWindow
+    return list(self._history)
+
+
+#######################################
+
+class CircularBuffer():
```
scottpurdy: Can we just use `collections.deque`?

tests/unit/nupic/utils_test.py
```diff
@@ -84,23 +84,19 @@ def testMovingAverageInstance(self):
     newAverage = ma.next(3)
     self.assertEqual(newAverage, 3.0)
     self.assertListEqual(ma.getSlidingWindow(), [3.0])
-    self.assertEqual(ma.total, 3.0)
```
scottpurdy: Why are we removing these test conditions?

nupic/frameworks/opf/clamodel.py
```diff
@@ -32,7 +32,9 @@
 import itertools
 import logging
 import traceback
-from collections import deque
+import time
+from collections import defaultdict, deque
```
scottpurdy: `defaultdict` is not used. Neither is `timedelta` on the next line

nupic/frameworks/opf/clamodel.py
```diff
@@ -419,7 +423,7 @@ def run(self, inputRecord):
     if self.__logger.isEnabledFor(logging.DEBUG):
       self.__logger.debug("inputRecord: %r, results: %r" % (inputRecord,
                                                             results))
-
+    results.iterationTime = time.clock() - startTime
```
scottpurdy: Since this isn't used internally can I suggest that clients calling the `run` method that want to know how long each run takes simply time it themselves? Or did you have some use case in mind that that wouldn't work for?

scottpurdy:
@breznak - please see my comments. And can you explain in more detail what you are trying to do with the iteration time?

Also thanks for cleaning up imports in metrics.py and removing duplicate MovingAverage code

nupic/utils.py
breznak: I think we cannot. Please see the comments below in description, this is a circular buffer with O(1) random access, which deque isn't. https://stackoverflow.com/questions/6427985/can-i-improve-the-speed-of-random-access-in-a-bounded-length-deque-circular-buf 

Maybe we can further optimize this CB and the MovingAverage, but it's not the goal of this PR, I just extracted the classes hidden in metrics.py

tests/unit/nupic/utils_test.py
breznak: the new implementation of MOvingAverage I took from metrics.py does not have the internal `total` field.

nupic/frameworks/opf/clamodel.py
breznak: yes, I want to create a new metric that has access to this information, since metric is called from swarming, the process of calling `model.run()` has to be same, I cannot use an alternative path..or?

> this metric would expose time per iteration from Model.run(), I imagine this combined with #1712 like `CombinedMetric(TimeMetric(window=50), 0.3 (=weight), altMAPE(window=50), 0.7)`

breznak:
@scottpurdy thanks for the review, fixed & commented. 

> can you explain in more detail what you are trying to do with the iteration time?

I want a new metric that measures speed of `model.run()`, I figured the only way to get this is to time it in the CLAModel.run() method and somehow expose the value of `model.iterationTime` further up to the Metrics. Please see #1711 for more details. 

Most importantly, I'm looking for advice how to propagate this variable from CLAModel to the Metric class called during swarming..?? 
I tried adding it to the `results` returned by model.run() https://github.com/numenta/nupic/blob/master/nupic/frameworks/opf/clamodel.py#L423 
But the `MetricsIFace` takes only `groundTruth and prediction` https://github.com/numenta/nupic/blob/master/nupic/frameworks/opf/metrics.py#L299 
Or which class in the "swarming" has access to both `Model`, inference and `Metrics`? so I could do something like: 
```
if metric.inInstance: SpeedMetric: 
  inference[prediction]=model.iterationTime
```

or something like that?

nupic/utils.py
scottpurdy: @breznak - good catch on the performance issue

tests/unit/nupic/utils_test.py
scottpurdy: ok makes sense!

scottpurdy: actually, if it doens't have an internal total then does that mean it recomputes the sum every iteration?

scottpurdy:
@breznak - Thanks, I understand much better what you are trying to do. I am a little hesitant to include an extra field on the model result that isn't going to be useful otherwise. I am also skeptical that swarming on model runtime speed will be useful. You mention it as a secondary metric but we don't have any way to have a secondary metric as far as I know.

But I would also like to support experimentation with this so I asked @chetan51 to take a look as well so we can see if there is a clean way to support this.

breznak:
@scottpurdy Thanks for helping out with experimenting with this idea.. 
>  I am also skeptical that swarming on model runtime speed will be useful. 

I can't tell without trying, but I believe it would, in some cases. Eg. I've seen a model where 1024 cols. was enough, but 4096 cols had tiny performance top, but visibli slower in processing time. This metric is to help this issue. 

> You mention it as a secondary metric but we don't have any way to have a secondary metric as far as I know.

True, this can be useful only as a secondary metric, that's why I created #1710 which should basically be weighted sum of other metrics, eg: `0.3*speedMetric + 0.7*altMAPE`

tests/unit/nupic/utils_test.py
breznak: good point, my lazy head missed that, will add it back.

chetan51:
I'm not very familiar with the OPF / swarming interface, but I'm taking a look at this now. I'll write up my impressions as I go.

As far as I can tell, the metrics interface currently assumes the definition of a metric being a difference between a prediction and ground truth. It doesn't seem to be general purpose enough to support metrics that are not about prediction accuracy. I don't see a clean way to support metrics of iteration time without rethinking the metrics interface. This sort of redesign might be even more necessary if we want to support secondary metrics or metrics of metrics, and swarm over them correctly.

Again, this is my first time reading this code, so take my impression with a grain of salt.

breznak:
@scottpurdy @chetan51 
ok, I found some way to go forward, interested to know your opinion.

`Metric.addInstance(self, groundTruth, prediction, record = None)` , like Chetan said, gt & prediction are not useful for this scenario, but we could hijack the record field. 

Currently: 
`
RECORD=
{'_timestamp': datetime.datetime(2010, 7, 4, 14, 0), 'consumption': 40.0, 'timestamp': datetime.datetime(2010, 7, 4, 14, 0), 'gym': 'Balgowlah Platinum', '_sequenceId': 1033868715996551153, 'address': 'Shop 67 197-215 Condamine Street Balgowlah 2093', '_timestampRecordIdx': 17614958, '_category': None, '_reset': 0}
`

1/ CLAModel appends `interationTime` to InputRecord (instead of ModelResults) and the metric can easily read it. (it already contains some helper fields - like _category,...) 

1a/ RecordSensor would do the work, instead of CLAModel -  attaching just the accessTime (time when the field was read), instead of the diff (startTime-time.now()), and the metric itself would compute the diff. 

2/ We change the `record` type from InputRecord to ModelResult which contains superset of the information (open possibilities for new metrics using the new extra info)

3/ adding extra argument to `addInstance(self, groundTruth, prediction, record = None, extraArgs = None)`

Personally I find 2/ the best - leaving more options for future, and less work to change. 1/ would be the minimal change imho

nupic/frameworks/opf/clamodel.py
```diff
@@ -418,7 +420,9 @@ def run(self, inputRecord):
...
+    # add to input record (passed to metrics)
+    inputRecord['time'] = results.iterationTime
```
breznak: @scottpurdy I can do it 2 ways here: 
* append the `time` field to `InputRecord` which is sent to the Metric
* use the functionality of `GlobalStore.get()` as in AnomalyMetric to access TP directly from the metric and read it from there. 

Which one do you think is better? To me it's the former I think.

scottpurdy: definitely not the global store

nupic/frameworks/opf/clamodel.py
breznak: actually, after a long time let to lay off, I came with a clean idea: 
How inaccurate do you think it would be if the time is measured:
* between each invocation of `metric.compute()` (or how is it called)
* instead of in `CLAModel.run()` at start-end of the method? 

This would radically simplify this metric/PR and make it stand-alone. Would that make sence? @scottpurdy @chetan51

breznak:
please comment on this: https://github.com/numenta/nupic/pull/1726#discussion_r30644427

rhyolight:
This PR is marked as ready for review @scottpurdy @chetan51 @subutai @BoltzmannBrain

nupic/frameworks/opf/metrics.py
```diff
@@ -261,22 +264,11 @@ def __init__(self, metricSpec):
...
-
-      groundTruthEncoding:
-        The binary encoding of the groundTruth value (as a numpy array). Right
-        now this is only used by CLA networks
-      predictionEncoding:
-        The binary encoding of the prediction value (as a numpy array). Right
-        now this is only used by CLA networks
+      @param groundTruth The actual measured value at the current timestep
+      @param prediction  The value predicted by the network at the current timestep
+      @param record (opt) TODO explain?
```
chetan51: Address TODO

nupic/frameworks/opf/metrics.py
```diff
@@ -261,22 +264,11 @@ def __init__(self, metricSpec):
...
 
-        return:
-            The average error as computed over the metric's window size
+      @return return The average error as computed over the metric's window size
```
chetan51: Extra `return`

nupic/frameworks/opf/metrics.py
```diff
@@ -748,48 +740,6 @@ class CustomErrorMetric(MetricsIface):
   """
     Custom Error Metric class that handles user defined error metrics
   """
-  class CircularBuffer():
```
chetan51: Please keep unrelated refactoring in separate PRs to make reviews simpler. Thanks!

nupic/utils.py
```diff
@@ -25,32 +25,57 @@
...
-    if not isinstance(windowSize, numbers.Integral):
-      raise TypeError("MovingAverage - windowSize must be integer type")
-    if  windowSize <= 0:
-      raise ValueError("MovingAverage - windowSize must be >0")
-
-    self.windowSize = windowSize
-    if existingHistoricalValues is not None:
-      self.slidingWindow = existingHistoricalValues[
-                              len(existingHistoricalValues)-windowSize:]
+    self._windowSize = windowSize
```
chetan51: Could you please explain the gist of these changes?

tests/unit/nupic/frameworks/opf/opf_metrics_test.py
```diff
@@ -787,5 +787,20 @@ def testMultiMetric(self):
 
 
 
+  def testSpeedMetric(self):
+    """testing the speed metric"""
```
chetan51: Unit tests should have assertions

david-ragazzi:
@breznak ping...

nupic/frameworks/opf/metrics.py
breznak: I'm not sure, I just filled missing param and marked that with TODO

breznak:
Thanks @chetan51 for review. Before I continue with it, can you please consider this one? 
https://github.com/numenta/nupic/pull/1726#discussion_r30644427

david-ragazzi:
> Thanks @chetan51 for review. Before I continue with it, can you please consider this one?
#1726 (comment)

@chetan51 ping...

nupic/frameworks/opf/clamodel.py
chetan51: I'm not familiar with the code you're referring to, so I'll defer to @scottpurdy on this one.

scottpurdy: @breznak - That would certainly limit the impact on the code. For the normal use case it would be accurate enough (it just needs to be accurate relative to other models being swarmed) but doesn't seem like a perfect solution either.

breznak:
@chetan51 @scottpurdy when you have time, please review. I did a bigger revamp, the PR is much simpler now. Thanks!

nupic/frameworks/opf/clamodel.py
```diff
@@ -396,11 +396,10 @@ def run(self, inputRecord):
       inferences = self._multiStepCompute(rawInput=inputRecord)
     # For temporal classification. Not used, and might not work anymore
     elif self._isClassificationModel():
-      inferences = self._classificationCompute()
-
-    results.inferences.update(inferences)
-
-    inferences = self._anomalyCompute()
+      inferences = self._classifcationCompute()
```
chetan51: `classification` is misspelled

nupic/frameworks/opf/metrics.py
```diff
@@ -1482,3 +1475,23 @@ def __repr__(self):
...
+    super(MetricSpeed, self).__init__(metricSpec)
+    self._last = time.time() # get current time
+
+  def getMetric(self):
+    return {'value': self.err, "stats" : {"weights" : self.weights}}
+
+  def addInstance(self, groundTruth, prediction, record = None):
+    elapsed = time.time() - float(self._last) # compute diff
+    self._last = time.time() 
+    assert elapsed >= 0
```
chetan51: Do we need this `assert`?

tests/unit/nupic/frameworks/opf/opf_metrics_test.py
```diff
@@ -787,5 +788,19 @@ def testMultiMetric(self):
...
+    ms = MetricSpec(field='a', metric='speed',  inferenceElement='multistepBestPredictions', params={'window': 1, 'steps': 1})
+    mSpeed = getModule(ms)
+    print mSpeed
+
+    for _ in xrange(5):
+      rand_delay=np.random.randint(1, 10)/float(10)
+      time.sleep(rand_delay)
+      metricValue = float(mSpeed.addInstance(0, 0))
+#      print metricValue
```
chetan51: Remove commented-out code

tests/unit/nupic/frameworks/opf/opf_metrics_test.py
```diff
@@ -787,5 +788,19 @@ def testMultiMetric(self):
 
 
 
+  def testSpeedMetric(self):
+    """testing the speed metric"""
+    ms = MetricSpec(field='a', metric='speed',  inferenceElement='multistepBestPredictions', params={'window': 1, 'steps': 1})
+    mSpeed = getModule(ms)
+    print mSpeed
```
chetan51: Remove `print` statement in unit test

breznak:
Thanks @chetan51 ! Addressed, I still have a problem with some example when I fix things in f61188f and a8316fe  :(

tests/unit/nupic/frameworks/opf/opf_metrics_test.py
breznak: done

breznak:
@rhyolight this looks very simple, any idea why https://travis-ci.org/numenta/nupic/jobs/67079268#L3982
would be failing?

rhyolight:
@breznak I don't know, but here is the actual error message:

> TestIdx 17: Experiment experiments/anomaly/temporal/simple failed. 
> The actual result for anomalyScore:passThruPrediction:window=1000:field=f (0.0) was outside the allowed range of (0.13, 0.14)

rhyolight:
@breznak Does this test pass for you locally?

breznak:
@rhyolight nope, it's also failing here. Not sure why, but sometimes I get prediction=None, I did some small hacks not to crash on that situation, but then I get this scenario where avg becomes 0.0

> Thanks @chetan51 ! Addressed, I still have a problem with some example when I fix things in f61188f and a8316fe :(

nupic/frameworks/opf/clamodel.py
```diff
@@ -396,10 +396,9 @@ def run(self, inputRecord):
     # For temporal classification. Not used, and might not work anymore
     elif self._isClassificationModel():
       inferences = self._classificationCompute()
-
-    results.inferences.update(inferences)
-
-    inferences = self._anomalyCompute()
+    else:
```
scottpurdy: I'm pretty sure this changes the results. Before the final results.inferences would contain the anomalyCompute results in addition to results from one of the previous calls.

PR #1730, opened 2015-01-07
Faking extensions to get platform-specific wheels.

### LINKED ISSUE #1729: Ensure platform-specific wheel files are produced during release
Currently, only platform-independent files are created.

rhyolight:
Also starts producing and uploading sdist and bdist along with
bdist_wheel files to pypi. This does not fix the fact that linux
platform wheels cannot be uploaded to pypi, but it does at least create
the linux wheel file.

Fixes #1729.


rhyolight:
@oxtopus

oxtopus:
:+1:

PR #1760, opened 2015-01-17
Removes StdRandom class that was defined in math.i but never used.

### LINKED ISSUE #1759: StdRandom class defined in math.i but never used
https://github.com/numenta/nupic/blob/ef487ebe7ef67ca31202a1bee2d9cdd555569ae5/nupic/bindings/math.i#L369

Looks like it was supposed to provide some numpy.random functionality but this conflicts with our goal of sharing the functionality between C++ and Python. And it was never used so should be removed.

scottpurdy:
fixes #1759

### DIFF
--- a/nupic/bindings/math.i
+++ b/nupic/bindings/math.i
@@ -363,19 +363,3 @@ inline PyObject* shuffle(PyObject* obj)
 }
 
 } // End extend nupic::Random.
-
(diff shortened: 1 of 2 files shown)


scottpurdy:
@chetan51 please review

chetan51:
:+1:

PR #1764, opened 2015-01-20
Redzone filter set at five 9s; fixes #1722

### LINKED ISSUE #1722: Anomaly filtering bug
There are cases where anomaly_likelihood._filterLikelihoods() does not behave as expected -- moves an anomaly into the yellowzone when it should not. This has been evident in the Taurus app and needs to be tested in anomaly_likelihood_test.py.

BoltzmannBrain:
@subutai please review. Fixes #1722 
Changing the redzone threshold from 0.9999 to 0.99999, to be tested qualitatively.


subutai:
:+1: I think you need to put fixes #1722 in the commit message.

BoltzmannBrain:
@rhyolight please merge, thanks!

PR #1800, opened 2015-02-03
Use gcc in default docker configuration to match nupic.core binary release.

### LINKED ISSUE #1802: Dockerfile doesn't build
I think it has to do with the binary nupic.core release, which is built with gcc.  The default Dockerfile uses clang to build nupic, but installs nupic.core binaries.

oxtopus:
Also increases resources in coreos configuration.

Not sure if there's an issue for this yet, but will create one if not.

@rhyolight @allanino This works for me in the included CoreOS+Vagrant configuration.

Fixes #1802


rhyolight:
:+1: No issue yet.

rhyolight:
p

PR #1839, opened 2015-02-19
Removes isDelta method from encoder base class.

### LINKED ISSUE #1829: Remove delta encoder special casing
The CLAModel has special casing for delta encoders that enable the delta encoder to be the predicted field, where the delta is fed to the model but the classifier predicts the actual, non-delta, future value. I'm not sure this is a real use-case so wanted to open an issue for discussion. The delta encoder makes sense as an auxiliary field, similar to the various ways to encode dates, but not as a way to predict the original values.

If this isn't a real use case then we can clean up a lot of special-casing code in the encoder base class and the CLAModel.

@subutai - can you comment on this?

scottpurdy:
CLAModel now uses `isinstance(encoder, DeltaEncoder)` to special case the delta encoder.

@subutai please review

fixes #1829

### DIFF
--- a/nupic/encoders/base.py
+++ b/nupic/encoders/base.py
@@ -85,13 +85,6 @@ class Encoder(object):
     raise Exception("getWidth must be implemented by all subclasses")
 
   ############################################################################
-  def isDelta(self):
-    """
-    @returns true if the underlying encoder works on deltas
-    """
-    return False
-
-  ############################################################################
   def encodeIntoArray(self, inputData, output):
     """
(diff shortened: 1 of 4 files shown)


breznak:
:+1:

rhyolight:
Ignore the coveralls failure. I don't know when they decided to start using the status API.

cogmission:
Saturday or Sunday apparently... :(  Got me too!

rhyolight:
:+1:

scottpurdy:
@chetan51 - please review

chetan51:
:+1:

PR #1845, opened 2015-02-21
Anomaly: new  binary output mode

### LINKED ISSUE #1840: Add new output mode: binary anomaly (threshold) to Anomaly class
this will compare the `anomalyScore >= threshold? anomalyScore=1 : anomalyScore=0 ` 
I saw this feature somewhere else in the algorithm (AnomalyRegion?) and I think it would be simpler to put the logic to the Anomaly class and remove it elsewhere.

breznak:
Fixes #1840


rhyolight:
@subutai or @scottpurdy ?

scottpurdy:
@breznak - I will take a look at this tomorrow. One initial thing is that it would be nice if you squashed the initial commits together. It is ok to have more than one if there are logical units of work done but we should try to keep the commit history somewhat clean.

breznak:
:+1: thanks! Squashed (maybe too much :package: )

nupic/algorithms/anomaly.py
```diff
@@ -140,6 +148,17 @@ def compute(self, activeColumns, predictedColumns,
...
 
+    # apply binary discretization if required
+    if self._binaryThreshold is not None:
+      if score >= self._binaryThreshold:
+        score = 1.0
+      else:
+        score = 0.0
+
+    # final check on score within [0,1]
+    score = int(score * 100000000)/float(100000000) # 7 dec places
```
scottpurdy: you shouldn't have to do this

nupic/algorithms/anomaly.py
```diff
@@ -140,6 +148,17 @@ def compute(self, activeColumns, predictedColumns,
...
+    assert score >= 0 and score <= 1
```
scottpurdy: `assert score >= 0.0 and score <= 1.0`

nupic/algorithms/anomaly.py
```diff
@@ -100,6 +102,12 @@ def __init__(self, slidingWindowSize = None, mode=MODE_PURE):
       raise ValueError("Invalid anomaly mode; only supported modes are: "
                        "Anomaly.MODE_PURE, Anomaly.MODE_LIKELIHOOD, "
                        "Anomaly.MODE_WEIGHTED; you used: %r" % self._mode)
+    self._binaryThreshold = binaryAnomalyThreshold
+    if binaryAnomalyThreshold is not None and (
+          (binaryAnomalyThreshold >= 1  or binaryAnomalyThreshold <= 0) or
```
scottpurdy: `(binaryAnomalyThreshold >= 1.0  or binaryAnomalyThreshold <= 0.0)`

nupic/algorithms/anomaly.py
```diff
@@ -100,6 +102,12 @@ def __init__(self, slidingWindowSize = None, mode=MODE_PURE):
...
+          not isinstance(binaryAnomalyThreshold, float) ):
```
scottpurdy: put check for whether it is a float before the bounds checks

nupic/algorithms/anomaly.py
```diff
@@ -140,6 +148,17 @@ def compute(self, activeColumns, predictedColumns,
     if self._movingAverage is not None:
       score = self._movingAverage.next(score)
 
+    # apply binary discretization if required
+    if self._binaryThreshold is not None:
```
scottpurdy: In the case where there is a binary threshold specified and it is also doing a moving average, should the threshold be applied before or after the moving average? Make this clear in the doc string for the binary threshold in the constructor.

nupic/algorithms/anomaly.py
breznak: i do, probably due to numeric instability bug (in numpy?) - I was getting a CLA run with anomaly scores = x.xxxx*10e-13

scottpurdy:
You will need to set the binary threshold attribute in the `__setstate__` method or old models will end up with exceptions.

nupic/algorithms/anomaly.py
scottpurdy: as long as it is positive it shouldn't be a problem?

breznak: sorry, my bad. but the problem was sum(anomalyScore) was negative... should I remove the rounding and let assert prove/catch it?

nupic/algorithms/anomaly.py
breznak: good point! actually, both make some sence. What do you think? 
CC @subutai

breznak:
> You will need to set the binary threshold attribute in the __setstate__ method or old models will end up with exceptions.

I do in https://github.com/numenta/nupic/pull/1845/files#diff-5bf13635ee2cded8c4bfe689ea9a297dR194 

..thank you for review @scottpurdy , I addressed the issues

nupic/algorithms/anomaly.py
scottpurdy: Are you adding this for a real need that you have or just because you think it may be useful to someone? In the former case, which is useful for you?

nupic/algorithms/anomaly.py
```diff
@@ -100,6 +103,13 @@ def __init__(self, slidingWindowSize = None, mode=MODE_PURE):
       raise ValueError("Invalid anomaly mode; only supported modes are: "
                        "Anomaly.MODE_PURE, Anomaly.MODE_LIKELIHOOD, "
                        "Anomaly.MODE_WEIGHTED; you used: %r" % self._mode)
+    self._binaryThreshold = binaryAnomalyThreshold
+    if binaryAnomalyThreshold is not None and ( 
+          not isinstance(binaryAnomalyThreshold, float) or
+          binaryAnomalyThreshold >= 1.0  or
```
scottpurdy: actual since we are checking out of bounds this and next line should be strictly `>` and `<`

scottpurdy:
With regards to backwards compatibility, the code you linked to just shows that it will be set for new instances. Old serialized models will not have it set and will break when you try to use them. If you aren't familiar with Python serialization with pickle, please read up on pickle, `__getstate__`, and `__setstate__`.

nupic/algorithms/anomaly.py
breznak: I am using `moving avg, then discretize` to have fewer but compact packets of anomaly detected events; now I don't see much utility in doing it the other way. 

Actualy, the problem of ordering when to do MA gives a lot of possibilities we could judge. 

`pure _ (opt. likelihood/weighted) _ (discretize) _` where `moving average` can be done at any `_`

subutai: We always apply a threshold after the likelihood calculation. Thresholding the raw anomaly score (whether averaged or not) usually doesn't work.

nupic/algorithms/anomaly.py
breznak: actually, I defined the threshold on open interval (0,1) because 0/1 values would not make much sense: no anomalies/all anomalies.

breznak:
done

breznak:
@scottpurdy this should be ready..

nupic/utils.py
```diff
@@ -83,3 +83,12 @@ def next(self, newValue):
...
   def getSlidingWindow(self):
     return self.slidingWindow
+
+
+  def __setstate__(self, state):
+    self.__dict__.update(state)
+
+    if not hasattr(self, "total"):
+      self.total = 0
+      self.slidingWindow = []
```
scottpurdy: If the serialized version had a sliding window do you really want to reset it? How about replace these lines with:

```
self.total = sum(self.slidingWindow)
```

breznak:
@scottpurdy thanks! fixed. ready?

rhyolight:
@scottpurdy please approve.
@breznak please merge in master.

scottpurdy:
@breznak - Do you get better results with this functionality than with anomaly likelihood? To @subutai 's point, we did a lot of work figuring out how to compute useful anomaly scores and this seems contrary to our experience. I'm ok merging this, but want to make sure it is actually useful code.

breznak:
@scottpurdy I actually do. 
* This is more like a post-processing filter, so it's combinable with the likelihood mode. 
* it replaces existing functionality of `autoDetectThreshold`, just moving it from ClassifierRegion to Anomaly where it logically belongs
* my aim for this is using it in AnomalyMetric to reduce anomalies to certain percentage.

rhyolight:
@breznak Are there any docs?

breznak:
Not but the code comments/param descriptions. Should there be more?

rhyolight:
I see. I think that's good for now. The code docs will update themselves with the new param.

Issue #1848, opened 2015-02-21
Select swarm optimization criterion (minimize) by metric's ID, not "string"


breznak:
for OPF, metrics used in an experiment are defined in `swarm/description.py` in `Metrics= []` and in `permutations.py` an optimization criterion `minimize=` is defined, which must point to one of the metrics from description.py, but is defined by "a string representation of the metric" which can be parsed to cunstruct a metric. 

I suggest defining the `minimize=` by id of required metric instance, which will: 

* avoid problems where user changes (mistakenly) the metric description in permutations.py, but not in descrioption.py and gets an error. 

* allow us to create "special" metrics, that cannon be constructed by "parsing the string"

breznak:
` vim nupic/swarming/ModelRunner.py`

rhyolight:
> `vim nupic/swarming/ModelRunner.py`

@breznak What does your comment mean?

breznak:
a note where I should look

PR #1850, opened 2015-02-23
Metric multi

### LINKED ISSUE #1710: new (Multi)Metric that does linear combination of existing metrics

breznak:
new MetricMulti allows us to combine weighted values of many sub-metrics with different parameters. 
Also reuses shared MovingAverage code in metrics.
Implement `multi metric` + tests + cleanup

Fixes #1710


breznak:
there's a bug in the test values, I can't see why..

rhyolight:
@breznak Please don't assign PRs to milestones. I would only like issues to be assigned to milestones because it makes the roadmap cleaner.

breznak:
Yay! 
@scottpurdy please review

scottpurdy:
@chetan51 - can you take this one?

examples/opf/experiments/anomaly/temporal/hotgym/description.py
```diff
@@ -403,7 +403,7 @@
                  params={'errorMetric': 'aae', 'window': 1000, 'steps': steps}))
   control['metrics'].append(
       MetricSpec(field=config['predictedField'], metric='multiStep',
-                 inferenceElement='multiStepBestPredictions',
+                 inferenceElement='anomalyScore',
```
chetan51: What is the motivation for this change?

nupic/frameworks/opf/metrics.py
```diff
@@ -184,54 +186,14 @@ def getModule(metricSpec):
...
-
+  elif metricName == 'multi':
+    return MetricMulti(metricSpec)
   else:
     raise Exception("Unsupported metric type: %s" % metricName)
 
 ################################################################################
 #               Helper Methods and Classes                                    #
 ################################################################################
-class _MovingAverage(object):
```
chetan51: Nice cleanup!

nupic/frameworks/opf/metrics.py
```diff
@@ -338,7 +300,7 @@ class AggregateMetric(MetricsIface):
   """
   ___metaclass__ = ABCMeta
 
-
+  #FIXME @abstractmethod - this should be marked abstract method and required to be implemented
```
chetan51: Are these separate issues filed to take care of these and the other FIXME?

nupic/frameworks/opf/metrics.py
```diff
@@ -1459,7 +1424,63 @@ def addInstance(self, groundTruth, prediction, record = None):
...
 
+###################################################################################
+class MetricMulti(MetricsIface):
+  """Multi metric can combine multiple other (sub)metrics and 
+     weight them to provide combined score."""
 
+  def __init__(self, metricSpec):
+    """MetricMulti constructor using metricSpec is not allowed."""
+    raise ValueError("MetricMulti cannot be constructed from metricSpec string! "
+                     "Use MetricMulti(id,weights,metrics) constructor instead.")
```
chetan51: The order of arguments here isn't consistent with the order in the actual constructor below.

nupic/frameworks/opf/metrics.py
```diff
@@ -1459,7 +1424,63 @@ def addInstance(self, groundTruth, prediction, record = None):
...
+
+  def __init__(self, weights, metrics, id=None, window=None):
+    """MetricMulti 
+       @param id - unique name for this metric
+       @param weights - [list of floats] used as weights
+       @param (sub)metrics - [list of metrics] 
+       @param id - (opt) name of the metric
```
chetan51: Duplicate param

nupic/frameworks/opf/metrics.py
```diff
@@ -1459,7 +1424,63 @@ def addInstance(self, groundTruth, prediction, record = None):
...
+       @param window - (opt) window size for moving average, or None
```
chetan51: Please indicate what will happen if `None` is passed

nupic/frameworks/opf/metrics.py
```diff
@@ -1459,7 +1424,63 @@ def addInstance(self, groundTruth, prediction, record = None):
...
 
 
+  def addInstance(self, groundTruth, prediction, record = None):
+    err = 0.0
+    subResults = [m.addInstance(groundTruth, prediction, record) for m in self.metrics]
+    for i in xrange(len(self.weights)):
+      try:
+        err += subResults[i]*self.weights[i]
+      except(TypeError):
+        self.err = None
```
chetan51: Should this throw an exception?

nupic/frameworks/opf/metrics.py
```diff
@@ -1459,7 +1424,63 @@ def addInstance(self, groundTruth, prediction, record = None):
...
+     weight them to provide combined score."""
 
+  def __init__(self, metricSpec):
+    """MetricMulti constructor using metricSpec is not allowed."""
+    raise ValueError("MetricMulti cannot be constructed from metricSpec string! "
+                     "Use MetricMulti(id,weights,metrics) constructor instead.")
+
+  def __init__(self, weights, metrics, id=None, window=None):
+    """MetricMulti 
+       @param id - unique name for this metric
```
chetan51: What is this `id` used for?

tests/unit/nupic/frameworks/opf/opf_metrics_test.py
```diff
@@ -758,6 +758,34 @@ def testCustomErrorMetric(self):
...
+    multi = MetricMulti(weights=[0.2, 0.8], metrics=[metric10, metric1000], id='multi1')
+    multi.verbosity = 1
+    print multi 
+    # create reference metrics (must be diff from metrics above used in MultiMetric, as they keep history)
+    metric1000ref = getModule(ms1)
+    metric10ref = getModule(ms2)
+
+    
+    gt = [(i/4+1) for i in range(500, 1000)]
+    p = [i for i in range(500)]
```
chetan51: Can just be `p = range(500)`.

tests/unit/nupic/frameworks/opf/opf_metrics_test.py
```diff
@@ -758,6 +758,34 @@ def testCustomErrorMetric(self):
...
+    # create multi metric
+    multi = MetricMulti(weights=[0.2, 0.8], metrics=[metric10, metric1000], id='multi1')
+    multi.verbosity = 1
+    print multi 
+    # create reference metrics (must be diff from metrics above used in MultiMetric, as they keep history)
+    metric1000ref = getModule(ms1)
+    metric10ref = getModule(ms2)
+
+    
+    gt = [(i/4+1) for i in range(500, 1000)]
```
chetan51: Isn't this equivalent to the simpler `gt = range(100, 200)`?

examples/opf/experiments/anomaly/temporal/hotgym/description.py
breznak: some mistake, reverted. thanks

nupic/frameworks/opf/metrics.py
breznak: https://github.com/numenta/nupic/issues/1892

nupic/frameworks/opf/metrics.py
breznak: I think not, as it's the submetrics who can pass None, so I leave it upto the to handle that.

nupic/frameworks/opf/metrics.py
breznak: preparation for the next steps. in the `minimize=` field of `permutations.py` the MultiMetric cannot be described/constructed by the string, so I want to make it possible to identify the metric by `id`

tests/unit/nupic/frameworks/opf/opf_metrics_test.py
breznak: it's not. but actually the ranges do not matter much, I just copied them from other test.

breznak:
@chetan51 thanks for review! fixed

nupic/frameworks/opf/metrics.py
```diff
@@ -1459,7 +1424,62 @@ def addInstance(self, groundTruth, prediction, record = None):
...
 
+  def __init__(self, metricSpec):
+    """MetricMulti constructor using metricSpec is not allowed."""
+    raise ValueError("MetricMulti cannot be constructed from metricSpec string! "
+                     "Use MetricMulti(weights,metrics) constructor instead.")
+
+  def __init__(self, weights, metrics, id=None, window=None):
+    """MetricMulti 
+       @param weights - [list of floats] used as weights
+       @param (sub)metrics - [list of metrics]
```
chetan51: Replace with `@param metrics - [list of submetrics]` for clarity

nupic/frameworks/opf/metrics.py
chetan51: Sorry, I don't understand your reply, can you elaborate?

nupic/frameworks/opf/metrics.py
```diff
@@ -1459,7 +1424,62 @@ def addInstance(self, groundTruth, prediction, record = None):
...
+  def __init__(self, metricSpec):
+    """MetricMulti constructor using metricSpec is not allowed."""
+    raise ValueError("MetricMulti cannot be constructed from metricSpec string! "
+                     "Use MetricMulti(weights,metrics) constructor instead.")
+
+  def __init__(self, weights, metrics, id=None, window=None):
+    """MetricMulti 
+       @param weights - [list of floats] used as weights
+       @param (sub)metrics - [list of metrics] 
+       @param id - (opt) name of the metric
```
chetan51: @breznak explained what this `id` is for:

> preparation for the next steps. in the minimize= field of permutations.py the MultiMetric cannot be described/constructed by the string, so I want to make it possible to identify the metric by id

What about the other metrics? If `id` is optional, how do you plan to minimize over them?

nupic/frameworks/opf/metrics.py
breznak: I only return None when the existing metric would return None, so just keeping status quo..

nupic/frameworks/opf/metrics.py
breznak: other metrics can use the existing `minimize=<string>` form, multi metric cannot be constructed using this string, thus needs a special case - the id.

chetan51: I think we should consolidate them into a single approach for identifying metrics. Special cases can get messy. What would you propose as a single id mechanism?

breznak:
@chetan51 thanks! how does it look now? :)

nupic/frameworks/opf/metrics.py
chetan51: This try/catch is confusing and can introduce bugs that are hard to catch (if `subResults` contains a string instead of `None`, for example, this metric would silently return `None`). Please refactor.

nupic/frameworks/opf/metrics.py
breznak: I already have a ticket for this, https://github.com/numenta/nupic/issues/1848 I think the `minimize=` should take an ID, rather then the current string description, as the latter leads to dumb bugs when one modifies his metric in description.py but forgets to do the same modification in permutation.py's minimize=. But that's not goal of this PR. I only include id in MetricMulti because it's new and can't be created from string anyway.

chetan51:
I suggest that we either add `id` to all metrics, or none of them (for now). It's weird for just `MetricMulti` to have optional `id`, especially if it's not even being used anywhere.

breznak:
@chetan51 agree. removed

chetan51:
Great, looks good to me now. Thanks @breznak! Feel free to merge after tests pass.

PR #1851, opened 2015-02-23
implement SP.randomSP and remove FlatSpatialPooler

### LINKED ISSUE #627: Remove flat spatial poolers
Related to issue #340 

The flat spatial poolers are no longer required (except for one tiny thing in #697).  We can remove py/nupic/research/flat_spatial_pooler.py as well as the associated C++ files (now in nupic.core) plus the tests.

### LINKED ISSUE #697: Add randomSP flag to spatial pooler
Add randomSP flag to spatial pooler to mimic usage in flat spatial pooler. This flag is useful for exploring some algorithmic properties of spatial poolers and easy to implement. The flag is implemented in `FlatSpatialPooler` and is the only remaining useful thing in that class. If we put this into `SpatialPooler` we can finally get rid of that class and create some nice examples around this.

Depends on #627.

breznak:
Fixes #627 
Fixes #697


breznak:
for some reason, I get a failure at: 
`======================================================================
FAIL: testActiveColumnsEqualNumActive (__main__.SpatialPoolerTest)
----------------------------------------------------------------------
  File "../../tests/unit/nupic/research/spatial_pooler_unit_test.py", line 1841, in testActiveColumnsEqualNumActive
    self.assertEqual(sum(activeArray), numActive)
AssertionError: 2.0 != 1
`
..investigating

breznak:
https://travis-ci.org/numenta/nupic/jobs/51848014#L3169 

CC @chetan51 seems like `numActiveColumnsPerInhArea` is not enforcing (the same way) as in flat SP?

chetan51:
@breznak Hmm, I can't tell what's different between the two by just looking at the code. It might require some debugging.

chetan51:
Also, I recommend renaming the `randomSP` parameter to `learn` (default to `True`). That would be more clear what the effect is.

tests/unit/nupic/research/spatial_pooler_compatability_test.py
```diff
@@ -100,6 +100,8 @@ def compare(self, pySp, cppSp):
                            cppSp.getMinPctOverlapDutyCycles())
     self.assertAlmostEqual(pySp.getMinPctActiveDutyCycles(),
                            cppSp.getMinPctActiveDutyCycles())
+#    self.assertEqual(pySp.getRandomSP(),  #FIXME enable when done in C++ too
```
chetan51: Please create a new issue for addressing this (if there isn't one already), so it doesn't get lost.

breznak: ok, referenced in https://github.com/numenta/nupic.core/issues/350

chetan51: Thanks!

breznak:
> Also, I recommend renaming the randomSP parameter to learn (default to True). That would be more clear what the effect is. 

There's already a `learn` in `SP.compute()`, would it make the `randomSP` useless, or do we want the global functionality?

chetan51:
@breznak I think we should expose the existing `learn` parameter externally. That would be better than adding another `randomSP` parameter.

breznak:
@chetan51 and what should take preference, when `SP.setLearn(False) & SP.compute(learn=True) ` ?

chetan51:
Hmm, good question. Now that I think about it, I would prefer to be consistent with the new `TemporalMemory` implementation, and not have a global `learn` parameter (requiring the user to call `compute` with `learn=False` each time). But I'm not sure how this would work with the OPF currently. @subutai your thoughts?

subutai:
:+1: to passing in learn=True or False each time (default should be True). I believe this is what the OPF and spatial pooler already does, so should be ok. In addition I think we can remove the `stripNeverLearned` option. So randomSP would simply consist of always calling compute with learn=False

david-ragazzi:
@breznak Remove FlatSpatialPooler is a trick job because some tests could be using them as the SP class. To we remove safely all old/derived SP classes we have to adapt some fields of these examples/tests to SpatialPooler class. Example: 

```
coincidencesShape	=> columnDimensions
inputShape		=> inputDimensions
coincInputRadius	=> potentialRadius
numActivePerInhArea	=> numActiveColumnsPerInhArea
coincInputPoolPct	=> potentialPct
```

I have to do this to remove FDRCSpatial2 in:https://github.com/numenta/nupic/pull/1872 , in this same PR I also removed FlatSpatialPooler because it is inherited from FDRCSpatial2 and was causing problems.

So how about this PR fix the SP.randomSP issue and we leave the removal for https://github.com/numenta/nupic/pull/1872 ?

breznak:
@david-ragazzi that was not an issue here, flat_SP was not used much, and I "renamed" the params accordingly. 

Closing this in favor of #1872 if you `git cherry-pick b813110e03065f881823c9a0d469f0c7dbc26450` as I moved some tests from flat_SP's unittests there, adn add it to your PR.

Issue #1864, opened 2015-02-24
Remove obsolete implementations of SP


david-ragazzi:
As addressed here: https://github.com/numenta/nupic/issues/1862

Stuff to be removed:
`FDRCSpatial2.py` and the unused code in `fdrutilities.py`

PR #1872, opened 2015-02-25
Remove FDRCSpatial2

david-ragazzi:
Fixes: #1864

This PR basically removes FDRCSpatial2 class.

To we safely remove this class we have to change `spatialImp` parameter from `oldpy` to `py` and update some old field names of these examples/tests to the field names used in the current SpatialPooler class. Example:

```
coincidencesShape => columnDimensions
inputShape => inputDimensions
coincInputRadius => potentialRadius
coincInputPoolPct => potentialPct
numActivePerInhArea => numActiveColumnsPerInhArea
```

One positive effect of this PR is that field names become more updated and uniform throughout the code.. :smiley:


examples/opf/tools/sp_plotter.py
```diff
@@ -26,7 +26,7 @@
 import copy
 import csv
 import numpy as np
-from nupic.research import FDRCSpatial2
+from nupic.research import SpatialPooler
```
breznak: from nupic.research.SpatialPooler import SpatialPooler 

so you don't need to do sp = SpatialPooler.SpatialPooler below

david-ragazzi: > from nupic.research.SpatialPooler import SpatialPooler 

Done.

breznak:
forgotten import https://travis-ci.org/numenta/nupic/jobs/52161513#L2574

david-ragazzi:
> forgotten import https://travis-ci.org/numenta/nupic/jobs/52161513#L2574

Fixed. Actually I removed 2 functions from `algorithms.i` which also are used by TPs.

david-ragazzi:
Dont't merge this! Travis status is not shown!

rhyolight:
Travis is broken today :sob:

nupic/bindings/algorithms.i
```diff
@@ -137,97 +134,6 @@ using namespace nupic;
 
 // TODO: This __really__ belongs somewhere else, but no other good place for it.
 
-/* FDRCSpatialInfer
```
david-ragazzi: Here I remove binding for c++ FDRCSpatial2.

examples/opf/tools/sp_plotter.py
```diff
@@ -330,17 +330,16 @@ def testSPNew():
   
   
   # Setup a SP
-  sp = FDRCSpatial2.FDRCSpatial2(
```
david-ragazzi: This and other examples still use the FDRCSpatial2 instead of new spatial pooler. So I had to convert them.

tests/integration/nupic/opf/opf_experiment_results_test.py
```diff
@@ -185,7 +185,7 @@ def testExperimentResults(self):
         'results': {
           ('DefaultTask.NontemporalMultiStep.predictionLog.csv',
            "multiStepBestPredictions:multiStep:errorMetric='avg_err':steps=1:window=25:field=field1"):
-                    (0.0, 0.0),
+                    (0.0, 1.0),
```
david-ragazzi: Some ranges in this set of tests had to be readjusted due new SP emmits other values for the same metric.

tests/unit/nupic/research/flat_spatial_pooler_compatability_test.py
```diff
@@ -1,533 +0,0 @@
-#! /usr/bin/env python
```
david-ragazzi: This and other tests related to flat_spatial_pooler are removed.

tests/unit/nupic/research/inhibition_object_test.py
```diff
@@ -39,10 +39,11 @@
 
 class InhibitionObjectTest(unittest.TestCase):
 
+  @unittest.skip("Currently fails due to switch from FDRCSpatial2 to SpatialPooler.")
```
david-ragazzi: It is not possible use the new SP to perform this test due to it not have explicit methods to get inhibition. So I skipped it. Now can remove it, try find alternatives, or then keep it as `skipped` until future review as addressed in https://github.com/numenta/nupic/issues/1895 .

tests/unit/nupic/research/sp_learn_inference_test.py
```diff
@@ -138,6 +140,7 @@ def _runLearnInference(self,
       print "Test succeeded"
 
 
+  @unittest.skip("Currently fails due to switch from FDRCSpatial2 to SpatialPooler.")
```
david-ragazzi: It is not possible use the new SP to perform this test due to it not have explicit methods to compute inference. So I skipped it. Now can remove it, try find alternatives, or then keep it as `skipped` until future review as addressed in https://github.com/numenta/nupic/issues/1895 .

tests/unit/nupic/research/spatial_pooler_unit_test.py
```diff
@@ -1786,6 +1788,57 @@ def testWrite(self):
     self.assertSetEqual(indices1, indices2)
 
 
+# moved from flat_spatial pooler tests: 
+
+  def testRandomSPDoesNotLearn(self):
```
david-ragazzi: This method was migrated from `flat_spatial_pooler` by @breznak

david-ragazzi:
Ready to review!

david-ragazzi:
@chetan51 Ready to review!

tests/unit/nupic/research/inhibition_object_test.py
breznak: I'd suggest leaving as is for now, maybe investigate if we should add the
getInhibition() fctionality..?

david-ragazzi:
@chetan51 ping

chetan51:
Hi @david-ragazzi, I've been away at a conference since last Thursday, and I'll be on vacation until next Monday. Sorry I won't be able to review this until then. Perhaps someone else can take a look in the meanwhile? Otherwise I'll take a look on Monday.

david-ragazzi:
Hi @david-ragazzi, I've been away at a conference since last Thursday, and I'll be on vacation until next Monday. Sorry I won't be able to review this until then. Perhaps someone else can take a look in the meanwhile? Otherwise I'll take a look on Monday.

Thanks, @chetan51 . I saw @subutai post in FB, you guys should be exhausted. I wait until monday If noboby review this until there. :-)

examples/opf/tools/sp_plotter.py
chetan51: Did you run these examples before and after to make sure they still work as before?

nupic/bindings/algorithms.i
```diff
@@ -137,97 +134,6 @@ using namespace nupic;
 
 // TODO: This __really__ belongs somewhere else, but no other good place for it.
```
chetan51: Is this comment relevant anymore?

nupic/bindings/algorithms.i
```diff
@@ -1142,358 +778,9 @@ inline PyObject* generate2DGaussianSample(nupic::UInt32 nrows, nupic::UInt32 nco
 }
 
 //--------------------------------------------------------------------------------
-// Optimizations for FDRCSpatial2
+// Optimizations for FDR
```
chetan51: Why can't we delete this whole section?

nupic/regions/SPRegion.py
```diff
@@ -640,24 +588,17 @@ def _doBottomUpCompute(self, rfInput, resetSignal):
 
     # Run SFDR bottom-up compute and cache output in self._spatialPoolerOutput
```
chetan51: Is this comment obsolete?

tests/integration/nupic/opf/opf_experiment_results_test.py
chetan51: Does this mean the new SP has higher error rate than the old one?

tests/unit/nupic/algorithms/sp_overlap_test.py
```diff
@@ -176,7 +175,7 @@ def frequency(self,
     zeros = len([x for x in summ if x==0])
     factor = max(summ)*len(summ)/sum(summ)
     if len(reUsed) < 10:
-      self.assertLess(factor, 30,
+      self.assertLessEqual(factor, 40,
```
chetan51: Why was this change necessary?

tests/unit/nupic/research/inhibition_object_test.py
chetan51: Please update the skip comment to elaborate on why the switch causes this test to fail (as you said, the new SP doesn't have explicit methods to get inhibition).

tests/unit/nupic/research/sp_learn_inference_test.py
chetan51: Please update the skip comment to elaborate on why the switch causes this test to fail (as you said, the new SP doesn't have explicit methods to get inhibition).

tests/unit/nupic/research/spatial_pooler_unit_test.py
```diff
@@ -1786,6 +1788,57 @@ def testWrite(self):
...
+    sp.compute(inputArray, False, activeArray)
+    # Should have incremented general counter but not learning counter
+    self.assertEqual(sp._iterationNum, 1)
+    self.assertEqual(sp._iterationLearnNum, 0)
+
+    # Check the initial perm state was not modified either
+    self.assertEqual(sp._permanences, initialPerms)
+
+
+  @unittest.skip("Ported from FlatSP and fails") #FIXME
```
chetan51: Why does this fail? If there's a good reason it doesn't make sense to fix this now, please update the skip comment and elaborate.

chetan51:
@david-ragazzi I've left some comments and suggestions. Overall, great work!

david-ragazzi:
> @david-ragazzi I've left some comments and suggestions. Overall, great work!

Thanks, my friend.. I'm running into problems because currently I'm without my notebook.. :-( Soon I will answer you and implement the suggestions..

examples/opf/tools/sp_plotter.py
david-ragazzi: @chetan51 I didn't manage run these examples simply because they require previous data files that aren't in repository (`~/Desktop/ExperimentResults/Basil100x21.csv`, for example):

```
davidragazzi@davidragazzi-virtual-machine ~/Desktop/nupic $ python examples/opf/tools/sp_plotter.py
  File "examples/opf/tools/sp_plotter.py", line 572, in <module>
    testSPFile()
  File "examples/opf/tools/sp_plotter.py", line 472, in testSPFile
    file = open(inputFile, 'rb')
IOError: [Errno 2] No such file or directory: '~/Desktop/ExperimentResults/Basil100x21.csv'
```

Are you sure that this example is crucial and is updated? It seems that is quite old and a few or nobody really run this.

nupic/bindings/algorithms.i
david-ragazzi: > Is this comment relevant anymore?

Not, is not.. I removed it together with code which it is about.

nupic/bindings/algorithms.i
david-ragazzi: > Why can't we delete this whole section?

We cannot because the remaining `getSegmentActivityLevel` and `isSegmentActive` functions are currently used by TP (https://github.com/numenta/nupic/blob/master/nupic/research/TP.py#L36)

nupic/regions/SPRegion.py
david-ragazzi: > Is this comment obsolete?

Not, is not. Because the activity that it is refering (`compute and cache output`) continue being performed (https://github.com/david-ragazzi/nupic/blob/remove_old_SPs/nupic/regions/SPRegion.py#L596).

tests/unit/nupic/algorithms/sp_overlap_test.py
david-ragazzi: Because the computed factor is always 40 when we use the new spatial pooler.

tests/unit/nupic/research/inhibition_object_test.py
david-ragazzi: > Please update the skip comment to elaborate on why the switch causes this test to fail (as you said, the new SP doesn't have explicit methods to get inhibition).

Done.

tests/unit/nupic/research/sp_learn_inference_test.py
david-ragazzi: > Please update the skip comment to elaborate on why the switch causes this test to fail (as you said, the new SP doesn't have explicit methods to get inhibition).

Done.

tests/unit/nupic/research/spatial_pooler_unit_test.py
david-ragazzi: > Why does this fail? If there's a good reason it doesn't make sense to fix this now, please update the skip comment and elaborate.

We don't know. As @breznak already created an issue for this test (https://github.com/numenta/nupic/issues/1897), I simply referenced it in the code.

tests/integration/nupic/opf/opf_experiment_results_test.py
david-ragazzi: > Does this mean the new SP has higher error rate than the old one?

Not, on the contrary. We can simply use a range of 1.0 to 1.0.

david-ragazzi:
@chetan51 I answered all your questions and implemented some suggestions.

examples/opf/tools/sp_plotter.py
chetan51: @rhyolight Do you know the answer to this question?

rhyolight: This `sp_plotter` is expecting a path of `~/Desktop/ExperimentResults` to exist. I don't think this will ever work out of the box.

@david-ragazzi I think if you create the `~/Desktop/ExperimentResults` directory, you might be able to run the examples.

scottpurdy:
@david-ragazzi - a lot of good changes here! A few things after Chetan asked me about this:

- Can we separate the PR for removing the flat spatial pooler into a separate PR. That will be much simpler and easy to review/merge.
- The changes seem to be causing a lot of degradations in test result metrics. Why is this? Is this the result of changing from FDRCSpatial2 to the new spatial pooler? We don't expect this to happen so we need to find out why. Do you have any idea what is causing the degradations?

examples/opf/tools/sp_plotter.py
david-ragazzi: > @david-ragazzi I think if you create the ~/Desktop/ExperimentResults directory, you might be able to run the examples. 

It didn't work. Actually `Basil100x21.csv` already must exists because it is a input data file. I didn't find this file in anywhere! Is someone using this example? It seems a bit forgoten in repo.

david-ragazzi:
> - Can we separate the PR for removing the flat spatial pooler into a separate PR. That will be much simpler and easy to review/merge.

Yes, done: https://github.com/numenta/nupic/pull/2056

> - The changes seem to be causing a lot of degradations in test result metrics. Why is this? Is this the result of changing from FDRCSpatial2 to the new spatial pooler? We don't expect this to happen so we need to find out why. Do you have any idea what is causing the degradations?

Yes, this `the result of changing from FDRCSpatial2 to the new spatial pooler`. Probably some extra members should be explicitly initialized in the script. The problem is: "which members and which values?" 

Taking in account that these tests were pointless (until now) as you guys deprecated `FDRCSpatial2`, we have these options:
- Accept the new metrics returned by the same tests using the new Spatial Pooler. In other words, leave as is.
- Review these tests testing which missing members and their values could reach the original metrics.
- Rewrite the tests using new values and new metrics.

scottpurdy:
@david-ragazzi - It is disappointing to see you labelling code/tests as "useless." The FDRCSpatial2 class was THE spatial pooling implementation for a long time. We were fairly confident the new spatial pooler should have the same performance but when we see all of the tests get worse results that should be a big red flag. Now we need to see if we can simply update the params to get similar results to old implementation or if there is perhaps a regression in performance with the new SP.

david-ragazzi:
@scottpurdy I think you misunderstood me. I didn't mean that FDRCSpatial2 was useless at all! I meant that as the _tests_ are still measuring the old SP and not the old one, then they, in certain way,  weren't helping us to get confident that the new SP was making a good work..

scottpurdy:
> I meant that as the tests are still measuring it and not the new SP, then these tests, in certain way, weren't helping us to get confident that the new SP was making a good work..

I agree! So switching them to the new one rather than deleting them is great. But we need to make sure that the new one actually works well on them.

david-ragazzi:
> I agree! So switching them to the new one rather than deleting them is great. But we need to make sure that the new one actually works well on them.

Maybe neither all tests need be reviewed but only those with a significant performance difference.

david-ragazzi:
@scottpurdy Now seeing again the changes, I remember the `inputBorder` parameter (present in FDRCSpatial2 but not in the new SP) was not set simply because the new SP has not correspondent parameter (https://github.com/numenta/nupic/pull/1872/files#diff-dd62c22371c4ff274c9468440dd01d03L232). Most tests with divergences passed this parameter to the FDRCSpatial2 constructor, but not longer.

scottpurdy:
> Maybe neither all tests need be reviewed but only those with a significant performance difference.

Agreed that this would help us understand the issue and most likely all we need.

> Or even the problem is not with new SP: as I said before, my feeling is that maybe some missing parameters (remember that new SP introduced new parameters) need be calibrated to the test reach a same or similar value to the original.

Yes, this seems likely. Now we just have to figure out which parameters to fiddle with.

> Now seeing again the changes, I remember the inputBorder parameter (present in FDRCSpatial2 but not in the new SP) was not ported simply because the new SP has not correspondent parameter, or at least, I did't know which would be this (https://github.com/numenta/nupic/pull/1872/files#diff-25fa3ff923c02d1cc17f48830bfbb5dfL111). Most tests with divergences passed this parameter to the FDRCSpatial2 constructor, but not longer with the new SP.

That is a good clue. So next step is to figure out what this parameter did and if there is an equivalent? I'll try to dig in when I have a chance.

scottpurdy:
So the definition of inputBorder is here:
https://github.com/numenta/nupic/blob/de39fcceb9acd5983f2dfc8e9af8cf4554c79737/nupic/research/FDRCSpatial2.py#L126

I don't see any equivalent in the new spatial pooler. Maybe we need to add it back in? @subutai - thoughts?

subutai:
> So the definition of inputBorder is here:
> https://github.com/numenta/nupic/blob/de39fcceb9acd5983f2dfc8e9af8cf4554c79737/nupic/research/FDRCSpatial2.py#L126
> 
> I don't see any equivalent in the new spatial pooler. Maybe we need to add it back in? @subutai - thoughts?

So this is tied to having a topology. It determines how you situate columns over an input that has topology (such as an image). It could potentially be handled in other (perhaps cleaner) ways. I would recommend we not port it now until we know we need it. OK to delete any tests that tested this feature.

examples/opf/experiments/anomaly/temporal/simple/description.py
```diff
@@ -195,8 +195,6 @@
             'synPermActiveInc': 0.1,
 
             'synPermInactiveDec': 0.01,
-
-            'randomSP': 1
```
breznak: :+1:  was discussed in some other PR, this translates mostly to `learn=False in SP.compute()`

examples/opf/tools/sp_plotter.py
```diff
@@ -226,16 +226,15 @@ def testSP():
   while numRecords < 3:
     
     # Setup a SP
-    sp = FDRCSpatial2.FDRCSpatial2(
-           coincidencesShape=(2048, 1),
-           inputShape = (1, elemSize),
-           inputBorder = elemSize/2-1,
```
breznak: :+1:  I came across some documentation on this one a while ago, as @subutai explained, it was used for image processing, where 2D topology and "lonely columns at the borders" played part. It is not used in newer SP implementations.

examples/opf/tools/sp_plotter.py
```diff
@@ -260,10 +259,11 @@ def testSP():
       for i in xrange(inputSize):
         time.sleep(0.001)
         if iter == numIter - 1:
-          outputs[i] = sp.compute(inputs[i], learn=doLearn, infer=False)
+          sp.compute(inputs[i], learn=doLearn, activeArray=outputs[i])
```
breznak: btw, OT for this PR, but different implementations (cpp/py of SP/TP) behave inconsistently, 
sometimes it's:
* `output = xx.compute(input, params)`
* `xx.compute(input, params, output)`

For interoperability of the implementations, we should unify that. (as I come from java, i'd prefer the first approach, no biggie though)

nupic/bindings/algorithms.i
```diff
@@ -794,269 +691,6 @@ void forceRetentionOfImageSensorLiteLibrary(void) {
...
-      self->load(s);
-      return true;
-    } else {
-      throw std::runtime_error("Failed to load FDRCSpatial");
-      return false;
-    }
-  }
-}
-
-//--------------------------------------------------------------------------------
```
breznak: nice to see this go, @david-ragazzi !

nupic/regions/SPRegion.py
```diff
@@ -106,15 +103,6 @@ def _buildArgs(f, self=None, kwargs={}):
         argValue = argTuple[2]
       # Set as an instance variable if 'self' was passed in
       setattr(self, argName, argValue)
-      
-  # Translate some parameters for backward compatibility
-  if kwargs.has_key('numActivePerInhArea'):
-    setattr(self, 'numActiveColumnsPerInhArea', kwargs['numActivePerInhArea'])
-    kwargs.pop('numActivePerInhArea')
```
breznak: maybe write a comment somewhere, eg in `SpatialPooler` that 
"""
# Old params used in obsoleted FDRCSpatialPooler were: 
# deprecated FDRCSpatial -> current SP
# numActivePerInhArea -> numActiveColumnsPerInhArea
# InputBorder -> (unused)
# ...
"""

If someone had a (quite) old script and needed porting to a recent nupic.

nupic/regions/SPRegion.py
```diff
@@ -470,50 +452,16 @@ def _allocateSpatialFDR(self, rfInput):
...
       )
-      
-    else:
-      # Backward compatibility
-      autoArgs.pop('coincidencesShape')
-      autoArgs.pop('inputShape')
-      autoArgs.pop('inputBorder')
-      autoArgs.pop('coincInputRadius')
-      autoArgs.pop('cloneMap')
-      autoArgs.pop('numCloneMasters')
```
breznak: see my comment above, these are the params to document for compatibility reasons.

examples/opf/tools/sp_plotter.py
david-ragazzi: Agreed. I also think the first approach is better: it is more descriptive and explicit using a method return then simply passing reference parameters. The problem in the second approach is that you have to check inner and outer scope to figure out *which* value should be returned.

examples/opf/experiments/anomaly/temporal/simple/description.py
david-ragazzi: > :+1: was discussed in some other PR, this translates mostly to learn=False in SP.compute()

Done.

nupic/regions/SPRegion.py
david-ragazzi: It's ok. Maybe it's better we send a message to ML.

david-ragazzi:
@breznak

Replacing `RandomSP` to `learn` didn't work:

https://travis-ci.org/numenta/nupic/jobs/59119243#L3460

Actually `learn` is a parameter of `SP.compute()` (https://github.com/numenta/nupic/blob/master/nupic/research/spatial_pooler.py#L673), not `SP.__init__()` (https://github.com/numenta/nupic/blob/master/nupic/research/spatial_pooler.py#L52).

I'm reverting..

breznak:
@david-ragazzi sorry for misunderstanding, ^^^ that's what I meant. 
Please see https://github.com/numenta/nupic.core/issues/350 for last comments and then all those issues can be closed.

david-ragazzi:
@scottpurdy @subutai So.. can this PR being merged?

tests/unit/nupic/research/sp_learn_inference_test.py
```diff
@@ -138,6 +140,8 @@ def _runLearnInference(self,
       print "Test succeeded"
 
 
+  @unittest.skip("Currently fails due to switch from FDRCSpatial2 to SpatialPooler. \
+                 The new SP doesn't have explicit methods to get inference.")
```
scottpurdy: Change from backslash:
```
@unittest.skip("Currently fails due to switch from FDRCSpatial2 to SpatialPooler. \
               The new SP doesn't have explicit methods to get inference.")
```

to implicit string concatenation:

```
@unittest.skip("Currently fails due to switch from FDRCSpatial2 to SpatialPooler. "
               "The new SP doesn't have explicit methods to get inference.")
```

tests/unit/nupic/research/sp_learn_inference_test.py
```diff
@@ -108,7 +109,8 @@ def _runLearnInference(self,
       if spVerbosity > 0:
         print "Input #%d" % i
       encodedInput = inputs[i]
-      spLearnInfer.compute(encodedInput, learn=True, infer=False)
+      decodedOutput = np.zeros(columnDimensions)
+      spLearnInfer.compute(encodedInput, learn=True, activeArray=decodedOutput)
```
scottpurdy: Please file an issue for allowing `activeArray` to be `None` to allow no-inference calls and add a comment here (and anywhere else you had to do this) with a link to the issue.

tests/unit/nupic/research/inhibition_object_test.py
```diff
@@ -71,26 +73,25 @@ def testInhibition(self):
...
-                                inputBorder = inputLen/2 - 1,
-                                coincInputRadius = inputLen / 2,
-                                numActivePerInhArea = numActivePerInhArea,
-                                spVerbosity = spVerbosity,
-                                stimulusThreshold = stimulusThreshold,
-                                seed = spSeed
-                                )
+    spTest = SpatialPooler(
+                           columnDimensions=(columnDimensions, 1),
+                           inputDimensions = (1, inputLen),
```
scottpurdy: No spaces around `=` for keyword args here and elsewhere (I realize this is a remnant from old code but might as well fix while you're touching these lines).

tests/unit/nupic/research/inhibition_object_test.py
```diff
@@ -39,10 +39,12 @@
 
 class InhibitionObjectTest(unittest.TestCase):
 
+  @unittest.skip("Currently fails due to switch from FDRCSpatial2 to SpatialPooler. \
```
scottpurdy: Use implicit string concat rather than backslash as suggested below. Also please create an issue for adding method for getting inhibition or whatever is necessary to get this test working and put link in this comment.

examples/opf/experiments/anomaly/temporal/simple/description.py
scottpurdy: How did you make this experiment specify `learn=False` to SP compute?

scottpurdy:
Done with this pass. Please also resolve Travis issues and then ping me to look again.

tests/unit/nupic/research/inhibition_object_test.py
david-ragazzi: > Use implicit string concat rather than backslash as suggested below. Also please create an issue for adding method for getting inhibition or whatever is necessary to get this test working and put link in this comment.

Done.

tests/unit/nupic/research/inhibition_object_test.py
david-ragazzi: > No spaces around = for keyword args here and elsewhere (I realize this is a remnant from old code but might as well fix while you're touching these lines).

Done (in this and other SP initializations of this PR).

tests/unit/nupic/research/sp_learn_inference_test.py
david-ragazzi: > Please file an issue for allowing activeArray to be None to allow no-inference calls and add a comment here (and anywhere else you had to do this) with a link to the issue.

Done.

tests/unit/nupic/research/sp_learn_inference_test.py
david-ragazzi: > Change from backslash:
>to implicit string concatenation:

Done.

examples/opf/experiments/anomaly/temporal/simple/description.py
david-ragazzi: > How did you make this experiment specify learn=False to SP compute?

Actually I didn't. Look the discussion in this PR, specially the Subutai comment:

https://github.com/numenta/nupic/pull/1851

From I understood, as OPF tests pass `learn` parameter to `compute()`, then set randomSP is pointless..

david-ragazzi:
> Done with this pass. Please also resolve Travis issues and then ping me to look again.

@scottpurdy I answered your questions and implemented suggestions. 

(Travis is red only for OS X, I think because some external problem)

david-ragazzi:
@scottpurdy ping..

tests/unit/nupic/research/sp_learn_inference_test.py
scottpurdy: Can you put links in the code wherever there used to be `infer-False` so that it is possible to fix up all these cases after implementing the feature? Otherwise it would be next to impossible to find all the different places that aren't using the output.

scottpurdy:
Thanks @david-ragazzi - Just the one remaining issue:
https://github.com/numenta/nupic/pull/1872#discussion_r29170310

david-ragazzi:
> Thanks @david-ragazzi - Just the one remaining issue:
#1872 (comment)

Done. Once green can I merge it?

david-ragazzi:
> Thanks @david-ragazzi - Just the one remaining issue:
#1872 (comment)

Done. Can it be merged?

scottpurdy:
:+1: once green

scottpurdy:
I tried restarting the failed job, hopefully that will resolve it.

david-ragazzi:
> I tried restarting the failed job, hopefully that will resolve it.

Thanks.

oxtopus:
BTW, this was helpful in resolving an internal issue related to this change:

```
find REDACTED -name "*.REDACTED" -exec sed -i .bak -e 's/coincInputPoolPct/columnDimensions/g' -e 's/inputShape/inputDimensions/g' -e 's/coincInputRadius/potentialRadius/g' -e 's/coincInputPoolPct/potentialPct/g' -e 's/numActivePerInhArea/numActiveColumnsPerInhArea/g' {} \;
```

PR #1911, opened 2015-03-05
Fix broken build (just update SHA)

david-ragazzi:
Fixes https://github.com/numenta/nupic.core/issues/390


rhyolight:
I don't see how this will work, because it's the same version of `nupic.core` that was not working... I get the same error when building.

david-ragazzi:
>I don't see how this will work, because it's the same version of nupic.core that was not working... I get the same error when building.

I fixed this. Please check the SHA again.. Now it is pointing to:

https://github.com/numenta/nupic.core/commit/88a02a9730eca16d51c8107acee0dea8c09adbee

Issue #1920, opened 2015-03-12
Tests are being installed in an unexpected location


oxtopus:
In an Ubuntu 14.04.1 LTS environment I ran:

`pip install https://s3-us-west-2.amazonaws.com/artifacts.numenta.org/numenta/nupic/releases/nupic-0.2.1-cp27-none-linux_x86_64.whl`

NuPIC, and friends, installed just fine:

```
Successfully installed DBUtils-1.1 PyMySQL-0.6.2 PyYAML-3.10 asteval-0.9.1 astroid-1.3.5 cov-core-1.15.0 coverage-3.7.1 execnet-1.3.0 logilab-common-0.63.2 mock-1.0.1 numpy-1.9.2 nupic-0.2.1 ordereddict-1.1 pillow-2.3.0 prettytable-0.7.2 psutil-1.0.1 py-1.4.26 pycapnp-0.5.1 pylint-1.1.0 pyproj-1.9.3 pytest-2.4.2 pytest-cov-1.6 pytest-xdist-1.8 python-dateutil-2.1 six-1.9.0 tweepy-2.1 unittest2-0.5.1 validictory-0.9.1
```

Tests are installed into `/usr/local/lib/python2.7/dist-packages/tests/`:

```
# ls /usr/local/lib/python2.7/dist-packages/tests/integration/nupic/
__init__.py   config.py   external     mock.py   swarming     test_api.pyc  test_auth.pyc    test_cursors.pyc   test_resultset.pyc  test_streaming.pyc
__init__.pyc  config.pyc  integration  mock.pyc  test_api.py  test_auth.py  test_cursors.py  test_resultset.py  test_streaming.py   unit
```

```
# ls /usr/local/lib/python2.7/dist-packages/tests/integration/nupic/
__init__.py  __init__.pyc  algorithms  data  engine  opf  regions
```

Tests fail for a myriad of reasons.  Full test output at https://gist.github.com/oxtopus/0ed17d496ad19ae07dab

rhyolight:
Thanks for filing. I've never tried to run tests from a binary installation, honestly I never expected binary installers to attempt to run tests. 

Do you think it's a good idea to include the test binaries in the package? Most software deployments I've done in the past for other projects have not included tests like this.

oxtopus:
[...]

PR #1922, opened 2015-03-12
Run OPF experiment without source checkout

### LINKED ISSUE #1921: Run OPF experiment from command line without "scripts" installed.
Some python modules expose command-line utilities by allowing the user to invoke the module directly.  For example: `echo '{"foo":"bar"}' | python -m json.tool` and `python -m SimpleHTTPServer`, I'd like to be able to run an OPF experiment with only the binary installation.  Currently, the only way to invoke a single OPF experiment is to also clone the repository to gain access to `scripts/run_opf_experiment.py`, or replicate the functionality of that script separately.

oxtopus:
Shift bulk of run_opf_experiment.py into experiment_runner.py for cli-friendly python module.  This lets you run an experiment by invoking `python -m nupic.frameworks.opf.experiment_runner <path to experiment>` with only the binary installation and without needing access to scripts/run_opf_experiment.py (not included in binary).

Meanwhile, this maintains existing functionality of run_opf_experiment.py for non-binary installations.

See discussion in https://github.com/numenta/nupic/issues/1920#issuecomment-78571236 for motivation.

Fixes #1921

### DIFF
--- a/nupic/frameworks/opf/experiment_runner.py
+++ b/nupic/frameworks/opf/experiment_runner.py
@@ -41,6 +41,8 @@ from nupic.frameworks.opf.modelfactory import ModelFactory
 from nupic.frameworks.opf.opftaskdriver import OPFTaskDriver
 from nupic.frameworks.opf.opfutils import (InferenceElement, matchPatterns,
                                            validateOpfJsonValue)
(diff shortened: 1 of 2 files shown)


rhyolight:
:+1:

Issue #1300, opened 2014-09-13
Anomaly ignoring unfulfilled predictions


breznak:
"When you open the door and find an elephant in your room, it's an anomaly by current code; 
If you walk home and your house is not there - it should be an anomaly as well (but it's not now)." 

We treat an unexpected event that occurred as an anomaly, but not an expected event that did not occur. 

On the low level, we use (super-)set of predicted columns and treat active columns with no "predicted parent" as anomalous, while false-positives predictions are let ignored. 
For more details see https://github.com/numenta/nupic/wiki/Anomaly-Detection-and-Anomaly-Scores#computation 

I suggest: 
1/ changing `Anomaly` class to accept array of permanences of columns' as an input, so experiments can be implemented easily. 
2/ changing anomaly score so that `false positives` (predictions) are treated as an anomaly as well.
3/ changing anomaly score to use only the "true" predicted columns (not any column with permanence >0 as now) as "predicted column". That makes sense after 2/ is in place. 

Expected results:
-improved anomaly detection (added the other half of features - expected but missing columns)
-improved swarming on anomaly models (if it takes anomaly score in account(?))

I know this is a bit bigger cut, but changes to the code should be minimal and we could take advantage of the new anomaly class to implement different "versions" of anomaly. What do you guys think?

CC @subutai @scottpurdy 

---

TODO:
test more (on NAB) and revert 
breznak@3889fae
breznak@88bd6c6
when needed.

breznak:
TODO: test more (on NAB) and revert 
https://github.com/breznak/nupic/commit/3889fae55147b01e24c34454243ac109088c0111
https://github.com/breznak/nupic/commit/88bd6c6acf29cfeffc39afcbb48c795e747c3df2
when needed.

Issue #1841, opened 2015-02-20
AnomalyMetric for optimization of anomaly params during swarming (expected anomaly %)


breznak:
Currently, it is possible to swarm over various anomaly settings (`slidingWindowSize, mode, `) but it is actually pointless, as even the `TemporalAnomaly` model type optimizes for best prediction, the anomaly scores are nowhere taken to account in metric computation. 

I suggest tackling this by a new metric that would allow you to say "in this dataset, I expect anomaly to occur in about 1% of samples": `AnomalyMetric(windowSize=1000, expectedOccurence=0.01)` 

This would: 
* give a meaning to the anomaly params during swarming
* allow user to adjust how much anomalies they detect (sensitivity) in their data -- this is a common issue

breznak:
blocked by #1842

breznak:
Can I somehow access `anomalyScore` from OPF in `description.py`, or the actual Anomaly object ?

breznak:
I thought I could set `inferenceElement=anomalyScore` : 

`MetricSpec(field=u'ecg', metric='multiStep', inferenceElement='anomalyScore', params={'window': 101, 'steps': [1], 'errorMetric': 'rmse'})`

but that does not seem to work either

rhyolight:
@scottpurdy / @chetan51 / @subutai help?

subutai:
I think this is an interesting research area (figuring out a better metric for anomaly detection). Currently I optimize for prediction error, and then tune manually. And then tune manually some more. :smile: 

I am not familiar enough with the Metrics code to suggest a specific implementation though.

rhyolight:
I'm categorizing this as an enhancement instead of a bug.

PR #1301, opened 2014-09-13
Anomaly strict

breznak:
Work towards #1300 
Currently implements that also "predicted but not found" patterns contribute to anomaly score. 

TODO:
- [ ] 1/ change Anomaly class to take array of columns' permanences as input
- [ ] 1.2/ Then anomaly class can be a stateful machine and can take only 1 argument (actual permances) and and keep track of the "prevPredictedColumns" itself.
- [ ] 2/ revert `computeRawAnomalyScore` back and add a new `MODE_` for the new anomaly, vs. having the one and only anomaly computation method?
- [ ] 3/ use only the true predicted columns (permanence>=threshold) as "predicted" columns.

Please let me know what you think about the TODO items. 
:1234:

Issue #1236, opened 2014-08-22
Refactor swarming into standalone library


scottpurdy:
The swarming code is tightly coupled to the OPF but could be very useful elsewhere. It should be a library that has two main operations for a set of permute objects - ask for new parameter set, and give a score for a previously given parameter set.

The design hasn't been fully aligned with the current codebase. The current codebase makes assumptions throughout that tie it to the OPF and DAO. It is really important that we don't have regressions in functionality and it will likely be a pretty significant initiative to organize the code in a way so that we retain existing OPF swarming functions while providing an interface for other clients.

---
- [x] [Investigate what tasks are necessary to complete swarming as a library](https://github.com/numenta/nupic/issues/1753)
- [x] Create sub-tasks here from work in https://github.com/numenta/nupic/issues/1753
- [x] [Hypersearch Decoupling Milestone 1: Separate swarming code into core Hypersearch files and NuPIC-specific Client ](https://github.com/numenta/nupic/issues/2391)
- [ ] [Hypersearch Decoupling Milestone 2: Define the Hypersearch API ](https://github.com/numenta/nupic/issues/2393)
- [ ] [Hypersearch Decoupling Milestone 3: Migrate Hypersearch API to new design ](https://github.com/numenta/nupic/issues/2394)
- [ ] [Hypersearch Decoupling Milestone 4: Deploy Hypersearch to its own PyPI project and include in NuPIC requirement.txt ](https://github.com/numenta/nupic/issues/2395)
- [ ] [Hypersearch Decoupling Milestone 5: Move Hypersearch out of NuPIC into stand-alone repo ](https://github.com/numenta/nupic/issues/2396)

rhyolight:
As we talked about in our [last review meeting](https://www.youtube.com/watch?v=9Bkt-OC2vV8), this issue has been moved into milestone 0.3. The title needs to be updated and more granular subtasks should be created. @scottpurdy, please update the title of this ticket accordingly.

breznak:
pasting a relevant comment by @rcrowder 

>I wonder whether parameter tuning could interface to existing packages that do a similar job, such as Neurofitter (http://www.ncbi.nlm.nih.gov/pubmed/18974796?dopt=Abstract&holding=f1000,f1000m,isrctn). Maybe NeRvolver, http://www.biomedcentral.com/1471-2202/13/S1/P36 Interfacing with HTM maybe a large a hurdle that isn't necessary, but good to see how others have approached this issue.
_
[...]

PR #1934, opened 2015-03-18
Modular Anomaly.computeRAWAnomalyScore() 

### LINKED ISSUE #1980: Allow different methods to compute anomalyRAWScore and swarm over them
Provide "framework" to easily switch (and optimize over it) `anomaly.compute()` methods. 
See #1934 for reference, but these methods need tougher evaluation to justify them, this PR implements just the building blocks for it.

breznak:
Fixes #1980 

The PR adds new ways how to compute "raw" anomaly score from TP. 
And allows to swarm over it.

----
EDIT: was intended for #1300 but needs more testing there.


breznak:
@subutai this is the topic we've talked a while ago, allows easy experimentation and with 
#1841 for swarming. 

@oxtopus @scottpurdy please take a look, what I need to fix is `failing picke` because of the functions. 
Otherwise the unittests & hotgym example run fine.

rhyolight:
Closing and re-opening to try to convince GitHub to let me reopen #1301.

rhyolight:
I need more sleep. Never mind me.

rhyolight:
@breznak Marek this PR is marked as "fixes" another PR, which doesn't make sense. Is there a real issue it should be fixing? If not, can you make one with a good description of the problem this PR is solving? Thanks.

breznak:
@rhyolight my bad, i need sleep too :zzz: #1300 is the issue, #1301 was some old PR on the same topic.

breznak:
@cogmission since I saw your interest in `computeRawAnomalyScore()` issues, I'd like if you could take a look on this PR

cogmission:
@breznak normally I would be honored to do so! :)

Right now, I am up to my eyebrows in stuff to do. I'm working around the clock on the java version's Network API - and then there are other things I would like to get to to ready htm.java for real production use.

I can help clarify issues and answer community questions (on both nupic and htm.java) but - that's it for the next probably couple of months. I want to improve my Python skills though... so don't forget about me in the future!

rhyolight:
Assigned to @scottpurdy for review, but he's OOTO until Wednesday. @chetan51 Feel free to take a look if you have time.

examples/opf/experiments/anomaly/temporal/hotgym/description.py
```diff
@@ -44,7 +44,7 @@
                                             IterationPhaseSpecLearnOnly,
                                             IterationPhaseSpecInferOnly,
                                             IterationPhaseSpecLearnAndInfer)
-
+from nupic.algorithms import anomaly
```
breznak: this can go away once #1948 is in.

breznak:
@scottpurdy moving our discussion here, where it belongs

> > It might be worse in some cases, and better in others

>Perhaps I shouldn't jump to conclusions :) but Numenta folks did a good amount of work to make a robust anomaly detection solution so I think it would require really convincing experimental results to be considered for a change.

Ok, I see your concern about adding useless code to the repo. I will need to do more experiments to backup my claims then. Can we go like this: 
- I will remove the 2 new methods from the PR
- the vital code for experimenting with alt. anomaly compute methods (swarming) stays (core of this PR), so it's convenient for me/or others to carry such experiments
- Do we have "official" datasets for anomaly? So I can say here method XY is this better, but there fails. ? Like NAB, but I'm not sure if there's such data? Ideally we need datasets with different kinds of anomalies. CC @subutai

subutai:
@breznak I think the NAB dataset is the best on we have. You could run it before and after your change and see if the score gets better or worse.

breznak:
@scottpurdy I separated the problematic (new `computeFn()` parts), is this ok?

rhyolight:
@scottpurdy ping

breznak:
ping?

breznak:
pong?

scottpurdy:
@breznak - Sorry for the delay on this. Haven't had a lot of time to dig into it but will try to follow up more thoroughly tomorrow.

scottpurdy:
@breznak - similar to metrics, let's just pass the model to the Anomaly instance and then your various implementations can pull whatever data they want for the calculation. Does that work?

And please leave the raw anomaly function as it is.

By the way, the direction I plan to move here is allowing you to create your own anomaly region implementation. Currently the calculation is outside the network but I'd like to change that with #2059. There are still some things that would limit your ability to write your own though, such as #2057 and #1236 if you want to swarm. As such, I think just passing the model in to the anomaly class is the minimally-invasive option for now that let's you achieve whatever experimentation you want.

breznak:
@scottpurdy 
> similar to metrics, let's just pass the model to the Anomaly instance and then your various implementations can pull whatever data they want for the calculation. Does that work?

I reviewed and like your PR on AnomalyRegion, but this PR is not about "accessing any special data"(from a model), but about "implementing new ways(algorithms) to do `anomaly.compute()`" and swarming over them; the mentioned PRs address different issues. 

> And please leave the raw anomaly function as it is.

The method is exactly the same, I just added some documentation and moved it to the bottom (as before, there were more implementations in this PR) to place them together.

breznak:
@scottpurdy @rhyolight please review? I made the changes really minimally invasive, but allowing me or others to easily experiment with alternative anomaly implementations and swarming over them. This is not in conflict with the AnomalyRegion PR. I short, it allows you to swarm over `anomaly.computeFn` which can now be set as a param (a function is passed by param)

nupic/frameworks/opf/clamodel.py
```diff
@@ -190,8 +191,11 @@ def __init__(self,
     windowSize = anomalyParams.get("slidingWindowSize", None)
     mode = anomalyParams.get("mode", "pure")
     anomalyThreshold = anomalyParams.get("autoDetectThreshold", None)
-    self._anomalyInst = Anomaly(slidingWindowSize=windowSize, mode=mode,
-                                binaryAnomalyThreshold=anomalyThreshold)
+    computeFn = anomalyParams.get("computeFn", anomaly.compute_In1D_Satisfied) # default, aka "Raw"
```
rhyolight: Nit: move comment above this line.

rhyolight:
Theoretically, this seems like a decent change to me. @scottpurdy ?

breznak:
@rhyolight sorry to be insistent :wink: but this PR is getting stale..

scottpurdy:
Moving how-to-implement discussion to the issue #1980

breznak:
@rhyolight Travis sucked up https://travis-ci.org/numenta/nupic/jobs/63234775#L5

rhyolight:
@breznak Restarted.

breznak:
is the AppVeyor expected to work, or not yet?

nupic/algorithms/anomaly.py
```diff
@@ -174,3 +199,41 @@ def __setstate__(self, state):
...
     if not hasattr(self, '_binaryThreshold'):
       self._binaryThreshold = None
+    if not hasattr(self, '_computeFn'):
+      self._computeFn = computeRawAnomalyScore
+
+  ##################################################################
+  # these methods are alternative implementations of 'how to compute'
+  # anomaly score from the neural output (2 binary vectors)
+  ##################################################################
+def compute_In1D_Satisfied(activeColumns, prevPredictedColumns):
```
scottpurdy: I don't like moving the raw anomaly score here or changing the name to this (plus this name doesn't match camel case convention).

examples/opf/experiments/anomaly/temporal/hotgym/description.py
```diff
@@ -317,6 +313,8 @@
         'anomalyParams': {
            'mode': 'likelihood', # pure(=default) / weighted / likelihood
            'slidingWindowSize': 5, # >=0 / None
+           'binaryAnomalyThreshold': None, # None or from (0,1)
+           'computeFn': Anomaly.COMPUTE_RAW,
```
scottpurdy: I don't think we should add this since it is the default. Feel free to create an example specifically for custom compute functions though.

nupic/algorithms/anomaly.py
```diff
@@ -22,34 +22,43 @@
...
 import numpy
+import collections
 
 from nupic.algorithms.anomaly_likelihood import AnomalyLikelihood
 from nupic.utils import MovingAverage
 
 
 def computeRawAnomalyScore(activeColumns, prevPredictedColumns):
-  """Computes the raw anomaly score.
+    """Computes the raw anomaly score.
```
scottpurdy: The indentation was changed to 4 spaces for the body of this function but it should be left at 2

nupic/algorithms/anomaly.py
```diff
@@ -22,34 +22,43 @@
...
-    # array. Sum to get the total # of columns that are active and were
-    # predicted.
-    score = numpy.in1d(activeColumns, prevPredictedColumns).sum()
-    # Get the percent of active columns that were NOT predicted, that is
-    # our anomaly score.
-    score = (nActiveColumns - score) / float(nActiveColumns)
-  else:
-    # There are no active columns.
-    score = 0.0
+    The implementation is using in1D() function,
```
scottpurdy: This paragraph should be removed.

rhyolight:
@breznak said:
> is the AppVeyor expected to work, or not yet?

It's broken, and I can't figure out how to turn off the notifications. If you see an AppVeyor failure, ignore it.

rhyolight:
If you see an AppVeyor failure on a `nupic.core` PR, that is valid.

nupic/algorithms/anomaly.py
```diff
@@ -22,34 +22,43 @@
...
-    # our anomaly score.
-    score = (nActiveColumns - score) / float(nActiveColumns)
-  else:
-    # There are no active columns.
-    score = 0.0
+    The implementation is using in1D() function, 
+    "satisfied" means it only cares if activeColumns at time T have been 
+    predicted at T-1. 
+
+    Which means a temporal pooler TP with all columns in predictive state at T-1
```
scottpurdy: This reads a little funny. I'm not sure if the intent is to help users understand the computation better or point out this particular use case. Perhaps this wording?

"If all columns were predicted at previous time step (prevPredictedColumns) then the score will be zero regardless of which columns, if any, are active currently (activeColumns)."

Specifically this removes the reference to "temporal pooler TP" and the notion of T and T-1, which a programmer may not be familiar with.

scottpurdy:
@breznak - What do you think about this?
https://github.com/numenta/nupic/compare/numenta:master...scottpurdy:anomaly-custom-fn?expand=1

examples/opf/experiments/anomaly/temporal/hotgym/description.py
breznak: ok, then I'll just remove it here, in some other PR, I have a new "anomaly-swarming" example, so I'll stuff all the experimental anomaly functionality there.

breznak:
> https://github.com/numenta/nupic/compare/numenta:master...scottpurdy:anomaly-custom-fn?expand=1 

@scottpurdy thank you for the reference implementation! but I don't thinks that's the intention I have on my mind.. it uses <func> as a param, which we said we don't want; does not allow to combine with the other anomaly params (they all can be combined and affect the resulting score); and I'm not sure how would that integrate with swarming..which is a goal for this PR. 

On the other hand, the parameters `mode` and `computeFn` are kind of similar - we already have 3 "different" `compute()` methods here https://github.com/numenta/nupic/blob/master/nupic/algorithms/anomaly.py#L130-L144 
So I'll turn them into real mini-functions, this will give us also a first demo on the different modes. Basically it would work as the `Anomaly.MODE_*` which is already present, but instead of `elseif` branches, it will be a function, so people can design their own ... but that brings us back, this would really work better if the param was a `func` instead of a string, can it?

breznak:
@scottpurdy the current state combines your suggested solution and the prev. functionality, it's actually quite nice - one can use custom compute() fn defined anywhere, and there are "named" functions for known functions, so there are easily used from OPF, swarming should work, and the parameters are reduced (`computeFn` merged with `mode/Anomaly.MODE_*`) 
What do you think?

examples/opf/experiments/anomaly/temporal/hotgym/description.py
```diff
@@ -44,7 +40,7 @@
                                             IterationPhaseSpecLearnOnly,
                                             IterationPhaseSpecInferOnly,
                                             IterationPhaseSpecLearnAndInfer)
-
+from nupic.algorithms.anomaly import Anomaly
```
david-ragazzi: Use alphabetical order for import statemets.

nupic/algorithms/anomaly.py
```diff
@@ -22,6 +22,8 @@
 """Anomaly-related algorithms."""
 
 import numpy
+import collections
```
david-ragazzi: Alphabetical order here too. `collections` should be above `numpy`

nupic/algorithms/anomaly.py
```diff
@@ -48,7 +56,6 @@ def computeRawAnomalyScore(activeColumns, prevPredictedColumns):
   else:
     # There are no active columns.
     score = 0.0
-
   return score
```
david-ragazzi: It is a good pratice put a blank line before `return`.

tests/unit/nupic/algorithms/anomaly_test.py
```diff
@@ -109,11 +109,25 @@ def testAnomalyCumulative(self):
...
     anomalyComputer = anomaly.Anomaly(mode=anomaly.Anomaly.MODE_PURE)
     score = anomalyComputer.compute(array([2, 3, 6]), array([3, 5, 7]))
     self.assertAlmostEqual(score, 2.0 / 3.0)
 
 
+  def testComputeAnomalyEmpty(self):
+    """anomaly called with empty params """
+    score = anomaly.computeRawAnomalyScore(array([]), array([]))
+    self.assertEqual(score, 0)
+
```
david-ragazzi: The nupic convention recommend 2 blank lines between methods.

nupic/algorithms/anomaly.py
```diff
@@ -157,6 +143,30 @@ def compute(self, activeColumns, predictedColumns,
...
+    """Anomaly computed using the anomaly_likelihood score, 
+       which models probability of (input, anomalyScore) pair.
+    """
+    assert isinstance(self._likelihood, AnomalyLikelihood)
+    if inputValue is None:
+      raise ValueError("Selected anomaly mode 'Anomaly.MODE_LIKELIHOOD' "
+                       "requires 'inputValue' as parameter to compute() method. ")
+    rawScore = computeRawAnomalyScore(active, prevPredicted)
+    probability = self._likelihood.anomalyProbability(inputValue, rawScore, ts)
+    return probability
```
david-ragazzi: I didn't understand why this is returning the probability and not the score (`score = 1 - probability`) as was done earlier (https://github.com/numenta/nupic/pull/1934/files#diff-ac4ff5014c0e81ed147be47c27c7bc7aL140). Are you sure that this won't cause problems in code that uses this function?

tests/unit/nupic/algorithms/anomaly_test.py
```diff
@@ -109,11 +109,25 @@ def testAnomalyCumulative(self):
 
 
   def testComputeAnomalySelectModePure(self):
+    """anomaly with selected mode (pure) """
```
david-ragazzi: Uppercase for the first letter in a sentence.

tests/unit/nupic/algorithms/anomaly_test.py
```diff
@@ -109,11 +109,25 @@ def testAnomalyCumulative(self):
...
     self.assertAlmostEqual(score, 2.0 / 3.0)
 
 
+  def testComputeAnomalyEmpty(self):
+    """anomaly called with empty params """
+    score = anomaly.computeRawAnomalyScore(array([]), array([]))
+    self.assertEqual(score, 0)
+
+  def testAnomalyCustomFn(self):
+    """anomaly using custom compute() function"""
```
david-ragazzi: Uppercase for the first letter in a sentence.

nupic/algorithms/anomaly.py
breznak: Well, it would change results, but I'm not sure how crucial (=how many ppl are using it) it is. Anyway, the break would be easy to fix. 
Motivation is that you can use the Anomaly class in Likelihood mode for the "original" anomaly_likelihood code which is used in more cases.

breznak:
Thanks for review @david-ragazzi , fixed

nupic/algorithms/anomaly.py
david-ragazzi: To me this sounds better (althought I'm not expert user of anomalies features). So I think @scottpurdy or @chetan51 should review such improvent.

examples/opf/experiments/anomaly/temporal/hotgym/description.py
```diff
@@ -20,13 +20,11 @@
...
 """
-Template file used by the OPF Experiment Generator to generate the actual
-description.py file by replacing $XXXXXXXX tokens with desired values.
-
-This description.py file was generated by:
-'~/nupic/eng/lib/python2.6/site-packages/nupic/frameworks/opf/expGenerator/ExpGenerator.py'
+Anomaly hotgym example
 """
 
+from nupic.algorithms.anomaly import Anomaly
```
vitaly-krugl: `Anomaly` imported here, but not used. Please run the changed files through pylint with nupic pylint config and fix pylint findings related to your changes.

nupic/algorithms/anomaly.py
```diff
@@ -167,10 +178,35 @@ def __str__(self):
...
+    """
+    if mode == Anomaly.MODE_PURE:
+      return self.computeRaw
+    elif mode == Anomaly.MODE_LIKELIHOOD:
+      self._likelihood = AnomalyLikelihood() # probabilistic anomaly
+      return self.computeLikelihood
+    elif mode == Anomaly.MODE_WEIGHTED:
+      self._likelihood = AnomalyLikelihood() # probabilistic anomaly
+      return self.computeWeighted
+    elif isinstance(mode, FunctionType):
```
vitaly-krugl: `isinstance(mode, FunctionType)` is not general enough. Use `callable` instead:

```
class Computer(object):
  @staticmethod
  def static_meth():
    return
  def instance_meth(self):
    return

isinstance(Computer.static_meth, FunctionType)
True
isinstance(Computer.instance_meth, FunctionType)
False
c = Computer()
isinstance(c.instance_meth, FunctionType)
False
from functools import partial
my_file = partial(file, "zzz")
isinstance(my_file, FunctionType)
False
callable(c.static_meth)
True
callable(c.instance_meth)
True
callable(my_file)
True

f = lambda: None
callable(f)
True
```

nupic/algorithms/anomaly.py
```diff
@@ -167,10 +178,35 @@ def __str__(self):
...
+  def getComputeFn(self, mode):
+    """
+    assing compute() function to be used; 
+    all these functions must take these arguments:
+     activeColumns, prevPredictedColumns, inputValue=None, timestamp=None
+    and return anomaly score
+
+    @param mode   if string: we assign from Anomaly.MODE_* defined functions
+                  if func: we will use that function to compute anomaly score
+    @return a function that is used for anomaly computation
```
vitaly-krugl: Better to say callable instead of function. Also, describe the inputs and outputs of this callable.

nupic/algorithms/anomaly.py
```diff
@@ -167,10 +178,35 @@ def __str__(self):
...
+    if mode == Anomaly.MODE_PURE:
+      return self.computeRaw
+    elif mode == Anomaly.MODE_LIKELIHOOD:
+      self._likelihood = AnomalyLikelihood() # probabilistic anomaly
+      return self.computeLikelihood
+    elif mode == Anomaly.MODE_WEIGHTED:
+      self._likelihood = AnomalyLikelihood() # probabilistic anomaly
+      return self.computeWeighted
+    elif isinstance(mode, FunctionType):
+      print "Anomaly: using custom compute() function"
```
vitaly-krugl: Remove the `print` statement. If you need to output something, use python logging instead.

nupic/algorithms/anomaly.py
```diff
@@ -167,10 +178,35 @@ def __str__(self):
...
+      return mode
+    else:
+      raise ValueError("Anomaly: computeFn has to be one of '%s' but is '%s' " % (Anomaly._supportedModes, mode) )
```
vitaly-krugl: `Anomaly._supportedModes` accesses a private member. `_supportedModes` should probably be a public method instead.

vitaly-krugl: The exception format text should be `"Anomaly: mode has to be a callable or one of '%s' but is '%s' "`

vitaly-krugl:
@breznak, I am reviewing the code now. I will post a final comment when I am done. Please refrain from making fixes until then. Thx!

nupic/algorithms/anomaly.py
```diff
@@ -21,7 +21,9 @@
 
 """Anomaly-related algorithms."""
 
+import collections
```
vitaly-krugl: `collections` is newly imported, but not used. Please run your changes through pylint with nupic pylint configuration and  fix findings related to your changes.

nupic/algorithms/anomaly.py
```diff
@@ -84,22 +93,15 @@ def __init__(self, slidingWindowSize=None, mode=MODE_PURE, binaryAnomalyThreshol
               models probability of receiving this value and anomalyScore
           - "weighted" - "pure" anomaly weighted by "likelihood"
               (anomaly * likelihood)
+          - OR type function and will be used directly to compute score
```
vitaly-krugl: "OR **callable** that will be used directly to compute score". Also, document the inputs and outputs of this callable that would enable one to use this without having the plough through the code to figure it out.

vitaly-krugl: Also, I am not crazy about overloading args, such as mode, to pass a function instead of a string. I think that if I was working on this, I would likely add a new optional kwarg: e.g., `computeAnomaly=None`. A user that wishes to pass his own computeAnomaly callable, would invoke the constructor with `mode=None` and `computeAnomaly=<user's callable>`

nupic/algorithms/anomaly.py
```diff
@@ -167,10 +178,35 @@ def __str__(self):
   def __setstate__(self, state):
     """deserialization"""
     self.__dict__.update(state)
-
     if not hasattr(self, '_mode'):
```
vitaly-krugl: Since the change allows `_mode` to have a callable as its value, what would be the implications on pickling and unpickling? Would it work properly in that case? Is there a test for it?

CC @scottpurdy

nupic/algorithms/anomaly.py
```diff
@@ -157,6 +144,30 @@ def compute(self, activeColumns, predictedColumns,
     return score
 
 
+############################################
```
vitaly-krugl: @scottpurdy: didn't we do away with separators like `###############################` in the nupic coding conventions?

nupic/algorithms/anomaly.py
```diff
@@ -157,6 +144,30 @@ def compute(self, activeColumns, predictedColumns,
...
+# implementations of compute() function
+  def computeRaw(self, active, prevPredicted, inputValue=None, timestamp=None):
+    return computeRawAnomalyScore(active, prevPredicted)
+
+  def computeLikelihood(self, active, prevPredicted, inputValue, ts=None):
+    """Anomaly computed using the anomaly_likelihood score, 
+       which models probability of (input, anomalyScore) pair.
+    """
```
vitaly-krugl: document inputs and outputs

nupic/algorithms/anomaly.py
```diff
@@ -157,6 +144,30 @@ def compute(self, activeColumns, predictedColumns,
     return score
 
 
+############################################
+# implementations of compute() function
+  def computeRaw(self, active, prevPredicted, inputValue=None, timestamp=None):
```
vitaly-krugl: document inputs and outputs

nupic/algorithms/anomaly.py
```diff
@@ -123,25 +126,9 @@ def compute(self, activeColumns, predictedColumns,
                               	(used in anomaly-likelihood)
     @return the computed anomaly score; float 0..1
```
vitaly-krugl: From @breznak's comment below regarding broken compatibility:
> Well, it would change results, but I'm not sure how crucial (=how many ppl are using it) it is. Anyway, the break would be easy to fix. Motivation is that you can use the Anomaly class in Likelihood mode for the "original" anomaly_likelihood code which is used in more cases.

What might be returned here is no longer compatible with the documentation `@return the computed anomaly score; float 0..1`. In some cases, this will be a score (e.g., raw anomaly score), in other cases it will be the probability (e.g., from the new `computeLikelihood`), and something else entirely when the user passes in a custom function for computing anomaly.

vitaly-krugl: @scottpurdy, what is your take on this break in return value compatibility?

nupic/algorithms/anomaly.py
```diff
@@ -157,6 +144,30 @@ def compute(self, activeColumns, predictedColumns,
...
+    """
+    assert isinstance(self._likelihood, AnomalyLikelihood)
+    if inputValue is None:
+      raise ValueError("Selected anomaly mode 'Anomaly.MODE_LIKELIHOOD' "
+                       "requires 'inputValue' as parameter to compute() method. ")
+    rawScore = computeRawAnomalyScore(active, prevPredicted)
+    probability = self._likelihood.anomalyProbability(inputValue, rawScore, ts)
+    return probability
+
+  def computeWeighted(self, active, prevPredicted, inputValue, timestamp=None):
```
vitaly-krugl: Need docstring with description of method and its inputs/outputs

nupic/algorithms/anomaly.py
```diff
@@ -157,6 +144,30 @@ def compute(self, activeColumns, predictedColumns,
...
+    prob = computeLikelihood(active, prevPredicted, inputValue, timestamp)
```
vitaly-krugl: Implement tests for all the builtin modes. A test should have failed in this case, since `computeLikelihood` doesn't exist at global scope.

nupic/algorithms/anomaly.py
scottpurdy: Yes. I usually don't complain if that is how the file is already but at some point we should clean them out.

nupic/algorithms/anomaly.py
```diff
@@ -157,6 +144,30 @@ def compute(self, activeColumns, predictedColumns,
...
+  def computeLikelihood(self, active, prevPredicted, inputValue, ts=None):
+    """Anomaly computed using the anomaly_likelihood score, 
+       which models probability of (input, anomalyScore) pair.
+    """
+    assert isinstance(self._likelihood, AnomalyLikelihood)
+    if inputValue is None:
+      raise ValueError("Selected anomaly mode 'Anomaly.MODE_LIKELIHOOD' "
+                       "requires 'inputValue' as parameter to compute() method. ")
+    rawScore = computeRawAnomalyScore(active, prevPredicted)
+    probability = self._likelihood.anomalyProbability(inputValue, rawScore, ts)
```
vitaly-krugl: No need to save the value in a local variable if the next statement simply returns it. More concise as: 

`return self._likelihood.anomalyProbability(inputValue, rawScore, ts)`

nupic/algorithms/anomaly.py
```diff
@@ -123,25 +126,9 @@ def compute(self, activeColumns, predictedColumns,
...
-
-      probability = self._likelihood.anomalyProbability(
-          inputValue, anomalyScore, timestamp)
-      # low likelihood -> hi anomaly
-      score = 1 - probability
-    elif self._mode == Anomaly.MODE_WEIGHTED:
-      probability = self._likelihood.anomalyProbability(
-          inputValue, anomalyScore, timestamp)
-      score = anomalyScore * (1 - probability)
+    score = self.getComputeFn(self._mode)(activeColumns, predictedColumns, inputValue, timestamp)
```
vitaly-krugl: The code shouldn't be repeating the `getComputeFn` computation every time, since `self._mode` is not changing after instantiation of the class. The computation function should be saved in a member variable during instantiation of the class, instead.

nupic/algorithms/anomaly.py
vitaly-krugl: `computeRaw`, `computeLikelihood`, and `computeWeighted` are private methods and must have a leading underscore in their names; e.g., `_computeWeighted`

nupic/algorithms/anomaly.py
```diff
@@ -167,10 +178,35 @@ def __str__(self):
...
-
     if not hasattr(self, '_mode'):
       self._mode = Anomaly.MODE_PURE
     if not hasattr(self, '_movingAverage'):
       self._movingAverage = None
     if not hasattr(self, '_binaryThreshold'):
       self._binaryThreshold = None
+
+
+  def getComputeFn(self, mode):
```
vitaly-krugl: The name of the function `getComputeFn` doesn't fully reflect what it does and that it is private; it doesn't just return compute function; there are side-effects in it, such as `self._likelihood = ...`. This function shouldn't exist at all; as discussed in one of my comments about the `compute` method, we shouldn't be calling it for every compute. The logic probably belongs in the constructor, where it also records the computation callable in a member variable for use by the `compute` method instead of calling `getComputeFn`.

nupic/algorithms/anomaly.py
```diff
@@ -167,10 +178,35 @@ def __str__(self):
...
+    and return anomaly score
+
+    @param mode   if string: we assign from Anomaly.MODE_* defined functions
+                  if func: we will use that function to compute anomaly score
+    @return a function that is used for anomaly computation
+    """
+    if mode == Anomaly.MODE_PURE:
+      return self.computeRaw
+    elif mode == Anomaly.MODE_LIKELIHOOD:
+      self._likelihood = AnomalyLikelihood() # probabilistic anomaly
```
vitaly-krugl: `self._likelihood` (and other member variables) should be defined in the constructor first, even if it's just something like `self._likelihood = None`, and should have a comment above it (in the constructor) describing what would be assigned to it.

nupic/algorithms/anomaly.py
vitaly-krugl: The logic needs tests

nupic/algorithms/anomaly.py
```diff
@@ -84,22 +93,15 @@ def __init__(self, slidingWindowSize=None, mode=MODE_PURE, binaryAnomalyThreshol
               models probability of receiving this value and anomalyScore
           - "weighted" - "pure" anomaly weighted by "likelihood"
```
vitaly-krugl: It would be better to fix the docstring to refer to `Anomaly.MODE_WEIGHTED`, etc., instead of the raw strings "weighted", "likelihood", etc.

tests/unit/nupic/algorithms/anomaly_test.py
```diff
@@ -109,11 +109,26 @@ def testAnomalyCumulative(self):
```
vitaly-krugl: Also need tests for weighted and likelihood.

tests/unit/nupic/algorithms/anomaly_test.py
```diff
@@ -109,11 +109,26 @@ def testAnomalyCumulative(self):
...
+    """Anomaly called with empty params """
+    score = anomaly.computeRawAnomalyScore(array([]), array([]))
+    self.assertEqual(score, 0)
+
+
+  def testAnomalyCustomFn(self):
+    """Anomaly using custom compute() function"""
+    def dummyCompute(active, pred, inputVal, timestamp):
+      return 0.1337
+    anomalyComputer = anomaly.Anomaly(mode=dummyCompute)
```
vitaly-krugl: Need to test pickling/unpickling and then using it.

nupic/algorithms/anomaly.py
scottpurdy: The likelihood probability is expressed as a fraction from 0.0 to 1.0, right? Wasn't it doing that before as well?

scottpurdy: Vitaly explained this to me.

@breznak - Since you integrated this class into the CLAModel, the output must still use the same likelihood logic. We can't change the CLAModel behavior without a thorough evaluation of the anomaly detection performance.

nupic/algorithms/anomaly.py
scottpurdy: Yes, it seems like this poses a serialization problem.

vitaly-krugl:
@breznak, I am done with the code review. There is a bunch of feedback. Also, see @scottpurdy's comment about your changes in `Anomaly.compute`:
> @breznak - Since you integrated this class into the CLAModel, the output must still use the same likelihood logic. We can't change the CLAModel behavior without a thorough evaluation of the anomaly detection performance.

nupic/algorithms/anomaly.py
breznak: I agree, decoupled the `mode`(now string only) and callable `customComputeFn`. A disadvantage is that now CLAModel has to be updated to "read" customComputeFn param from OPF description files.

nupic/algorithms/anomaly.py
breznak: I liked it here, as it separates general logic of `Anomaly` and a place where default and custom compute functions can be placed.

nupic/algorithms/anomaly.py
breznak: @vitaly-krugl about anomalyScore being in range of float 0..1: the docstring should still be valid, documentation for custom functions require the score to be such.

@scottpurdy about weighted mode: the change is from `weighted = raw * (1-likelihood)` to `weighted = raw * likelihood` ...so the performance should not change, only the interpretation, should I revert, and open new issue?

breznak:
@vitaly-krugl thank you for review! I've implemented the changes. Serialization test is blocking this and waiting in a separate PR, it seems there is the problem with `callable` serialization, if I recall correctly, that was also reason why I had the `getComputeFn()` called every loop, instead of storing the callable as an internal variable. Thinking about it, is it really too bad to call the if switch everytime?

breznak:
waiting for #2245 first

nupic/algorithms/anomaly.py
david-ragazzi: Not problem to me on keep this separator (if really necessary) but please move it to the midle of the modules not above the next `def`. Something like this:

```
def funcion1():
  ...


##########


def function2():
   ..
```

nupic/algorithms/anomaly.py
```diff
@@ -107,6 +124,27 @@ def __init__(self, slidingWindowSize=None, mode=MODE_PURE, binaryAnomalyThreshol
...
+      self._computeAnomaly =  self._computeRaw
+    elif mode == Anomaly.MODE_LIKELIHOOD:
+      self._likelihood = AnomalyLikelihood() # probabilistic anomaly
+      self._computeAnomaly = self._computeLikelihood
+    elif mode == Anomaly.MODE_WEIGHTED:
+      self._likelihood = AnomalyLikelihood() # probabilistic anomaly
+      self._computeAnomaly = self._computeWeighted
+    else:
+      raise ValueError("Anomaly: mode has to be one of '%s' but is '%r' "
+                         % (Anomaly.supportedModes, mode) )
```
vitaly-krugl: Can access `Anomaly.supportedModes` from here as `self.supportedModes`

nupic/algorithms/anomaly.py
```diff
@@ -107,6 +124,27 @@ def __init__(self, slidingWindowSize=None, mode=MODE_PURE, binaryAnomalyThreshol
...
                        "or None if disabled.")
+    # switch and setup compute() based on mode
+    self._mode = mode
+    self._likelihood = None
+
+    if mode == Anomaly.MODE_CUSTOM:
+      if callable(customComputeFn):
+        self._computeAnomaly = customComputeFn
+      else:
+        raise ValueError("Anomaly: with mode='custom' a callable must be provided as 'customComputeFn' param")
```
vitaly-krugl: To help with debugging, show what was actually passed; e.g., 

```
raise ValueError("Anomaly: with mode='custom' a callable must be provided as 'customComputeFn' param, but got %r" % (customComputeFn,))

Comment (vitaly-krugl) [inline nupic/algorithms/anomaly.py]: Need to run through pylint with Nupic pylint config file, and fix findings. In this case, I suspect that the exception and its text go beyond 80 columns

File: nupic/algorithms/anomaly.py (hunk 26 above)
Comment (vitaly-krugl) [inline nupic/algorithms/anomaly.py]: @scottpurdy, what's your answer for

> so the performance should not change, only the interpretation, should I revert, and open new issue?

File: nupic/algorithms/anomaly.py [hunk 38]
```diff
@@ -107,6 +124,27 @@ def __init__(self, slidingWindowSize=None, mode=MODE_PURE, binaryAnomalyThreshol
           binaryAnomalyThreshold <= 0.0 ):
       raise ValueError("Anomaly: binaryAnomalyThreshold must be from (0,1) "
                        "or None if disabled.")
+    # switch and setup compute() based on mode
+    self._mode = mode
+    self._likelihood = None
+
+    if mode == Anomaly.MODE_CUSTOM:
+      if callable(customComputeFn):
+        self._computeAnomaly = customComputeFn
```
vitaly-krugl: How does `self._computeAnomaly` get restored during deserialization?

vitaly-krugl:
@breznak: I am done with this code review. Please see my feedback. In addition, please resolve the following issue with @scottpurdy:
> about weighted mode: the change is from weighted = raw * (1-likelihood) to weighted = raw * likelihood ...so the performance should not change, only the interpretation, should I revert, and open new issue?

breznak:
@vitaly-krugl implemented your remarks, now things on the list: 
[ ] merge #2246 
[ ] resolve the `weighted mode` issue
[ ] resolve serialization of callables - can CAPNP do it?/"wont implement for CUSTOM"/use the getComputeFn() call in each iteration and avoid storing a callable at all.

nupic/algorithms/anomaly.py
```diff
@@ -62,45 +68,61 @@ class Anomaly(object):
         anomaly scores
     MODE_WEIGHTED - multiplies the likelihood result with the raw anomaly score
         that was used to generate the likelihood
+    MODE_CUSTOM - user defined callable used in compute()
```
vitaly-krugl: nit: "user defined" --> "user-defined"

nupic/algorithms/anomaly.py
```diff
@@ -62,45 +68,61 @@ class Anomaly(object):
...
+                           inputValue: raw input (default=None) 
+                           timestamp: (default=None)
+                         return:
+                           anomaly score: float [0..1]
+                       @see description for compute() method for details
     """
     self._mode = mode
+    self._likelihood = None
+    self._movingAverage = None
+    self._likelihood = None
```
vitaly-krugl: This is redundant: `self._likelihood` was initialized two lines above already.

tests/unit/nupic/algorithms/anomaly_test.py
```diff
@@ -109,11 +110,73 @@ def testAnomalyCumulative(self):
...
+
+    for a in inst:
+      stored = pickle.dumps(a)
+      restored = pickle.loads(stored)
+      #TODO add anomaly _eq_
+      self.assertEqual(a._computeAnomaly, restored._computeAnomaly)
+
+
+
+  def testSerialization(self):
```
vitaly-krugl: `testSerialization`: method with same name already exists above.

vitaly-krugl:
@breznak: Thank you for your work on this. I am done with this round of code review. I found a couple of anomalies in the code. Please see my feedback. In addition, please resolve this issue with @scottpurdy:
> about weighted mode: the change is from weighted = raw * (1-likelihood) to weighted = raw * likelihood ...so the performance should not change, only the interpretation, should I revert, and open new issue?

I am going to be off the grid for the rest of the week.

breznak:
@rhyolight have we witnessed some quantum bit-flip? :game_die: :wink: 
https://travis-ci.org/numenta/nupic/jobs/69118397#L3644
Only 1 compiler failing, on an unrelated SP test.

rhyolight:
WAT? I restarted the job :confused:

nupic/algorithms/anomaly.py
scottpurdy: Yes, revert. That is a pretty invasive change and we would want to coordinate the change. I personally don't see a benefit from it. Plus it should be in a separate PR regardless.

nupic/algorithms/anomaly.py
scottpurdy: remove this line

scottpurdy:
This PR needs some clean up. It seems to be doing three different things: adding the ability to register a custom compute function, refactoring the organization of the currently implemented compute functions, and changing the likelihood compute function to return the inverse of what it previously returned. Each of these should be in a separate PR.

There are also 58 commits so far. The commits are what actually live on in the repository history so they are the official log record. Please squash these down into logical units with clear commit messages.

Finally, please review the [Python style guide](https://github.com/numenta/nupic/wiki/Python-Style-Guide), [PEP 8](https://www.python.org/dev/peps/pep-0008/), and [PEP-257](https://www.python.org/dev/peps/pep-0257/). The latter two have guidelines on indentation and doc strings that you should apply to these changes.

breznak:
I should get back to this..! :)

PR #1948, opened 2015-03-24
Anomaly init factory

### LINKED ISSUE #1870: Separate anomaly types into separate subclasses
The anomaly class groups different types of computations that require different inputs and different internal states. We should separate it into an abstract base class (define common methods, not constructor or internal state) and subclasses that define internal state, constructor, and serialization. We will need a factory function that takes a string name and parameters and returns the appropriate instance.

Later we should also move the anomaly computations into a separate network region but that is outside the scope of this issue.

breznak:
Fixes #1870 

...partly, adds `Anomaly.initFactory()`, I believe Anomaly sub-classes are not needed yet. 
adds Anomaly `==` 
turns runtime checks into asserts

### DIFF
--- a/nupic/algorithms/anomaly.py
+++ b/nupic/algorithms/anomaly.py
@@ -102,15 +102,16 @@ class Anomaly(object):
     if self._mode == Anomaly.MODE_LIKELIHOOD or \
        self._mode == Anomaly.MODE_WEIGHTED:
       self._likelihood = AnomalyLikelihood() # probabilistic anomaly
-    assert self._mode in Anomaly._supportedModes, "Invalid anomaly mode; \
-                       only supported modes are: \n\
-                       Anomaly.MODE_PURE, Anomaly.MODE_LIKELIHOOD, \n\
-                       Anomaly.MODE_WEIGHTED;\n you used: %r" % self._mode
+    if not self._mode in Anomaly._supportedModes:
+      raise ValueError("Invalid anomaly mode; only supported modes are: "
+                       "Anomaly.MODE_PURE, Anomaly.MODE_LIKELIHOOD, "
+                       "Anomaly.MODE_WEIGHTED; you used: %r" % self._mode)
     self._binaryThreshold = binaryAnomalyThreshold
-    assert isinstance(binaryAnomalyThreshold, float) and \
-       binaryAnomalyThreshold < 1.0  and \
-       binaryAnomalyThreshold >= 0.0, "Anomaly: \
-      binaryAnomalyThreshold must be from [0,1); 0 for disabled."
+    if not isinstance(binaryAnomalyThreshold, float) or \
+       binaryAnomalyThreshold >= 1.0  or \
+       binaryAnomalyThreshold < 0.0:
+      raise ValueError("Anomaly: binaryAnomalyThreshold must be from [0,1) "
+                       "0 for disabled.")
 
 
   def compute(self, activeColumns, predictedColumns, 
@@ -134,8 +135,9 @@ class Anomaly(object):
     if self._mode == Anomaly.MODE_PURE:
       score = anomalyScore
     elif self._mode == Anomaly.MODE_LIKELIHOOD:
-      assert inputValue is not None, "Selected anomaly mode \
-        'Anomaly.MODE_LIKELIHOOD' requires 'inputValue' as parameter to compute() method. "
+      if inputValue is None:
+        raise ValueError("Selected anomaly mode 'Anomaly.MODE_LIKELIHOOD' "
+                 "requires 'inputValue' as parameter to compute() method. ")
 
       probability = self._likelihood.anomalyProbability(
           inputValue, anomalyScore, timestamp)
--- a/nupic/algorithms/anomaly.py
+++ b/nupic/algorithms/anomaly.py
@@ -135,15 +135,15 @@ class Anomaly(object):
     if self._mode == Anomaly.MODE_PURE:
       score = anomalyScore
     elif self._mode == Anomaly.MODE_LIKELIHOOD:
-      if inputValue is None:
-        raise ValueError("Selected anomaly mode 'Anomaly.MODE_LIKELIHOOD' "
-                 "requires 'inputValue' as parameter to compute() method. ")
-
+      assert inputValue is not None, ("Selected anomaly mode "
+        "'Anomaly.MODE_LIKELIHOOD' requires 'inputValue' as parameter to compute() method. ")
       probability = self._likelihood.anomalyProbability(
           inputValue, anomalyScore, timestamp)
       # low likelihood -> hi anomaly
       score = 1 - probability
     elif self._mode == Anomaly.MODE_WEIGHTED:
+      assert inputValue is not None, ("Selected anomaly mode "
+        "'Anomaly.MODE_WEIGHTED' requires 'inputValue' as parameter to compute() method. ")
       probability = self._likelihood.anomalyProbability(
           inputValue, anomalyScore, timestamp)
       score = anomalyScore * (1 - probability)
--- a/nupic/algorithms/anomaly.py
+++ b/nupic/algorithms/anomaly.py
@@ -75,10 +75,11 @@ class Anomaly(object):
   _supportedModes = (MODE_PURE, MODE_LIKELIHOOD, MODE_WEIGHTED)
 
 
-  def __init__(self, slidingWindowSize=None, mode=MODE_PURE, binaryAnomalyThreshold=None):
+  def __init__(self, slidingWindowSize=0, mode=MODE_PURE, binaryAnomalyThreshold=0.0):
     """
     @param slidingWindowSize (optional) - how many elements are summed up;
-        enables moving average on final anomaly score; int >= 0
+        enables moving average on final anomaly score; int >= 0;
+        0 means disabled (default)
     @param mode (optional) - (string) how to compute anomaly;
         possible values are:
           - "pure" - the default, how much anomal the value is;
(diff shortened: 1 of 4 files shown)


rhyolight:
@scottpurdy

nupic/algorithms/anomaly.py
```diff
@@ -90,26 +91,26 @@ def __init__(self, slidingWindowSize=None, mode=MODE_PURE, binaryAnomalyThreshol
...
+                       Anomaly.MODE_PURE, Anomaly.MODE_LIKELIHOOD, \n\
+                       Anomaly.MODE_WEIGHTED;\n you used: %r" % self._mode
     self._binaryThreshold = binaryAnomalyThreshold
-    if binaryAnomalyThreshold is not None and ( 
-          not isinstance(binaryAnomalyThreshold, float) or
-          binaryAnomalyThreshold >= 1.0  or 
-          binaryAnomalyThreshold <= 0.0 ):
-      raise ValueError("Anomaly: binaryAnomalyThreshold must be from (0,1) "
-                       "or None if disabled.")
+    assert isinstance(binaryAnomalyThreshold, float) and \
```
rhyolight: `isinstance(binaryAnomalyThreshold, float)` will always be true because you changed the default value to `0.0`.

breznak: not if you pass it "iBreakRules" value :)

nupic/algorithms/anomaly.py
```diff
@@ -90,26 +91,26 @@ def __init__(self, slidingWindowSize=None, mode=MODE_PURE, binaryAnomalyThreshol
...
+       self._mode == Anomaly.MODE_WEIGHTED:
       self._likelihood = AnomalyLikelihood() # probabilistic anomaly
-    if not self._mode in Anomaly._supportedModes:
-      raise ValueError("Invalid anomaly mode; only supported modes are: "
-                       "Anomaly.MODE_PURE, Anomaly.MODE_LIKELIHOOD, "
-                       "Anomaly.MODE_WEIGHTED; you used: %r" % self._mode)
+    assert self._mode in Anomaly._supportedModes, "Invalid anomaly mode; \
+                       only supported modes are: \n\
+                       Anomaly.MODE_PURE, Anomaly.MODE_LIKELIHOOD, \n\
+                       Anomaly.MODE_WEIGHTED;\n you used: %r" % self._mode
```
rhyolight: Put strings in parens instead of using the `\` line continuation.

nupic/algorithms/anomaly.py
```diff
@@ -90,26 +91,26 @@ def __init__(self, slidingWindowSize=None, mode=MODE_PURE, binaryAnomalyThreshol
...
 
-    if self._mode == Anomaly.MODE_LIKELIHOOD or self._mode == Anomaly.MODE_WEIGHTED:
+    if self._mode == Anomaly.MODE_LIKELIHOOD or \
+       self._mode == Anomaly.MODE_WEIGHTED:
       self._likelihood = AnomalyLikelihood() # probabilistic anomaly
-    if not self._mode in Anomaly._supportedModes:
-      raise ValueError("Invalid anomaly mode; only supported modes are: "
-                       "Anomaly.MODE_PURE, Anomaly.MODE_LIKELIHOOD, "
-                       "Anomaly.MODE_WEIGHTED; you used: %r" % self._mode)
+    assert self._mode in Anomaly._supportedModes, "Invalid anomaly mode; \
```
rhyolight: Using `assert` like this isn't a good idea because "The current code generator emits no code for an assert statement when optimization is requested at compile time." See https://docs.python.org/2/reference/simple_stmts.html#the-assert-statement

nupic/algorithms/anomaly.py
rhyolight: Also replace `assert` with explicit condition (see comment above).

nupic/algorithms/anomaly.py
```diff
@@ -133,9 +134,8 @@ def compute(self, activeColumns, predictedColumns,
     if self._mode == Anomaly.MODE_PURE:
       score = anomalyScore
     elif self._mode == Anomaly.MODE_LIKELIHOOD:
-      if inputValue is None:
-        raise ValueError("Selected anomaly mode 'Anomaly.MODE_LIKELIHOOD' "
-                 "requires 'inputValue' as parameter to compute() method. ")
+      assert inputValue is not None, "Selected anomaly mode \
```
rhyolight: Remove `assert`.

nupic/algorithms/anomaly.py
```diff
@@ -90,26 +91,26 @@ def __init__(self, slidingWindowSize=None, mode=MODE_PURE, binaryAnomalyThreshol
...
-    if binaryAnomalyThreshold is not None and ( 
-          not isinstance(binaryAnomalyThreshold, float) or
-          binaryAnomalyThreshold >= 1.0  or 
-          binaryAnomalyThreshold <= 0.0 ):
-      raise ValueError("Anomaly: binaryAnomalyThreshold must be from (0,1) "
-                       "or None if disabled.")
+    assert isinstance(binaryAnomalyThreshold, float) and \
+       binaryAnomalyThreshold < 1.0  and \
+       binaryAnomalyThreshold >= 0.0, "Anomaly: \
+      binaryAnomalyThreshold must be from [0,1); 0 for disabled."
```
rhyolight: Put strings in parens instead of using the `\` line continuation.

nupic/algorithms/anomaly.py
```diff
@@ -160,6 +160,39 @@ def compute(self, activeColumns, predictedColumns,
...
+    Use None to automatically supply default values.
+
+    @param slidingWindowSize (str) length of sliding window
+    @param mode (str) Anomaly mode
+    @param binaryAnomalyThreshold (str) binary discretization threshold
+    @return Anomaly instance, call Anomaly.initFactory(....) 
+    """
+    # defaults
+    if slidingWindowSize is None:
+      slidingWindowSize = "0"
```
rhyolight: Why is this a string instead of a number?

nupic/algorithms/anomaly.py
```diff
@@ -160,6 +160,39 @@ def compute(self, activeColumns, predictedColumns,
...
+    if mode is None:
+      mode = "pure"
+    if binaryAnomalyThreshold is None:
+      binaryAnomalyThreshold = "0.0"
```
rhyolight: Why is this a string instead of a number?

nupic/algorithms/anomaly.py
```diff
@@ -160,6 +160,39 @@ def compute(self, activeColumns, predictedColumns,
...
+
+    try:
+      iWindow = int(slidingWindowSize)
+      fThreshold = float(binaryAnomalyThreshold)
```
rhyolight: Ok, I see what you're doing here, but I still think they should be initialized in the correct type above.

rhyolight:
I added comments mostly about code style. @scottpurdy will have to do the functional review.

nupic/algorithms/anomaly.py
breznak: I knew that and that's what I wanted, on the second thought maybe I'll revert it, as this is run only once, and it's better to get err message that slightly faster code there. Ok, good point :+1:

nupic/algorithms/anomaly.py
breznak: on the contrary to the other cases, I think here assert should stay. compute() is executed each iteration, and the check is "just for extra safety", it's fine to optimize it out in production.

nupic/algorithms/anomaly.py
breznak: correct arguments for "factory" are strings (by the use-case)

breznak:
Thanks @rhyolight, commented on some, will fix elsewhere.

scottpurdy:
Factories are used when you have different classes and you want to instantiate one of them based on user input or something. So the factory uses a mapping from string identifiers to the actual class to instantiate.

But in this case there is only one class. So I'm not sure the purpose of the factory when you can just pass the parameters to the class constructor.

breznak:
I thought it was your idea Scott, the usecase here is initialization from
OPF: we want "typed" init parameters in the default constructor, but this
factory allows convenient initialization from strings in description.py;
the most prominent example can be found in the other PR, which adds a new
param (type function) - to initialize the Anomaly class, one would have to
import anomaly and from there use anomaly.functionXY to pass it as the
param, with factory you'd just use "functionXY"...

scottpurdy:
@breznak - PR link?

And yes I'd like to move to subclassing an Anomaly class which would necessitate a factory but I wouldn't want to pass different types around as strings.

breznak:
@scottpurdy check PR descr please, "partly fixes https://github.com/numenta/nupic/issues/1870", under that Issue I give my opinion why I think anomaly subclasses are not needed yet. 
See https://github.com/numenta/nupic/pull/1934 for obvious motivation for this PR, but I agree I can change the cast of numeric types from strings back to numeric.

breznak:
What can we do to proceed on this 1?

rhyolight:
@scottpurdy ping

scottpurdy:
I don't think we should do this. It isn't necessary to have a factory in this case and it encourages "bad behavior" by allowing users to instantiate the class via string rather than directly.

Not only this, but the issue that this claims to fix doesn't actually get addressed - all of the anomaly implementations are part of the same Anomaly class still.

breznak:
@scottpurdy 
>  It isn't necessary to have a factory in this case and it encourages "bad behavior" by allowing users to instantiate the class via string rather than directly. 

I'm ok to remove the "string" parameters for the integer etc arguments (so it's type-correct), please see the PR for "alternative anomaly score implementations" to understand motivation for this "factory". I pass a function as an argument there, therefore this PR helps (eg in OPF model, where one does not have to import all the functions and can define them by "string" only.) Maybe those 2 PRs can be merged to make more sense?

Apart from that, there are some commits that should not be lost if we closed this PR (__cmp__, avoided None for some params in anomaly)

breznak:
closing, replacing by light version in #2146

Issue #1959, opened 2015-03-27
Add deserialization constructor for SP


scottpurdy:
The new capnp deserialization for the SP requires that you first construct an instance which is wasteful. We should use a `__new__` class method that takes a proto instance to bypass the unnecessary initialization.

FYI @oxtopus

Issue #1715, opened 2015-01-05
Integrate Cap'n Proto serialization with encoders


scottpurdy:
The encoders should have `read` and `write` methods that take proto references and serialize or deserialize.

* * *

- [x] ScalarEncoder - #1961
- [x] RandomDistributedScalarEncoder - #1962
- [x] DateTimeEncoder - #1963
- [x] CategoryEncoder - #1964
- [x] SDRCategoryEncoder - #1965
- [x] AdaptiveScalarEncoder - #1986
- [x] CoordinateEncoder - #1987
- [x] DeltaEncoder - #1988
  - Merged in #2007
- [x] GeospatialCoordinateEncoder - #1989
  - Merged in #2007
- [x] LogEncoder - #1990
  - Merged in #2007
- [x] MultiEncoder - #1991
  - In review at #2007
- [x] PassThroughEncoder - #1992
  - Merged in #2007
- [x] ScalarSpaceEncoder - #1993
  - Deprecated?
- [x] SparsePassThroughEncoder - #1994 
  - Merged in #2007

rhyolight:
@scottpurdy @oxtopus Good job filling this ticket out with subtasks. I appreciate it greatly. :+1:

oxtopus:
@scottpurdy All tasks are either completed or pending review.  Do you mind giving them a look?

rhyolight:
:boom: Nice job Austin.

PR #1960, opened 2015-03-27
Integrate capnproto serialization into ScalarEncoder re: #1715

### LINKED ISSUE #1961: Add capnp serialization to ScalarEncoder
See parent #1715

oxtopus:
re: #1715 and #1959 this adds the requisite schema, `read()`, and `write()` methods for capnproto serialization and deserialization with basic test.

fixes #1961

@scottpurdy


nupic/encoders/scalar.py
```diff
@@ -714,3 +716,49 @@ def dump(self):
     print "  nInternal: %d" % self.nInternal
     print "  rangeInternal: %f" % self.rangeInternal
     print "  padding: %d" % self.padding
+
+
+  def __new__(cls, *args, **kwargs):
```
scottpurdy: Could we instead change `read` to be a classmethod that does this? Or rename this function to something other than `__new__`? It seems odd to pass a capnp proto as the `cls` argument of `__new__`.

nupic/encoders/scalar.capnp
```diff
@@ -0,0 +1,20 @@
+@0xfa7d16f86048a6e4;
+
+struct ScalarEncoderProto {
+  w @0 :Int32;
```
scottpurdy: `UInt32`

nupic/encoders/scalar.capnp
```diff
@@ -0,0 +1,20 @@
...
+  minval @1 :Float32;
+  maxval @2 :Float32;
+  periodic @3 :Bool;
+  n @4 :Int32;
```
scottpurdy: `UInt32`

nupic/encoders/scalar.capnp
```diff
@@ -0,0 +1,20 @@
...
+  radius @5 :Float32;
+  resolution @6 :Float32;
```
scottpurdy: Drop `resolution` since it is redundant when you have `minval` and `maxval`

nupic/encoders/scalar.capnp
```diff
@@ -0,0 +1,20 @@
...
+  name @7 :Text;
+  verbosity @8 :Int8;
```
scottpurdy: `UInt8`?

tests/unit/nupic/encoders/scalar_test.py
```diff
@@ -410,6 +413,36 @@ def testGetBucketInfoIntResolution(self):
...
+    # Write the proto to a temp file and read it back into a new proto
+    with tempfile.TemporaryFile() as f:
+      proto1.write(f)
+      f.seek(0)
+      proto2 = ScalarEncoderProto.read(f)
+
+    encoder = ScalarEncoder.__new__(proto2)
+
+    self.assertIsInstance(encoder, ScalarEncoder)
+    self.assertEqual(encoder.w , self._l.w)
```
scottpurdy: Remove space before comma here and in following lines

nupic/encoders/scalar.capnp
```diff
@@ -0,0 +1,20 @@
...
+  clipInput @9 :Bool;
+
+  halfwidth @10 :Int32;
```
scottpurdy: It would be nice to have descriptions of these attributes. I don't know why `halfwidth` is needed for example. If you don't know why some of these exist then we should figure it out.

tests/unit/nupic/encoders/scalar_test.py
```diff
@@ -410,6 +413,36 @@ def testGetBucketInfoIntResolution(self):
     self.assertEqual(4.5, encoder.topDownCompute(encoder.encode(4.5))[0].scalar)
 
 
+  def testCapNProtoSerialization(self):
+    """Test ScalarEncoder CapNProto serialization implementation"""
+    proto1 = ScalarEncoderProto.new_message()
+    self._l.write(proto1)
```
scottpurdy: While this would be a useful additional test, we should have one where you encode something before serialization and then again at the end to make sure the deserialized and original get same encodings

scottpurdy:
Done with initial review

nupic/encoders/scalar.capnp
oxtopus: You may not have `minval` and `maxval`.

nupic/encoders/scalar.capnp
oxtopus: Ultimately, I think that `halfwidth` and the others that are normally derived in `__init__()` and `_initEncoder()` should be deprecated in the schema, but in this implementation, we're bypassing the `__init__()` method, so I opted to do a straight mapping of values for the full state of the object.  The current `__init__()` and `_initEncoder()` methods need to know the original arguments to the constructor, which aren't preserved in the serialization, and fixing that up would expand the scope of this task.

I'd rather do as you suggest and remove `halfwdith`, etc., but only after `__init__()` and `_initEncoder()` are fixed as I describe.

scottpurdy:
Ok it actually looks like it is n, radius, and resolution that are redundant. In `initEncoder` it checks which is specified and sets the others based on that. So just pick one of them (I suggest `n`) to store and then when you load you can call `initEncoder` to calculate the other two

nupic/encoders/scalar.capnp
scottpurdy: Ok it actually looks like it is n, radius, and resolution that are redundant. In `initEncoder` it checks which is specified and sets the others based on that. So just pick one of them (I suggest `n`) to store and then when you load you can call `initEncoder` to calculate the other two

david-ragazzi:
Just a question: why serialize an encoder?

scottpurdy:
@david-ragazzi - you need an encoder to use a model. Even if the encoder has no learning (which is not the case with the ScalarEncoder) you need to at least store the parameters for the encoder.

oxtopus:
@scottpurdy I believe your comments have been addressed.  Mind giving it one more review?

nupic/encoders/scalar.py
```diff
@@ -30,6 +32,11 @@
 
 
 
+DEFAULT_RADIUS = 0
```
scottpurdy: Really odd that these are `0` instead of `None` but outside the scope of this change.

scottpurdy:
:+1:

nupic/encoders/scalar.capnp
```diff
@@ -0,0 +1,14 @@
+@0xfa7d16f86048a6e4;
+
+struct ScalarEncoderProto {
+  # ScalarEncoder() constructor signature
```
scottpurdy: I'm not sure that this comment is necessary. I'd rather have the schema explain the parameters rather than rely on the code implementations.

nupic/encoders/scalar.capnp
```diff
@@ -0,0 +1,13 @@
+@0xfa7d16f86048a6e4;
+
+# Next ID: 10
```
scottpurdy: Oops my change to your PR didn't update this!

oxtopus: I'll take care of it in a future pr.

PR #1971, opened 2015-03-30
Remove explicit int casts and update tests to allow ints or longs.

scottpurdy:


scottpurdy:
@oxtopus please review

oxtopus:
:+1:

PR #2028, opened 2015-04-09
Cleanup date encoder unit tests

### LINKED ISSUE #2027: Cleanup Date encoder unit tests
Part of #2000

oxtopus:
Fixes #2027

### DIFF
--- a/tests/unit/nupic/encoders/date_test.py
+++ b/tests/unit/nupic/encoders/date_test.py
@@ -34,17 +34,17 @@ from nupic.encoders.date_capnp import DateEncoderProto
 
 
 
-#########################################################################
 class DateEncoderTest(unittest.TestCase):
-  '''Unit tests for DateEncoder class'''
+  """Unit tests for DateEncoder class"""
 
 
   def setUp(self):
-    ##TODO: comment and code don't match - weekend?!!
-    # 3 bits for season, 1 bit for day of week, 2 for weekend, 5 for time of day
-     # use of forced is not recommended, used here for readibility, see scalar.py
-    self._e = DateEncoder(season=3, dayOfWeek=1, weekend=3, timeOfDay=5)
-    # in the middle of fall, Thursday, not a weekend, afternoon - 4th Nov, 2010, 14:55
+    # 3 bits for season, 1 bit for day of week, 1 for weekend, 5 for time of
+    # day
+    # use of forced is not recommended, used here for readibility, see scalar.py
+    self._e = DateEncoder(season=3, dayOfWeek=1, weekend=1, timeOfDay=5)
+    # in the middle of fall, Thursday, not a weekend, afternoon - 4th Nov,
+    # 2010, 14:55
     self._d = datetime.datetime(2010, 11, 4, 14, 55)
     self._bits = self._e.encode(self._d)
     # season is aaabbbcccddd (1 bit/month) # TODO should be <<3?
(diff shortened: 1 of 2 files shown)


tests/unit/nupic/encoders/date_test.py
```diff
@@ -34,17 +34,17 @@
...
 
-#########################################################################
 class DateEncoderTest(unittest.TestCase):
-  '''Unit tests for DateEncoder class'''
+  """Unit tests for DateEncoder class"""
 
 
   def setUp(self):
-    ##TODO: comment and code don't match - weekend?!!
-    # 3 bits for season, 1 bit for day of week, 2 for weekend, 5 for time of day
```
oxtopus: @scottpurdy Any idea the history of this comment?  IT seems neither the comment, nor the call to constructor are correct.  According to docstring, it should only be 0 or 1, although technically it could be any odd number since it becomes `w` for the `ScalarEncoder`.

scottpurdy: No I don't know the history. I'd guess they wrote the doc string, found out it had to be odd so changed to 3 instead of 2. Sounds like you are saying the encoder doc string description of weekend is wrong as well?

oxtopus: Yep.  Weekend need only be 1, which differs from docstring.

tests/unit/nupic/encoders/date_test.py
```diff
@@ -57,69 +57,69 @@ def setUp(self):
...
-    # time of day has radius of 4 hours and w of 5 so each bit = 240/5 min = 48min
-    # 14:55 is minute 14*60 + 55 = 895; 895/48 = bit 18.6
+    # time of day has radius of 4 hours and w of 5 so each bit = 240/5
+    # min = 48min 14:55 is minute 14*60 + 55 = 895; 895/48 = bit 18.6
     # should be 30 bits total (30 * 48 minutes = 24 hours)
-    timeOfDayExpected = [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,1,1,1,1,1,0,0,0,0,0,0,0,0,0]
-    self._expected = numpy.array(seasonExpected + dayOfWeekExpected + weekendExpected \
-                          + timeOfDayExpected, dtype=defaultDtype)
+    timeOfDayExpected = (
+      [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,1,1,1,1,1,0,0,0,0,0,0,0,0,0])
```
david-ragazzi: @oxtopus Why this and other lists in this file doesn't have space after collon while others have?

david-ragazzi: Never mind. I'm confunsing lists with parameters to a function.

david-ragazzi:
:+1:

tests/unit/nupic/encoders/date_test.py
```diff
@@ -57,69 +57,69 @@ def setUp(self):
...
     self.assertEqual(sum(mvOutput), 0)
 
+
   def testDecoding(self):
-    '''decoding date'''
+    """decoding date"""
     decoded = self._e.decode(self._bits)
 
-    (fieldsDict, fieldNames) = decoded
+    (fieldsDict, _) = decoded
```
scottpurdy: sometimes it is useful to keep a descriptive name like `_fieldNames`. May not be important here but just to keep in mind that lint will understand it as unused still

tests/unit/nupic/encoders/date_test.py
scottpurdy: You're right that there should be spaces after commas (everywhere AFAIK) but in this case it might be prohibitively verbose. A possible alternative would be something like this:

```
timeOfDayExpected = "00000000000000000000111110000000".split("")
```

But that may incur some computation cost (which is probably fine for a test)

PR #2059, opened 2015-04-17
Adds AnomalyRegion implementation

### LINKED ISSUE #2058: Create an AnomalyRegion for computing the raw anomaly score
Currently, you have to extract the right info from the network and calculate the raw anomaly score manually. There should be a region that performs this computation.

scottpurdy:
- Adds AnomalyRegion for computing the raw anomaly score.
- Updates the network api example to use the new anomaly region
- Updates PyRegion to have better error messages.

Note - the lookup logic for Python regions requires the same module name as the class name, hence the camelcase file name.

fixes #2058

@chetan51 - please review


nupic/regions/AnomalyRegion.py
```diff
@@ -0,0 +1,94 @@
...
+  def __init__(self, *args, **kwargs):
+    self.prevPredictedColumns = numpy.zeros([], dtype="float32")
+
+
+  def initialize(self, inputs, outputs):
+    pass
+
+
+  def compute(self, inputs, outputs):
+    activeColumns = inputs["activeColumns"].nonzero()[0]
```
breznak: :+1:  I like this, as it potentially allows new anomaly computations that access and use different "data structures" from Sp/TP,..

nupic/regions/AnomalyRegion.py
```diff
@@ -0,0 +1,94 @@
...
+
+    rawAnomalyScore = anomaly.computeRawAnomalyScore(
+        activeColumns, self.prevPredictedColumns)
```
breznak: TODO: (CLA)Model must be updated to use this `getAnomalyRegion()` instead of current `anomalyInstance`; 
also, then the ugly code in CLAModel to get `activecolumns and prevPredicted columns` can be moved to the anomaly region.

scottpurdy: That is fine with me but outside the scope of this PR. I'm going to be trying to make the network api as simple and clean to use as possible so things like the CLAModel that encapsulate certain network configurations will become less needed.

nupic/regions/AnomalyRegion.py
```diff
@@ -0,0 +1,93 @@
...
+        "commands": {
+        },
+    }
+
+
+  def __init__(self, *args, **kwargs):
+    self.prevPredictedColumns = numpy.zeros([], dtype="float32")
+
+
+  def initialize(self, inputs, outputs):
```
chetan51: Is this necessary?

chetan51:
:+1:

PR #2064, opened 2015-04-20
Topology for PassThrough encoders

### LINKED ISSUE #2062: PassThrough Encoders should support topology
as `spatial pooler` now has support for topology (in inhibition), so should be the encoders, atleast some of them. 
It definitely makes sense for (sparse) `pass-through` encoder;
Use-case is for example in vision

breznak:
Encoders return output in same dimensions as the input; 
now used for PassThrough encoder 
Fixes #2062

### DIFF
--- a/nupic/encoders/base.py
+++ b/nupic/encoders/base.py
@@ -127,9 +127,10 @@ class Encoder(object):
     call.
 
     @param inputData TODO: document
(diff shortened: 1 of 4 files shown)


breznak:
Closing this PR, reasons explained in the issue. If someone bothers to cherry-pick the comment fixes, they are welcome!

Issue #1449, opened 2014-10-16
Platform-Independent Model Serialization - Port existing code


rhyolight:
After the new serialization technique has been implemented, the old technique must be phased out. Both methods should work for a period of time to give client code a chance to switch over. Users need enough time to use a conversion method to re-serialize their models using the new protocol.

Details and sample PRs here:
https://github.com/numenta/nupic/wiki/Serialization#new-format

---
- [x] [Use newer pycapnp dependency and clean up README](https://github.com/numenta/nupic/issues/1686)
- [x] [Enable passing pycapnp builders/readers to C++](https://github.com/numenta/nupic/issues/1794)
- [x] Port misc. components (sparse matrix, RNG, etc)
  - Some of these are done, like sparse matrix implementations and RNG but there are likely others. I think it makes most sense to do these when we need them for one of the algorithms.
- [x] [Integrate Cap'n Proto serialization with encoders](https://github.com/numenta/nupic/issues/1715)
- [x] [Port spatial pooling](https://github.com/numenta/nupic/issues/1836)
- [x] [Port Python temporal memory](https://github.com/numenta/nupic/issues/2089)
- [x] [Port C++ temporal memory](https://github.com/numenta/nupic.core/issues/408)
- [x] [Port TP.py](https://github.com/numenta/nupic.core/issues/425)
- [ ] Port C++ Cells4 and Python TP10X2.py
- [x] [Port C++ CLA classifier](https://github.com/numenta/nupic.core/issues/402)
- [x] [Port Python CLA classifier](https://github.com/numenta/nupic/issues/2090)
- [x] Port C++ Network API (core components, not region implementations) - numenta/nupic.core#441
- [x] Port Python Network API (core components, not region implementations) - #2241
- [x] Port C++ Region implementations (separate for each region) - numenta/nupic.core#458
- [ ] [Port Python Region implementations (separate for each region)](https://github.com/numenta/nupic/issues/2158)
- [ ] Port CLAModel
- [ ] [Port KNN classifier](https://github.com/numenta/nupic/issues/2092)
- [ ] [Create conversion utility to switch from old model serialization format to new one](https://github.com/numenta/nupic/issues/1418)
- [ ] [Make old serialization method obviously deprecated](https://github.com/numenta/nupic/issues/1419)
- [ ] [Remove old serialization function](https://github.com/numenta/nupic/issues/1420)
[...]

PR #1928, opened 2015-03-16
Bench tools

breznak:
simple encoders/SP/TP benchmarks used for `cProfile` module in Python.

Fixes #1984

Issue #1418, opened 2014-10-07
Create conversion utility to switch from old model serialization format to new one


rhyolight:
Depends on #1336.

Issue #1385, opened 2014-09-30
The _mapPotential method and the SparseBinaryArray, return a dense array


cogmission:
_mapPotential is called from the bottom of the  __init__ method where there is a loop which iterates over columns and initializes the assignment of the potentialPool (collection of input bits each column is associated with). The first line of the loop calls _mapPotential which returns a binary array of ones and zeroes (ones in the places indicating the index of the input vector's bits a given column has in its "pool", zeros in the places indicating the indexes of bits unconnected).

This return value should simply be an array of the actual indexes, resulting in a much shorter array.
(for example:  [1,2,4] instead of [0,1,1,0,1])

Loop over 2048 columns--
**Iteration 1**
The reason for this is that this array is then passed to _potentialPools.replaceSparseRow() calling             [].nonzero()[0]. This internally iterates over the indexes in the array and returns [1,2,4], sticking with our example. In the normal case, this is an array of 2048 items, keep in mind. 

**Iteration 2**
The updatePermanencesForColumn() method is then called which in the first line calls:
maskPotential = numpy.where(self._potentialPools.getRow(index) > 0)[0] in order to obtain the array [1,2,4].
self._potentialPools.getRow(index) returns a "dense" array ([0,1,1,0,1]) which numpy must iterate over to extract the indexes of nonzero items seen above. This amounts to 2048 \* 2048.

All of the above iteration would not be needed if _mapPotential returned a "sparse" array and self._potentialPools returned the sparse array ([1,2,4]) instead of the dense array (0,1,1,0,1]).

Again, there would be absolutely no iteration if the above were fixed.

rhyolight:
### Please review this issue

This issue needs to be reviewed by the original author or another contributor for applicability to the current codebase. The issue might be obsolete or need updating to match current standards and practices. If the issue is out of date, please close. Otherwise please leave a comment to justify its continuing existence. It may be closed in the future if no further activity is noted.

breznak:
Still valid, not so performance critical, as this is `init` code only (on the other hand, pySP init takes ages!); 

complexity O(n^2) with possible reduce from dense (2048x2048) to sparse (2%x2%) which means (42x42)!

rhyolight:
I would like to see more discussion of this issue by @numenta/nupic-committers.

PR #2066, opened 2015-04-20
Fixes #1903 - Re-use Connections class in SP

### LINKED ISSUE #1903: Re-use Connections class in SpatialPooler class
In order to simplify/uniformizy current spatial and temporal implementations and ease future changes, I think it's very insteresting we implement Connections class in SpatialPooler class. This already was done nicely in htm.java by @cogmission and it seems that works fine. We would move such class into a `connections.py` file and make TemporalMemory and SpatialPooler import it.

Connections class in Java:
https://github.com/numenta/htm.java/blob/master/src/main/java/org/numenta/nupic/Connections.java

SpatialPooler class in Java re-using connections:
https://github.com/numenta/htm.java/blob/master/src/main/java/org/numenta/nupic/research/SpatialPooler.java#L73

TemporalMemory class in Java re-using connections:
https://github.com/numenta/htm.java/blob/master/src/main/java/org/numenta/nupic/research/TemporalMemory.java#L67

@chetan51 @cogmission

david-ragazzi:
Fixes: https://github.com/numenta/nupic/issues/1903
Depends: https://github.com/numenta/nupic/issues/1449

Ladies and gentlemen, now SP also has connections! Connections class is a prime by @chetan51 which allow us to separate net specification from net state. This seems simple but has a lot of implications to make nupic easier to understanding and mantaining. This class is so abstract when comes to segments and synapses that I managed adapt it to SP without need create any additional method but only re-using the existing ones.

Some important changes:

- `_potentialPools` was removed: Now SP create synapses to all input bits that are in potential area of a column. So when methods like `_adaptSynapses`, `_bumpUpWeakColumns`, `_updatePermanencesForColumn`, need update synapses's permanencies, they don't need handle a `mapPotential` but simply handle directly such synapses of the column or get the indices of the input bits which they are connected.

- `_permanences` was removed: Now SP uses a dict to get permanencies from an individual column and put them into a temporary numpy array to perform changes (`perm`). After permanencies operations is done, SP updates the synapses of a column with the new permanencies values (using _updatePermanencesForColumn method).

I compared the performance without or with these changes and I feel litle difference. But it is desirable we have more eyes on this.

This PR still is not ready because I still have to implement serialization for Connections and more tests are need to we find areas for improvement. 

TODO:
- [X] Move Connections class from TemporalMemory into a separated file.
- [X] Adapt SP to use Connections.
- [ ] Create serialization for Connections.
- [ ] Compare performance.

### DIFF
--- a/nupic/research/spatial_pooler.py
+++ b/nupic/research/spatial_pooler.py
@@ -1253,7 +1253,7 @@ class SpatialPooler(object):
 
     # As each column has only one segment, we can reference it with own column's
     # index
-    if index in self.connections._segmentsForElement:
+    if index in self.connections._segmentsForCell:
       synapses = set(self.connections.synapsesForSegment(index))
       for synapse in synapses:
         self.connections.destroySynapse(synapse)
--- a/nupic/research/spatial_pooler.py
+++ b/nupic/research/spatial_pooler.py
@@ -1359,7 +1359,8 @@ class SpatialPooler(object):
     """
     overlaps = numpy.zeros(self._numColumns).astype(realDType)
     for i in xrange(self._numColumns):
-      inputIndices = self._getInputIndicesFromSynapsesForColumn(i)
+      inputIndices = self._getInputIndicesFromSynapsesForColumn(i,
+                                                                onlyConnected=True)
       overlaps[i] = inputVector[inputIndices].sum()
     overlaps[overlaps < self._stimulusThreshold] = 0
     return overlaps
--- a/nupic/research/spatial_pooler.py
+++ b/nupic/research/spatial_pooler.py
@@ -22,10 +22,9 @@
 import itertools
 
 import numpy
-from nupic.bindings.math import (SM32 as SparseMatrix,
-                                 SM_01_32_32 as SparseBinaryMatrix,
-                                 GetNTAReal,
+from nupic.bindings.math import (GetNTAReal,
                                  Random as NupicRandom)
+from nupic.research.connections import Connections
 
 
 realDType = GetNTAReal()
@@ -233,31 +232,8 @@ class SpatialPooler(object):
     # initialize the random number generators
     self._seed(seed)
 
-    # Store the set of all inputs that are within each column's potential pool.
-    # 'potentialPools' is a matrix, whose rows represent cortical columns, and
-    # whose columns represent the input bits. if potentialPools[i][j] == 1,
-    # then input bit 'j' is in column 'i's potential pool. A column can only be
-    # connected to inputs in its potential pool. The indices refer to a
-    # flattened version of both the inputs and columns. Namely, irrespective
-    # of the topology of the inputs and columns, they are treated as being a
-    # one dimensional array. Since a column is typically connected to only a
-    # subset of the inputs, many of the entries in the matrix are 0. Therefore
-    # the potentialPool matrix is stored using the SparseBinaryMatrix
-    # class, to reduce memory footprint and computation time of algorithms that
-    # require iterating over the data structure.
-    self._potentialPools = SparseBinaryMatrix(numInputs)
-    self._potentialPools.resize(numColumns, numInputs)
-
-    # Initialize the permanences for each column. Similar to the
-    # 'self._potentialPools', the permanences are stored in a matrix whose rows
-    # represent the cortical columns, and whose columns represent the input
-    # bits. If self._permanences[i][j] = 0.2, then the synapse connecting
-    # cortical column 'i' to input bit 'j'  has a permanence of 0.2. Here we
-    # also use the SparseMatrix class to reduce the memory footprint and
-    # computation time of algorithms that require iterating over the data
-    # structure. This permanence matrix is only allowed to have non-zero
-    # elements where the potential pool is non-zero.
-    self._permanences = SparseMatrix(numColumns, numInputs)
+    # Initialize member variables
+    self.connections = Connections(self._numColumns)
 
     # Initialize a tiny random tie breaker. This is used to determine winning
     # columns where the overlaps are identical.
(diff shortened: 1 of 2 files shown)


breznak:
wow :cool: A really exciting change! I think I'll need a brain for the review, so I'll leave that for tmr, but it looks really :+1: :+1:

david-ragazzi:
@breznak haha.. Really this won't an easy job, I even had took advantage that I was with fresh brain and without distractions to finally work on this.. If this improvement has similar or better performance (who knows), then we should say goodbye to the sparse matrix.. and so finally SP will be more intuitive to understand.. :-D

Talking about performance, in the current state of this PR, SP has to transverse synapses to get the input indices which they are connected at every time that such indices are need. I'm thinking of use dicts for perform this operation once and so gain performance.. I'm open to suggestions..

breznak:
I think this would also fix this old pal, https://github.com/numenta/nupic/issues/1385 
because the methods are not used at all after this PR

breznak:
I've just cleaned up the python benchmark PR, so you can use https://github.com/numenta/nupic/pull/1928 to measure the new SP performance.

nupic/research/spatial_pooler.py
```diff
@@ -965,9 +964,10 @@ def _adaptSynapses(self, inputVector, activeColumns):
     permChanges.fill(-1 * self._synPermInactiveDec)
     permChanges[inputIndices] = self._synPermActiveInc
     for i in activeColumns:
-      perm = self._permanences.getRow(i)
-      maskPotential = numpy.where(self._potentialPools.getRow(i) > 0)[0]
-      perm[maskPotential] += permChanges[maskPotential]
+      synapses = self.connections.synapsesForSegment(i)
+      inputIndices = self._getInputIndexesFromSynapses(synapses)
+      perm = self._getPermanencesFromColumn(i)
+      perm += permChanges[inputIndices]
```
breznak: some trouble in adaptSynapses() and the failing test https://travis-ci.org/numenta/nupic/jobs/59624260#L3493

nupic/research/spatial_pooler.py
```diff
@@ -1091,19 +1108,16 @@ def _initPermNonConnected(self):
...
+    array the size of the input, where each entry in the array represents the
+    initial permanence value between the input bit at the particular index in
+    the array, and the column represented by the 'index' parameter.
 
     Parameters:
     ----------------------------
-    @param potential: A numpy array specifying the potential pool of the column.
-                    Permanence values will only be generated for input bits
-                    corresponding to indices for which the mask value is 1.
+    @param index:        The index identifying a column.
```
breznak: Some more details how is the index computed? eg for a 2D SP?

nupic/research/spatial_pooler.py
```diff
@@ -1162,16 +1173,37 @@ def _mapColumn(self, index):
...
+  def _createSynapsesForColumn(self, index, wrapAround=False):
+    """
+    Maps a column to its input bits, create synapses from the column to these,
+    and initializes the permanences of these synapses. This method encapsulates
+    the topology of the region. It takes the index of the column as an argument
+    and determines what are the indices of the input vector that are located
+    within the column's potential pool. The return value is a list containing
+    the indices of the input bits. The current implementation of the base class
+    only supports a 1 dimensional topology of columns with a 1 dimensional
+    topology of inputs. To extend this class to support 2-D topology you will
```
breznak: does it mean `Connections` does not support 2D SP columns, while current implementation does?

nupic/research/connections.py
```diff
@@ -0,0 +1,283 @@
...
+    self._synapsesForSegment = dict()
+    self._synapsesForPresynapticCell = defaultdict(dict)
+
+    # Index of the next segment to be created
+    self._nextSegmentIdx = 0
+    # Index of the next synapse to be created
+    self._nextSynapseIdx = 0
+
+
+  def elementForSegment(self, segment):
```
breznak: nit: maybe call this `getElement()`?

nupic/research/spatial_pooler.py
david-ragazzi: > some trouble in adaptSynapses() and the failing test https://travis-ci.org/numenta/nupic/jobs/59624260#L3493

Yep, I did fix it.

nupic/research/spatial_pooler.py
david-ragazzi: > Some more details how is the index computed? eg for a 2D SP?

Could be. I simply get existing info and add more some details.

nupic/research/spatial_pooler.py
david-ragazzi: > does it mean Connections does not support 2D SP columns, while current implementation does?

Actually this description has text that was "stolen" from `_mapPotential` which its desciption is focused in 2D.

nupic/research/connections.py
david-ragazzi: > nit: maybe call this getElement()?

Could be. `elementForSegment` is original call form used by @chetan51 .

david-ragazzi:
This PR is blocked because the "pickes": it doesn't allow Connections class be serialized. Let's wait CapnProto be the official serialiation method to move forward.

In the meantime, I'll create a PR for move Connections class from TemporalMemory and so decrease the files to review.

rhyolight:
Depends on https://github.com/numenta/nupic/issues/1449

chetan51:
@david-ragazzi You can now consolidate this PR with https://github.com/numenta/nupic/pull/2080.

david-ragazzi:
> @david-ragazzi You can now consolidate this PR with #2080.

Yep, when I go home I'll rebase this PR and revert the related commit (https://github.com/david-ragazzi/nupic/commit/b3af7302f2e3c68d533e4136f5b19803167001af).

david-ragazzi:
> @david-ragazzi You can now consolidate this PR with #2080.

Done.

cogmission:
@david-ragazzi  @rhyolight  Was this merged along with #2080 ?

david-ragazzi:
@david-ragazzi @rhyolight Was this merged along with #2080 ?

Nope. #2080 was only to move Connections from TemporalMemory...

nupic/research/spatial_pooler.py
```diff
@@ -1308,7 +1358,9 @@ def _calculateOverlap(self, inputVector):
                     the spatial pooler.
     """
     overlaps = numpy.zeros(self._numColumns).astype(realDType)
-    self._connectedSynapses.rightVecSumAtNZ_fast(inputVector, overlaps)
+    for i in xrange(self._numColumns):
+      inputIndices = self._getInputIndicesFromSynapsesForColumn(i)
```
david-ragazzi: TODO: I need use `onlyConnected=True` here.

cogmission:
Does anyone know the status of this?

breznak:
afaik, blocked by "serialization stuff"

cogmission:
Ah thanks!

chetan51:
Serialization for Connections is completed: https://github.com/numenta/nupic/pull/2100

david-ragazzi:
> Serialization for Connections is completed: #2100

I think this is not enough because models saved with pickle need be re-serialized to capnproto using the script addressed by https://github.com/numenta/nupic/issues/1418.

PR #2077, opened 2015-04-25
Reflect using pkg_resources to locate resource files as implemented in #1902.

### LINKED ISSUE #2078: NuPIC Walkthrough notebook is still using datasethelpers which was removed in #1902

chrox:
Fixes #2078.

### DIFF
--- a/examples/NuPIC Walkthrough.ipynb
+++ b/examples/NuPIC Walkthrough.ipynb
@@ -881,28 +881,6 @@
      ]
     },
     {
-     "cell_type": "code",
-     "collapsed": false,
-     "input": [
-      "model = ModelFactory.create(MODEL_PARAMS)"
-     ],
-     "language": "python",
-     "metadata": {},
-     "outputs": [
-      {
-       "ename": "NameError",
-       "evalue": "name 'ModelFactory' is not defined",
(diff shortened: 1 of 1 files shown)


rhyolight:
Thanks for the fix. I'll take a look Monday.

examples/NuPIC Walkthrough.ipynb
```diff
@@ -881,28 +881,6 @@
      ]
     },
     {
-     "cell_type": "code",
```
david-ragazzi: @chrox Why this block is being removed?

rhyolight:
Sorry, I totally forgot about this PR. I have the same question as @david-ragazzi.

examples/NuPIC Walkthrough.ipynb
```diff
@@ -881,28 +881,6 @@
...
-     "collapsed": false,
```
chrox: This cell won't run because neither `ModelFactory` nor `MODEL_PARAMS` is defined so it will raise NameError and stop the "run all cells" command. And this line is already in the "Loading Data" section thus unnecessary here except that this cell is teaching how to fail the ipython notebook which I think is highly unlikely.

Issue #2079, opened 2015-04-26
Move Connections class from TemporalMemory to a separate module


david-ragazzi:
As SpatialPooler and other classes can re-use Connections class, it's interesting that this class be called from a specific file than `temporal_memory.py`, i.e. `connections.py`.

rhyolight:
The `status:ready` label is usually used for PRs that are ready for review.

breznak:
:+1:

david-ragazzi:
> The status:ready label is usually used for PRs that are ready for review. 

Thanks..Talking about... What kind of labels an issue should have, or it is not necessary?

rhyolight:
At least one subject and one type, probably.

PR #2080, opened 2015-04-26
Move Connections class from TemporalMemory to a separate module

david-ragazzi:
Fixes: https://github.com/numenta/nupic/issues/2079


david-ragazzi:
@chetan51 समीक्षा करें। धन्यवाद। :smile_cat:

breznak:
good and simple move :+1:

david-ragazzi:
> good and simple move :+1: 

@breznak What do you think about methods renaming? Do you think is necessary, or keep `cellForSegment` term even to get a segment `column` is not confusing...

breznak:
I think it's fine to keep for now, maybe just put a comment to SP and
Connections, that for SP: cell = column, because each column uses just 1
cell, in TP the meaning is separated as 1 col consists of more cells.

david-ragazzi:
> I think it's fine to keep for now, maybe just put a comment to SP and
Connections, that for SP: cell = column, because each column uses just 1
cell, in TP the meaning is separated as 1 col consists of more cells.

Yeah.. I think you're right.. I update the PR, now it only move Connections from `temporal_memory.py` to a file called `connections.py`

david-ragazzi:
@rhyolight But this PR is ready for review.. I asked to @chetan51 in hindi.. :smile:

rhyolight:
Ok, thanks for clarifying.

chetan51:
@david-ragazzi Obrigado!

david-ragazzi:
> Ok, thanks for clarifying. 

आपका स्वागत है

david-ragazzi:
> @david-ragazzi Obrigado!

Por nada!

Issue #334, opened 2013-10-25
Update Pylint


rhyolight:
Add things like:
- 3 newlines before top level classes and functions, 2 before methods
- Check for copyright notice?
- Add string functions like string.join to the disallowed functions list.
- spaces after commas
- Files must end with exactly one newline character.
- No tabs should be used anywhere.
- prefer double quotes (not sure exact implementation for that)
- Wrapping arguments. If some are left on first line then match the indentation of the first one on subsequent lines.  If none are left on the first line then add one or two levels of indentation (two levels are required when necessary to differentiate between a new scope on the following line, like arguments in a function definition).

PR #2088, opened 2015-04-27
Match pylintrc to Numenta internal, partial resolution for #334

jcasner:
@scottpurdy @rhyolight please review - not urgent at all!

This is a partial resolution for #334


scottpurdy:
:+1:

Issue #2090, opened 2015-04-28
Add capnp serialization to Python CLAClassifier


scottpurdy:
Related to numenta/nupic.core#402

See parent #1449

PR #2099, opened 2015-04-30
Port ClaClassifier to capnp

david-ragazzi:
Fixes: https://github.com/numenta/nupic/issues/2090

### DIFF
--- a/nupic/algorithms/CLAClassifier.py
+++ b/nupic/algorithms/CLAClassifier.py
@@ -213,6 +213,36 @@ class BitHistory(object):
     self._version = BitHistory.__VERSION__
 
 
+  def write(self, proto):
+    proto.id = self._id
+
+    statsProto = proto.init("stats", len(self._stats))
+    for (bucketIdx, dutyCycle) in enumerate(self._stats):
+      statsProto[bucketIdx].index = bucketIdx
+      statsProto[bucketIdx].dutyCycle = dutyCycle
+
+    proto.lastTotalUpdate = self._lastTotalUpdate
+    proto.learnIteration = self._learnIteration
+
+
+  @classmethod
+  def read(cls, proto):
+    bitHistory = object.__new__(cls)
+
+    bitHistory._id = proto.id
+
+    for statProto in proto.stats:
+      statsLen = len(bitHistory._stats) - 1
+      if statProto.index > statsLen:
+        bitHistory._stats.extend(itertools.repeat(0.0, statProto.index - statsLen))
+      bitHistory._stats[statProto.index] = statProto.dutyCycle
+
+    bitHistory._lastTotalUpdate = proto.lastTotalUpdate
+    bitHistory._learnIteration = proto.learnIteration
+
+    return bitHistory
+
+
 
 class CLAClassifier(object):
   """
@@ -510,3 +540,95 @@ class CLAClassifier(object):
       pass
 
     self._version = CLAClassifier.__VERSION__
+
+
+  @classmethod
+  def read(cls, proto):
+    classifier = object.__new__(cls)
+
+    classifier.steps = []
+    for step in proto.steps:
+      classifier.steps.append(step)
+
+    classifier.alpha = proto.alpha
+    classifier.actValueAlpha = proto.actValueAlpha
+    classifier._learnIteration = proto.learnIteration
+    classifier._recordNumMinusLearnIteration = proto.recordNumMinusLearnIteration
+
+    classifier._patternNZHistory = deque(maxlen=max(classifier.steps) + 1)
+    patternNZHistoryProto = proto.patternNZHistory
+    learnIteration = classifier._learnIteration - len(patternNZHistoryProto) + 1
+    for i in xrange(len(patternNZHistoryProto)):
+      classifier._patternNZHistory.append((learnIteration, list(patternNZHistoryProto[i])))
+      learnIteration += 1
+
+    classifier._activeBitHistory = dict()
+    activeBitHistoryProto = proto.activeBitHistory
+    for i in xrange(len(activeBitHistoryProto)):
+      stepBitHistories = activeBitHistoryProto[i]
+      nSteps = stepBitHistories.steps
+      for indexBitHistoryProto in stepBitHistories.bitHistories:
+        bit = indexBitHistoryProto.index
+        bitHistory = BitHistory.read(indexBitHistoryProto.history)
+        classifier._activeBitHistory[(bit, nSteps)] = bitHistory
+
+    classifier._maxBucketIdx = proto.maxBucketIdx
+
+    classifier._actualValues = []
+    for actValue in proto.actualValues:
+      if actValue == 0:
+        classifier._actualValues.append(None)
+      else:
+        classifier._actualValues.append(actValue)
+
+    classifier._version = proto.version
+    classifier.verbosity = proto.verbosity
+
+    return classifier
+
+
+  def write(self, proto):
+    stepsProto = proto.init("steps", len(self.steps))
+    for i in xrange(len(self.steps)):
+      stepsProto[i] = self.steps[i]
+
+    proto.alpha = self.alpha
+    proto.actValueAlpha = self.actValueAlpha
+    proto.learnIteration = self._learnIteration
+    proto.recordNumMinusLearnIteration = self._recordNumMinusLearnIteration
+
+    patternNZHistory = []
+    for (iteration, learnPatternNZ) in self._patternNZHistory:
+      patternNZHistory.append(learnPatternNZ)
+    proto.patternNZHistory = patternNZHistory
+
+    i = 0
+    activeBitHistoryProtos = proto.init("activeBitHistory", len(self._activeBitHistory))
+    if len(self._activeBitHistory) > 0:
+      for nSteps in self.steps:
+        stepBitHistory = {bit: self._activeBitHistory[(bit, step)]
+                          for (bit, step) in self._activeBitHistory.keys()
+                          if step == nSteps}
+        stepBitHistoryProto = activeBitHistoryProtos[i]
+        stepBitHistoryProto.steps = nSteps
+        indexBitHistoryListProto = stepBitHistoryProto.init("bitHistories", len(stepBitHistory))
+        j = 0
+        for indexBitHistory in stepBitHistory:
+          indexBitHistoryProto = indexBitHistoryListProto[j]
+          indexBitHistoryProto.index = indexBitHistory
+          bitHistoryProto = indexBitHistoryProto.history
+          stepBitHistory[indexBitHistory].write(bitHistoryProto)
+          j += 1
+        i += 1
+
+    proto.maxBucketIdx = self._maxBucketIdx
+
+    actualValuesProto = proto.init("actualValues", len(self._actualValues))
+    for i in xrange(len(self._actualValues)):
+      if self._actualValues[i] is not None:
+        actualValuesProto[i] = self._actualValues[i]
+      else:
+        actualValuesProto[i] = 0
+
+    proto.version = self._version
+    proto.verbosity = self.verbosity
--- a/tests/unit/nupic/algorithms/cla_classifier_test.py
+++ b/tests/unit/nupic/algorithms/cla_classifier_test.py
@@ -28,8 +28,11 @@ import cPickle as pickle
 import unittest2 as unittest
 
 import numpy
+import tempfile
+import capnp
 
 from nupic.algorithms.CLAClassifier import CLAClassifier
+from nupic.bindings.proto import ClaClassifier_capnp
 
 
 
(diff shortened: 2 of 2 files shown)


david-ragazzi:
I'm pushing this because I can't build in my local machine and want see what Travis says. Python is not managing import `ClaClassifier_capn`.

tests/unit/nupic/algorithms/cla_classifier_test.py
```diff
@@ -28,7 +28,9 @@
 import unittest2 as unittest
 
 import numpy
+import tempfile
 
+from nupic.bindings.proto import ClaClassifier_capnp
```
david-ragazzi: @scottpurdy Do you know what is missing to `CLAClassifier.capnp` be imported?

david-ragazzi: Never mind.. I forgot add this to `algorihtms.i`:

`#include <nupic/proto/ClaClassifier.capnp.h>`

david-ragazzi: Not solved.. 

@scottpurdy What is missing? Could you help me?

scottpurdy: Seems right. Did you do a rebuild of nupic? And if you have it set up to use a local nupic.core, did you rebuild the latest from that first?

scottpurdy: OH you have to do `import capnp` before importing any schema. When you do `import capnp` it sets up the import hooks so that Python knows how to find and parse the `.capnp` files.

david-ragazzi: Thanks! This was the problem. Now I can test it locally..

david-ragazzi:
Almost there.. But some weird things happening.. For example, when I use:

`proto.alpha = self.alpha`

If self.alpha is `0.1`, the proto.alpha become `0.100000000149`. This makes `assertEqual` emits error.

david-ragazzi:
Almost there. All fields (except patternNZHistory) are being written and read with pycapn with success. Note that FastClassifier handle stepBitHistories in a different way (this delayed my work because I had to try understand it out and re-use the schema).

@chetan51 @oxtopus @scottpurdy Do you know how nested list like `patternNZHistory` should be initialized using pycapn? Being more specific, this line is presenting problems:

https://github.com/numenta/nupic/pull/2099/files#diff-9dc7913cbee348b16a4fb7b5bcb4a843R553

I check all internet but without success. :/

rhyolight:
FYI: I removed the `priority:1` label. I know we haven't defined these priorities well yet. But I would like to use them in this fashion:

- `priority:1`: **Must fix now!** Bug or deficiency preventing the proper installation or runtime of NuPIC for most users.
- `priority:2`: **Must fix soon.** Bug that prevents proper usage of NuPIC for a minority of users. New feature that is important to work on soon, or that is blocking the progress of other issues.
- `priority:3`: **Normal priority.** Things that should be done, but nothing pressing.
- `priority:4`: **Low priority.** Work that would be "nice to have", but might never get done and no one would care that much.

chetan51:
@david-ragazzi That line should instead be:

    patternProto = patternNZHistoryProtos[i].init(len(pattern))

nupic/algorithms/CLAClassifier.py
```diff
@@ -213,6 +213,31 @@ def __setstate__(self, state):
     self._version = BitHistory.__VERSION__
 
 
+  def write(self, proto):
+    proto.id = self._id
+
+    i = 0
+    statsProto = proto.init("stats", len(self._stats))
+    for (bucketIdx, dc) in enumerate(self._stats):
```
scottpurdy: nit: spell out `dutyCycle` and parens are optional when unpacking

nupic/algorithms/CLAClassifier.py
```diff
@@ -213,6 +213,31 @@ def __setstate__(self, state):
...
+      statsProto[i].index = bucketIdx
```
scottpurdy: You can use `bucketIdx` instead of `i` and then you don't need `i` at all

nupic/algorithms/CLAClassifier.py
```diff
@@ -213,6 +213,31 @@ def __setstate__(self, state):
...
+
+    proto.lastTotalUpdate = self._lastTotalUpdate
+    proto.learnIteration = self._learnIteration
+
+
+  def read(self, proto):
+    self._id = proto.id
+
+    self._stats = array.array("f")
+    for i in xrange(len(proto.stats)):
```
scottpurdy: Are capnp list readers iterable? IE does this work?

```
for statProto in proto.stats:
  self._stats[stat.index] = stat.dutyCycle
```

nupic/algorithms/CLAClassifier.py
```diff
@@ -244,7 +269,7 @@ class CLAClassifier(object):
   __VERSION__ = 2
 
 
-  def __init__(self, steps=(1,), alpha=0.001, actValueAlpha=0.3, verbosity=0):
+  def __init__(self, steps=[1,], alpha=0.001, actValueAlpha=0.3, verbosity=0):
```
scottpurdy: Undo this. Default parameters should be immutable whenever possible.

tests/unit/nupic/algorithms/cla_classifier_test.py
```diff
@@ -28,7 +28,10 @@
 import unittest2 as unittest
 
 import numpy
+import tempfile
+import capnp
 
+from nupic.bindings.proto import ClaClassifier_capnp
```
scottpurdy: nit: alphabetical order for each block of imports

tests/unit/nupic/algorithms/cla_classifier_test.py
```diff
@@ -238,6 +241,53 @@ def testSerialization(self):
...
 
 
+  def testWriteRead(self):
+    c1 = CLAClassifier([1], 0.1, 0.1, 0)
+
+    # Create a vector of input bit indices
+    input1 = [1, 5, 9]
+    result = c1.compute(recordNum=0,
+                        patternNZ=input1,
+                        classification={'bucketIdx': 4, 'actValue': 34.7},
```
scottpurdy: nit: double quotes

scottpurdy:
done with review, thanks for putting the PR together!

nupic/algorithms/CLAClassifier.py
david-ragazzi: > spell out dutyCycle

@scottpurdy I though this but `dc` is being everywhere in this code..

tests/unit/nupic/algorithms/cla_classifier_test.py
david-ragazzi: > nit: double quotes

This file uses only single quotes (check `testMissingRecords`, for example). So I think that if we use double quotes here, then we should fix the other occurrences..

scottpurdy: K fine to leave single quotes for this PR

nupic/algorithms/CLAClassifier.py
scottpurdy: I'd prefer at least this occurrence use `dutyCycle`. If you want to fix existing occurrences in this file that is fine as well, although might be easier to get this PR merged first and send one just for fixing existing occurrences.

tests/unit/nupic/algorithms/cla_classifier_test.py
david-ragazzi: > nit: alphabetical order for each block of imports

Done.

nupic/algorithms/CLAClassifier.py
david-ragazzi: > I'd prefer at least this occurrence use dutyCycle. If you want to fix existing occurrences in this file that is fine as well, although might be easier to get this PR merged first and send one just for fixing existing occurrences.

Done.

nupic/algorithms/CLAClassifier.py
david-ragazzi: > You can use bucketIdx instead of i and then you don't need i at all

Done.

nupic/algorithms/CLAClassifier.py
david-ragazzi: > Are capnp list readers iterable? IE does this work?

Yes, it works. Changed.

nupic/algorithms/CLAClassifier.py
david-ragazzi: > Undo this. Default parameters should be immutable whenever possible.

Undone.

david-ragazzi:
@scottpurdy All tests passing in my local machine. Now, waiting Travis.. 

Ah.. I improved some `asserts` in `testWriteRead` in order to tests be more exact.

david-ragazzi:
Travis is green for Linux and red for OS X because `timeout`.. Just restart the OS X build that everything going be green.

david-ragazzi:
@scottpurdy Travis is green.

nupic/algorithms/CLAClassifier.py
```diff
@@ -213,6 +213,31 @@ def __setstate__(self, state):
     self._version = BitHistory.__VERSION__
 
 
+  def write(self, proto):
+    proto.id = self._id
+
+    statsProto = proto.init("stats", len(self._stats))
```
oxtopus: I don't believe you need to manually init `stats` and enumerate.  There's a performance hit, so if we can avoid the enumeration, that would be better.

nupic/algorithms/CLAClassifier.py
```diff
@@ -213,6 +213,31 @@ def __setstate__(self, state):
...
+      statsProto[bucketIdx].dutyCycle = dutyCycle
+
+    proto.lastTotalUpdate = self._lastTotalUpdate
+    proto.learnIteration = self._learnIteration
+
+
+  def read(self, proto):
+    self._id = proto.id
+
+    for statProto in proto.stats:
```
oxtopus: Complement to above comment re: `stats` enumeration.

nupic/algorithms/CLAClassifier.py
david-ragazzi: Do you mean that I can simply use this:

`proto.stats = self._stats`

oxtopus: Since `self._stats` is an array, you'll need to get a python `list`: `proto.stats = self._stats.tolist()` and `self._stats = array.array("f", proto.stats)`

oxtopus: Actually, never mind.  At least not without modifying the BitHistory.capnp schema to match.  @scottpurdy What do you think?

david-ragazzi: BitHistory of c++ and python implemented in different fashion. Either python classifier adapts itself to re-use BitHistory schema from c++ (or vice-versa) or than change python bit histories to have the same architecture of its c++ version.

scottpurdy: So it looks like the C++ stores `stats` internally as a map while the Python stores it as an array. We should probably figure out which one makes sense and standardize it. I think the enumeration is fine for now.

nupic/algorithms/CLAClassifier.py
```diff
@@ -213,6 +213,31 @@ def __setstate__(self, state):
...
+    statsProto = proto.init("stats", len(self._stats))
+    for (bucketIdx, dutyCycle) in enumerate(self._stats):
+      statsProto[bucketIdx].index = bucketIdx
+      statsProto[bucketIdx].dutyCycle = dutyCycle
+
+    proto.lastTotalUpdate = self._lastTotalUpdate
+    proto.learnIteration = self._learnIteration
+
+
+  def read(self, proto):
```
scottpurdy: The examples on the wiki page might be out of date for this but I think we are going to implement `read` functions as class methods that construct a new object without initialization with `__new__` and then load everything from the proto. See this example:

https://github.com/numenta/nupic/blob/6c78a70ccbba9e642864b2d9d81714f19b1cb7e1/nupic/encoders/coordinate.py#L195

scottpurdy:
@oxtopus - I'm not sure how to optimize the parts you commented on but if you see a way then please propose it to David.

@david-ragazzi - this  is looking pretty close. Last request from me is to change the `read` methods to class methods. See the link in my comment.

nupic/algorithms/CLAClassifier.py
david-ragazzi: Yeah, I like this because it ensures that we not forget serialize future fields..The problem is that when I create a new instance instead of re-use the current, it simply prevents me of access some fields. Do you know this problem?

nupic/algorithms/CLAClassifier.py
oxtopus: It's `map<int, Real64>` in c++, which is fine if there's not a contiguous range of keys.  If there is, we should switch it to be an array in c++.  If not, we should adapt the Python implementation to use a dict or other mapping for consistency.

nupic/algorithms/CLAClassifier.py
scottpurdy: No I don't quite follow

nupic/algorithms/CLAClassifier.py
scottpurdy: It isn't necessarily contiguous so probably should be a map. Actually for RDSE the bucket indices can be positive or negative I believe so it would have to be a map.

https://github.com/numenta/nupic/issues/2106

oxtopus:
Is #2106 a blocker for this?  Resolution of that would certainly affect this implementation.

scottpurdy:
@oxtopus - I was not considering #2106 a blocker for this.

nupic/algorithms/CLAClassifier.py
david-ragazzi: > No I don't quite follow

No problem.. I solved it. I had to call `classifier = cls()` instead of `classifier = object.__new__(cls)`. It seems that `__new__` doesn't call `__init__` and so many fields are not created.

david-ragazzi:
> @david-ragazzi - this is looking pretty close. Last request from me is to change the read methods to class methods. See the link in my comment.

@scottpurdy Done. Merge after :green_apple: ?

nupic/algorithms/CLAClassifier.py
oxtopus: The original suggestion to use `__new__` as opposed to `__init__` was deliberate.  There is overlap in functionality here between `CLAClassifier.__init__()` and `CLAClassifier.read()`.  In Python, `__init__()` is not a traditional _constructor_, it _initializes_ an already existing _instance_ (created with `object.__new__(cls)`).

Since we're working with serialization, the object was already technically initialized before it was serialized, and then by calling `read()`, you're re-initing the object and then overwriting the values from the serialized object.

oxtopus:
:-1: please don't merge until `read()` properly deserializes the object.

david-ragazzi:
> please don't merge until read() properly deserializes the object.

But it is doing this, isn't?

oxtopus:
@david-ragazzi please see my comment regarding `__init__()`, `cls()`, and `__new__()`.

`read()` should create the instance with `object.__new__(cls)`, read the attrs, and if there are other attrs that would otherwise normally be set in `__init__()` (but not preserved between `read()` and `write()`), those attrs need to be set in `read()`.  Preferably, that can be done with a method (or set of methods) that can be shared between `__init__()` and `read()`.

david-ragazzi:
@oxtopus I just set manually `patternNZHistory` in `read()` and reverted to `object.__new__()`. Thanks for clarify. 

Ok for merge after Travis is green?

oxtopus:
That's better.  Thanks!  I was in the process of submitting a PR, but you beat me to it ;)

You should also make `BitHistory.read()` a `classmethod` before merging.

david-ragazzi:
> Thanks! I was in the process of submitting a PR, but you beat me to it ;)

:smile: 

> You should also make BitHistory.read() a classmethod before merging.

Oh yes.. I forgot.. I did right now.

Issue #2098, opened 2015-04-29
Add Cap'n Proto serialization to Python Connections


chetan51:

chetan51:
Depends on https://github.com/numenta/nupic/issues/2097.

PR #2100, opened 2015-05-01
Add Cap'n Proto serialization to Python Connections

chetan51:
Fixes https://github.com/numenta/nupic/issues/2098.

Depends on https://github.com/numenta/nupic.core/pull/423.

### DIFF
--- a/nupic/research/connections.py
+++ b/nupic/research/connections.py
@@ -281,6 +281,62 @@ class Connections(object):
     return connections
 
 
+  def __eq__(self, other):
+    """
+    Equality operator for Connections instances.
+    Checks if two instances are functionally identical
+    (might have different internal state).
+
+    @param other (Connections) Connections instance to compare to
+    """
+    if self.numCells != other.numCells: return False
+
+    for cell in xrange(self.numCells):
+      segmentSet = set()
+      for segment in self.segmentsForCell(cell):
+        synapseSet = self._synapseSetForSynapses(
+                       self.synapsesForSegment(segment))
+        segmentSet.add(synapseSet)
+
+      otherSegmentSet = set()
+      for segment in other.segmentsForCell(cell):
+        otherSynapseSet = other._synapseSetForSynapses(
+                       other.synapsesForSegment(segment))
+        otherSegmentSet.add(otherSynapseSet)
+
+      if segmentSet != otherSegmentSet: return False
+
+      synapseSet = self._synapseSetForSynapses(
+                     self.synapsesForPresynapticCell(cell))
+
+      otherSynapseSet = other._synapseSetForSynapses(
+                    other.synapsesForPresynapticCell(cell))
+
+      if synapseSet != otherSynapseSet: return False
+
+    return True
+
+
+  def _synapseSetForSynapses(self, synapses):
+    """
+    Returns a set containing synapse data for synapses.
+    Rounds synapse permanence values for comparison.
+    (Helper method used in __eq__.)
+
+    @param synapses (set) Synapse indicies
+
+    @return (set) Synapse data set
+    """
+    synapseSet = set()
+
+    for synapse in synapses:
+      synapseData = self.dataForSynapse(synapse)
+      synapseSet.add((synapseData.presynapticCell,
+                     round(synapseData.permanence, 7)))
+
+    return frozenset(synapseSet)
+
+
   def _validateCell(self, cell):
     """
     Raises an error if cell index is invalid.
--- a/nupic/research/connections.py
+++ b/nupic/research/connections.py
@@ -238,6 +238,54 @@ class Connections(object):
     return len(self._synapses)
 
 
+  def write(self, proto):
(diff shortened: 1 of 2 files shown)


chetan51:
@scottpurdy

rhyolight:
> E   ImportError: cannot import name ConnectionsProto_capnp

chetan51:
@rhyolight That's because numenta/nupic.core#423 needs to be merged first.

rhyolight:
@chetan51 Oh! Well it is ready to merge :smile:

nupic/research/connections.py
```diff
@@ -238,6 +238,54 @@ def numSynapses(self):
...
+          protoSynapse.permanence = synapseData.permanence
+
+          k += 1
+
+        j += 1
+
+      i += 1
+
+
+  def read(self, proto):
```
scottpurdy: Switch to classmethod and create instance with `__new__` like this:
https://github.com/numenta/nupic/blob/6c78a70ccbba9e642864b2d9d81714f19b1cb7e1/nupic/encoders/category.py#L239

nupic/research/connections.py
```diff
@@ -238,6 +238,54 @@ def numSynapses(self):
...
+      for j in xrange(len(protoSegments)):
+        protoSegment = protoSegments[j]
+        protoSynapses = protoSegment.synapses
+        segment = self.createSegment(i)
+
+        for k in xrange(len(protoSynapses)):
+          protoSynapse = protoSynapses[k]
+          synapse = self.createSynapse(segment,
+                                       int(protoSynapse.presynapticCell),
+                                       float(protoSynapse.permanence))
```
scottpurdy: do you really need to cast these? The permanence should be a float already. The other may be a `long`

nupic/research/connections.py
```diff
@@ -238,6 +238,54 @@ def numSynapses(self):
     return len(self._synapses)
 
 
+  def write(self, proto):
+    protoCells = proto.init('cells', self.numCells)
+
+    i = 0
```
scottpurdy: don't need `i` since it is same as `cell`

nupic/research/connections.py
```diff
@@ -238,6 +238,54 @@ def numSynapses(self):
...
+    for cell in xrange(self.numCells):
+      segments = self.segmentsForCell(cell)
+      protoSegments = protoCells[cell].init('segments', len(segments))
+
+      j = 0
```
scottpurdy: can you use enumerate in the for statement to avoid maintaining the index yourself? Same for `k` below

scottpurdy:
done with review

chetan51:
I just discovered a bug with the code in this PR, and added a failing test to demonstrate it. Do not merge until I have fixed this.

chetan51:
Fixed the bug, this PR is ready for review. @scottpurdy

nupic/research/connections.py
```diff
@@ -238,6 +238,105 @@ def numSynapses(self):
...
+
+          protoSynapse.presynapticCell = synapseData.presynapticCell
+          protoSynapse.permanence = synapseData.permanence
+
+
+  @classmethod
+  def read(cls, proto):
+    protoCells = proto.cells
+    connections = object.__new__(cls)
+    connections.__init__(len(protoCells))
```
scottpurdy: If you want to call `__init__` then just replace the `__new__` and `__init__` calls with a regular instantiation of the class. That said, we have usually used `__new__` to avoid the initialization that happens in `__init__` since it is usually redundant to the loading from the proto object.

So I'd say it is probably better to avoid `__init__` unless it contains logic that is necessary for initialization even when loading from a checkpoint, in which case just do `connections = cls(len(protoCells))` or similar. Also, if you keep the `__new__` call then it should be `connections = cls.__new__()`

nupic/research/connections.py
```diff
@@ -238,6 +238,105 @@ def numSynapses(self):
...
+
+    @param other (Connections) Connections instance to compare to
+    """
+    if self.numCells != other.numCells: return False
+
+    for cell in xrange(self.numCells):
+      segmentSet = set()
+      for segment in self.segmentsForCell(cell):
+        synapseSet = self._synapseSetForSynapses(
+                       self.synapsesForSegment(segment))
```
scottpurdy: nit: The indentation in this case should be one or two indents in from the start of the previous line.

nupic/research/connections.py
```diff
@@ -238,6 +238,105 @@ def numSynapses(self):
...
+    @return (set) Synapse data set
+    """
+    synapseSet = set()
+
+    for synapse in synapses:
+      synapseData = self.dataForSynapse(synapse)
+      synapseSet.add((synapseData.presynapticCell,
+                     round(synapseData.permanence, 7)))
+
+    return frozenset(synapseSet)
```
scottpurdy: Why turn into a `frozenset`? Seems unnecessary and adds a little overhead

scottpurdy:
Done with review

Issue #2089, opened 2015-04-28
Add Cap'n Proto serialization to Python Temporal Memory


scottpurdy:
See parent issue #1449 

Depends on #2097 and #2098.

It may be better to put the schema file in nupic.core from the start even if the C++ implementation isn't there yet. That way we don't have to go and move it over later.

chetan51:
Depends on https://github.com/numenta/nupic/issues/2097 and https://github.com/numenta/nupic/issues/2098.

rcrowder:
Let me know if there is anything you need. I'm currently working in a Ubuntu VM to test out the SWIG and Python unit tests.

PR #2112, opened 2015-05-08
Add Cap'n Proto serialization to Python Temporal Memory

chetan51:
Fixes https://github.com/numenta/nupic/issues/2089.

Depends on https://github.com/numenta/nupic/pull/2100 and https://github.com/numenta/nupic.core/pull/426.

### DIFF
--- a/nupic/research/temporal_memory.py
+++ b/nupic/research/temporal_memory.py
@@ -628,6 +628,11 @@ class TemporalMemory(object):
 
 
   def write(self, proto):
+    """
+    Writes serialized data to proto object
+
+    @param proto (DynamicStructBuilder) Proto object
+    """
     proto.columnDimensions = self.columnDimensions
     proto.cellsPerColumn = self.cellsPerColumn
     proto.activationThreshold = self.activationThreshold
@@ -649,6 +654,13 @@ class TemporalMemory(object):
 
   @classmethod
   def read(cls, proto):
+    """
+    Reads deserialized data from proto object
+
+    @param proto (DynamicStructBuilder) Proto object
+
+    @return (TemporalMemory) TemporalMemory instance
+    """
     tm = object.__new__(cls)
 
     tm.columnDimensions = list(proto.columnDimensions)
--- a/nupic/research/temporal_memory.py
+++ b/nupic/research/temporal_memory.py
@@ -673,6 +673,50 @@ class TemporalMemory(object):
     return tm
 
 
+  def __eq__(self, other):
+    """
+    Equality operator for TemporalMemory instances.
+    Checks if two instances are functionally identical
+    (might have different internal state).
+
+    @param other (TemporalMemory) TemporalMemory instance to compare to
+    """
+    epsilon = 0.0000001
+
+    if self.columnDimensions != other.columnDimensions: return False
+    if self.cellsPerColumn != other.cellsPerColumn: return False
+    if self.activationThreshold != other.activationThreshold: return False
+    if abs(self.initialPermanence - other.initialPermanence) > epsilon:
+      return False
+    if abs(self.connectedPermanence - other.connectedPermanence) > epsilon:
+      return False
+    if self.minThreshold != other.minThreshold: return False
+    if self.maxNewSynapseCount != other.maxNewSynapseCount: return False
+    if abs(self.permanenceIncrement - other.permanenceIncrement) > epsilon:
+      return False
+    if abs(self.permanenceDecrement - other.permanenceDecrement) > epsilon:
+      return False
+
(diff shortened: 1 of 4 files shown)


chetan51:
@scottpurdy Ready for review.

tests/unit/nupic/research/temporal_memory_test.py
```diff
@@ -578,6 +585,76 @@ def testMapCellsToColumns(self):
...
+      connectedPermanence=0.58,
+      minThreshold=4,
+      maxNewSynapseCount=18,
+      permanenceIncrement=0.23,
+      permanenceDecrement=0.08,
+      seed=91
+    )
+
+    # Run some data through before serializing
+    # TODO
```
scottpurdy: TODO?

david-ragazzi:
@chetan51 I love this style, i.e. use descriptive names instead of non-common names like `permanence` instead of `perm`. I'm ok with very know acronyms like `num`, `max`, `min`, `prc`, etc.

nupic/research/temporal_memory.py
```diff
@@ -627,6 +627,52 @@ def mapCellsToColumns(self, cells):
...
+    tm.initialPermanence = proto.initialPermanence
+    tm.connectedPermanence = proto.connectedPermanence
+    tm.minThreshold = int(proto.minThreshold)
+    tm.maxNewSynapseCount = int(proto.maxNewSynapseCount)
+    tm.permanenceIncrement = proto.permanenceIncrement
+    tm.permanenceDecrement = proto.permanenceDecrement
+
+    tm.connections = Connections.read(proto.connections)
+    tm._random = Random()
+    tm._random.read(proto.random)
```
chetan51: @scottpurdy Is this alright?

chetan51:
@david-ragazzi It's so much clearer to read that way :)

chetan51:
@scottpurdy Do you know what I need to do to expose the capnproto schema file from nupic.core for Connections here?

nupic/research/temporal_memory.py
scottpurdy: Yes

scottpurdy:
The connections schema should be in nupic/bindings/proto/ but I don't see a TM schema anywhere.

tests/unit/nupic/research/temporal_memory_test.py
```diff
@@ -578,6 +585,73 @@ def testMapCellsToColumns(self):
...
+    with tempfile.TemporaryFile() as f:
+      proto1.write(f)
+      f.seek(0)
+      proto2 = TemporalMemoryProto_capnp.TemporalMemoryProto.read(f)
+
+    # Load the deserialized proto
+    tm2 = TemporalMemory.read(proto2)
+
+    # Check that the two temporal memory objects have the same attributes
+    self.assertEqual(tm1.columnDimensions, tm2.columnDimensions)
```
scottpurdy: @chetan51 - do you think it would be better to define a `__eq__` method on TemporalMemory that makes these checks?

tests/unit/nupic/research/temporal_memory_test.py
```diff
@@ -578,6 +585,73 @@ def testMapCellsToColumns(self):
...
+    self.assertEqual(tm1.connections, tm2.connections)
+
+    self.assertEqual(tm1.activeCells, tm2.activeCells)
+    self.assertEqual(tm1.predictiveCells, tm2.predictiveCells)
+    self.assertEqual(tm1.winnerCells, tm2.winnerCells)
+
+    # Run a couple records through after deserializing and check results match
+    tm1.compute(self.patternMachine.get(0))
+    tm2.compute(self.patternMachine.get(0))
+    self.assertEqual(tm1.activeCells, tm2.activeCells)
```
scottpurdy: I'm not sure the types of each of these but am guessing in some of these cases that `assertSequenceEquals` or other assert would be more appropriate

scottpurdy:
done reviewing this

tests/unit/nupic/research/temporal_memory_test.py
chetan51: They're `set`s, so `assertEqual` should be fine right?

tests/unit/nupic/research/temporal_memory_test.py
chetan51: Sure, I'll do that.

tests/unit/nupic/research/temporal_memory_test.py
scottpurdy: `assertEqual` does the same thing as `assertSetEqual` IF they are sets but the later is good practice to enforce that they are sets if that is the API

PR #2136, opened 2015-05-15
Switch over to C++ SpatialPooler where possible to speed up tests/build.

### LINKED ISSUE #2128: Switch tests/experiments to C++ SpatialPooler
NuPIC tests are running much slower since the switch to Python SpatialPooler. In general, we should use `nupic.bindings.algorithms.SpatialPooler` everywhere unless we are specifically testing the Python version. This should speed up the build/tests a lot.

Part of this change is benchmarking the tests from SHA (@rhyolight please put SHA here from before change to new SP), SHA adf01ae4 (after switch to new Python SP), and the version after switching to the new C++ version.

scottpurdy:
fixes #2128

### DIFF
--- a/examples/opf/experiments/anomaly/temporal/simple/description.py
+++ b/examples/opf/experiments/anomaly/temporal/simple/description.py
@@ -157,7 +157,7 @@ config = {
             # 0: silent; >=1: some info; >=2: more info;
             'spVerbosity' : 0,
 
-            'spatialImp' : 'py',
+            'spatialImp' : 'cpp',
 
             'globalInhibition': 1,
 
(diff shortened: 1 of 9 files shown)


david-ragazzi:
Faster now (20~25 min for Linux).. I'm ignoring OS X builds because always present some problems..

scottpurdy:
@rhyolight - Does this look like it fixes the problem to you?

scottpurdy:
I think local timing would be more accurate than Travis.

david-ragazzi:
> I think local timing would be more accurate than Travis.

Agreed. Travis timing depends of its queue requests.

scottpurdy:
@chetan51 please review

rhyolight:
> @rhyolight - Does this look like it fixes the problem to you?

Hell yes. Much, much better. :+1:

rhyolight:
@chetan51 A quick turnaround on this review would be nice, because we are still getting build timeouts.

chetan51:
:+1:

PR #2142, opened 2015-05-18
Use different logic for determining whether or not to translate back …

### LINKED ISSUE #2115: Possible bug in c++ CLAClassifierRegion
Hi,

This is against HEAD/ ec270f3918012304263db557d969c412bb705ea5

I've constructed a simple NontemporalClassification model, with a single input. Both the input and the output classifier are of type "string".

It's trained with:

```
for r in range(1000):
	ip = str( random_integers( low=0, high=99 ) );
	res = model.run( { "input": "A" + ip, "output": "Z" + ip } )
	print "A"+ip,  res.inferences["multiStepBestPredictions"][0]
```
The output is correctly of string form ("Zxx")

However, when classification is attempted:
```
	ip = str( random_integers( low=0, high=99 ) );
	res=model.run( { "input": "A" + str(ip), "output": None } )
	print  res.inferences["multiStepBestPredictions"][0], res
```
The returned category is a float.

When I change the CLAClassifierRegion implementation from cpp to py, I recover the expected behaviour of a restring response.

If I were to hazzard a guess, I'd say the Classifier indexes the  category string and, in this case, is returning the index rather than the dereferenced value.

oxtopus:
…into actual values from bucket indices

Fixes #2115

### DIFF
--- a/nupic/frameworks/opf/clamodel.py
+++ b/nupic/frameworks/opf/clamodel.py
@@ -396,7 +396,7 @@ class CLAModel(Model):
       inferences = self._multiStepCompute(rawInput=inputRecord)
     # For temporal classification. Not used, and might not work anymore
     elif self._isClassificationModel():
-      inferences = self._classifcationCompute()
+      inferences = self._classificationCompute()
 
     results.inferences.update(inferences)
 
@@ -534,7 +534,7 @@ class CLAModel(Model):
                                         rawInput=rawInput)
 
 
-  def _classifcationCompute(self):
+  def _classificationCompute(self):
     inference = {}
     classifier = self._getClassifierRegion()
     classifier.setParameter('inferenceMode', True)
--- a/tests/unit/nupic/algorithms/cla_classifier_test.py
+++ b/tests/unit/nupic/algorithms/cla_classifier_test.py
@@ -25,6 +25,7 @@
 CL_VERBOSITY = 0
 
 import cPickle as pickle
+import types
 import unittest2 as unittest
 
 import numpy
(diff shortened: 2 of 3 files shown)


nupic/frameworks/opf/clamodel.py
```diff
@@ -534,7 +534,7 @@ def _multiStepCompute(self, rawInput):
                                         rawInput=rawInput)
 
 
-  def _classifcationCompute(self):
```
rhyolight: talk about broken windows... :confounded:

oxtopus:
@scottpurdy Mind taking a look?

scottpurdy:
Any tests we could update or add to get better coverage of this?

oxtopus:
I think so.  Something along the lines of https://github.com/numenta/nupic/blob/master/tests/unit/nupic/algorithms/cla_classifier_test.py#L182-L191 where it alternates between sending a value and not and comparing the return type.

oxtopus:
@scottpurdy https://github.com/oxtopus/nupic/commit/b068142cfa76b0b38b9713a98a82b1c67de374b2 adds coverage

tests/unit/nupic/algorithms/cla_classifier_test.py
```diff
@@ -190,6 +191,13 @@ def testComputeCategory(self):
     self.assertSetEqual(set(result.keys()), set(('actualValues', 1)))
     self.assertEqual(result['actualValues'][4], 'D')
 
+    predictResult = c.compute(recordNum=2, patternNZ=[1, 5, 9],
+                              classification={'bucketIdx': 5,
+                                              'actValue': None},
+                              learn=True, infer=True)
+    for value in predictResult['actualValues']:
+      self.assertIsInstance(value, (types.NoneType, types.StringType))
```
scottpurdy: Can't you assert exactly what the sequence will look like rather than just that all of the elements are of one of the two types?

scottpurdy:
:+1: - had one comment on the test but take it or leave it

tests/unit/nupic/algorithms/cla_classifier_test.py
oxtopus: You can, and I thought about it and concluded that it didn't exactly capture the intent of the test.  The original failure mode was that we were getting floats and None back instead of strs, so I wanted the assertion to match.

PR #2149, opened 2015-05-20
Anomaly cleanups

### LINKED ISSUE #2146: Anomaly tidbits
small cherry-picks for Anomaly from PRs I intend to close..

breznak:
Fixes #2146

### DIFF
--- a/nupic/algorithms/anomaly.py
+++ b/nupic/algorithms/anomaly.py
@@ -164,6 +164,17 @@ class Anomaly(object):
     return "Anomaly:\tmode=%s\twindowSize=%r" % (self._mode, windowSize)
 
 
+  def __cmp__(self, other):
+    if not isinstance(other, Anomaly): 
+      return -1
+    if (other._mode == self._mode and
+        other._binaryThreshold == self._binaryThreshold and
+        other._movingAverage == self._movingAverage):
+      return 0 #equal
+    else:
+      return -1
+
+
   def __setstate__(self, state):
     """deserialization"""
     self.__dict__.update(state)
--- a/nupic/algorithms/anomaly.py
+++ b/nupic/algorithms/anomaly.py
@@ -108,7 +108,7 @@ class Anomaly(object):
        binaryAnomalyThreshold >= 1.0  or \
        binaryAnomalyThreshold < 0.0:
       raise ValueError("Anomaly: binaryAnomalyThreshold must be from [0,1) "
-                       "0 for disabled.")
+                       "0 for disabled. It is %s" % (binaryAnomalyThreshold) )
 
 
   def compute(self, activeColumns, predictedColumns, 
--- a/nupic/algorithms/anomaly.py
+++ b/nupic/algorithms/anomaly.py
@@ -72,10 +72,11 @@ class Anomaly(object):
   _supportedModes = (MODE_PURE, MODE_LIKELIHOOD, MODE_WEIGHTED)
 
 
-  def __init__(self, slidingWindowSize=None, mode=MODE_PURE, binaryAnomalyThreshold=None):
+  def __init__(self, slidingWindowSize=0, mode=MODE_PURE, binaryAnomalyThreshold=0.0):
     """
     @param slidingWindowSize (optional) - how many elements are summed up;
-        enables moving average on final anomaly score; int >= 0
+        enables moving average on final anomaly score; int >= 0;
+        0 means disabled (default)
     @param mode (optional) - (string) how to compute anomaly;
         possible values are:
           - "pure" - the default, how much anomal the value is;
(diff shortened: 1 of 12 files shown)


breznak:
please review..should be simple cleanups/moving classes around

rhyolight:
Please review @scottpurdy

nupic/algorithms/anomaly.py
```diff
@@ -72,10 +72,11 @@ class Anomaly(object):
   _supportedModes = (MODE_PURE, MODE_LIKELIHOOD, MODE_WEIGHTED)
 
 
-  def __init__(self, slidingWindowSize=None, mode=MODE_PURE, binaryAnomalyThreshold=None):
+  def __init__(self, slidingWindowSize=0, mode=MODE_PURE, binaryAnomalyThreshold=0.0):
```
scottpurdy: Isn't `None` a better default for disabled?

nupic/algorithms/anomaly.py
```diff
@@ -87,26 +88,27 @@ def __init__(self, slidingWindowSize=None, mode=MODE_PURE, binaryAnomalyThreshol
...
     """
     self._mode = mode
-    if slidingWindowSize is not None:
+    if slidingWindowSize > 0:
       self._movingAverage = MovingAverage(windowSize=slidingWindowSize)
     else:
       self._movingAverage = None
 
-    if self._mode == Anomaly.MODE_LIKELIHOOD or self._mode == Anomaly.MODE_WEIGHTED:
+    if self._mode == Anomaly.MODE_LIKELIHOOD or \
```
scottpurdy: wrap the entire condition in parens instead of using a backslash:
https://www.python.org/dev/peps/pep-0008/#maximum-line-length

nupic/algorithms/anomaly.py
```diff
@@ -87,26 +88,27 @@ def __init__(self, slidingWindowSize=None, mode=MODE_PURE, binaryAnomalyThreshol
...
                        "Anomaly.MODE_PURE, Anomaly.MODE_LIKELIHOOD, "
                        "Anomaly.MODE_WEIGHTED; you used: %r" % self._mode)
     self._binaryThreshold = binaryAnomalyThreshold
-    if binaryAnomalyThreshold is not None and ( 
-          not isinstance(binaryAnomalyThreshold, float) or
-          binaryAnomalyThreshold >= 1.0  or 
-          binaryAnomalyThreshold <= 0.0 ):
-      raise ValueError("Anomaly: binaryAnomalyThreshold must be from (0,1) "
-                       "or None if disabled.")
+    if not isinstance(binaryAnomalyThreshold, float) or \
```
scottpurdy: I agree old was formatted oddly but stick with implicit line continuations inside parens

nupic/algorithms/anomaly.py
```diff
@@ -130,15 +132,15 @@ def compute(self, activeColumns, predictedColumns,
     if self._mode == Anomaly.MODE_PURE:
       score = anomalyScore
     elif self._mode == Anomaly.MODE_LIKELIHOOD:
-      if inputValue is None:
-        raise ValueError("Selected anomaly mode 'Anomaly.MODE_LIKELIHOOD' "
-                 "requires 'inputValue' as parameter to compute() method. ")
-
+      assert inputValue is not None, ("Selected anomaly mode "
```
scottpurdy: Don't use asserts when you really want to raise an exception. Assert statements are discarded when you run code in "optimized" mode so they are only appropriate for invariance checks that would be considered bugs if they were ever true.

nupic/algorithms/anomaly.py
```diff
@@ -164,6 +166,17 @@ def __str__(self):
     return "Anomaly:\tmode=%s\twindowSize=%r" % (self._mode, windowSize)
 
 
+  def __cmp__(self, other):
```
scottpurdy: What's the usage for this? I can imagine a use case for `__eq__` but `__cmp__` doesn't seem right.

nupic/utils.py
```diff
@@ -97,6 +97,17 @@ def __setstate__(self, state):
       self.slidingWindow = sum(self.slidingWindow)
 
 
+  def __cmp__(self, o):
```
scottpurdy: Again, this doesn't seem right.

nupic/utils.py
```diff
@@ -114,3 +125,48 @@ def write(self, proto):
     proto.windowSize = self.windowSize
     proto.slidingWindow = self.slidingWindow
     proto.total = self.total
+
+###########################################################################
+class CircularBuffer(object):
+  """
+  implementation of a fixed size constant random access circular buffer
+  #TODO can this be merged with MovingAverage somehow?
```
scottpurdy: Let's move this to a separate PR. The implementation of this class makes it look totally useless to me so I must not be understanding it correctly. We should take a look at the usage and probably replace with a regular Python list or `collections.deque`

nupic/algorithms/anomaly.py
breznak: In one case yes, but for swarming, it makes it much nicer if you can go from `0..N`

nupic/utils.py
breznak: I think this was the intention behind it: https://github.com/numenta/nupic/pull/2149#discussion-diff-31208896R132
But I agree it might be useless actually...this is just moving it from inner class somewhere to light in utils.py. An issue to investigate usefulness?

breznak:
Thanks guys for the review(s)! 
@rhyolight  I'll be AFK about a week, so I'll reply later on..

nupic/utils.py
```diff
@@ -26,7 +26,7 @@
 
 import numbers
 
-
+###################################################################################
```
david-ragazzi: Please remove this smudged line.

nupic/utils.py
```diff
@@ -114,3 +125,48 @@ def write(self, proto):
     proto.windowSize = self.windowSize
     proto.slidingWindow = self.slidingWindow
     proto.total = self.total
+
+###########################################################################
```
david-ragazzi: Please remove this smudged line.

nupic/algorithms/anomaly.py
scottpurdy: But then you're forced to swarm over parameters from `0..N` rather than the range you want, which may be `20..50`. I'd rather change the permute variable to have a `tryNone` option which determines whether it tries `None` in addition to some integer/float range.

nupic/utils.py
scottpurdy: Yeah I don't know if we should keep it or not. I'd just rather not complicate this PR with an unrelated task of moving this class around.

rhyolight:
@breznak Do you still plan to work on this?

PR #2167, opened 2015-05-31
Fixes #2092 - Port KNNClassifier serialization to capnproto

### LINKED ISSUE #2092: Add capnp serialization to Python KNN classifier
See parent #1449 

Depends on numenta/nupic.core#410

david-ragazzi:
Fixes https://github.com/numenta/nupic/issues/2092

### DIFF
--- a/nupic/algorithms/KNNClassifier.capnp
+++ b/nupic/algorithms/KNNClassifier.capnp
@@ -1,0 +1,40 @@
+@0x99e493bfb8fbad94;
+
+# Next ID: 31
+struct KNNClassifierProto {
+  verbosity @0 :UInt8;
+
+  cellsPerCol @1 :UInt32;
+  k @2 :UInt32;
+  distanceNorm @3 :Float32;
+  distanceMethod @4 :Text;
+  distanceThreshold @5 :UInt32;
+  doBinarization @6 :Bool;
+  binarizationThreshold @7 :Float32;
+  useSparseMemory @8 :Bool;
+  sparseThreshold @9 :Float32;
+  relativeThreshold @10 :Bool;
+  numWinners @11 :UInt32;
+  numSVDSamples @12 :Int32;
+  numSVDDims @13 :Int32;
+  fractionOfMax @14 :Float32;
+  replaceDuplicates @15 :Bool;
+  maxStoredPatterns @16 :Int32;
+
+  iterationIdx @17 :Int32;
+
+  numPatterns @18 :Int32;
+  memory @19 :List(List(Float32));
+  m @20 :List(List(Float32));
+  categoryList @21 :List(UInt32);
+  categoryRecencyList @22 :List(Int32);
+  partitionIdList @23 :List(Int32);
+  partitionIdArray @24 :List(Int32);
+
+  s @25 :List(Float64);
+  vt @26 :List(List(Float64));
+  mean @27 :List(Float64);
+
+  specificIndexTraining @28 :Bool;
+  nextTrainingIndices @29 :List(Int32);
+}
--- a/nupic/algorithms/KNNClassifier.py
+++ b/nupic/algorithms/KNNClassifier.py
@@ -143,6 +143,199 @@ class KNNClassifier(object):
     self.maxStoredPatterns = maxStoredPatterns
     self.clear()
 
+
+  @classmethod
+  def read(cls, proto):
+    knn = object.__new__(cls)
+
+    knn.version = KNNCLASSIFIER_VERSION
+    knn.verbosity = proto.verbosity
+
+    knn.cellsPerCol = proto.cellsPerCol
+    knn.k = proto.k
+    knn.distanceNorm = proto.distanceNorm
+    knn.distanceMethod = proto.distanceMethod
+    knn.distThreshold = proto.distanceThreshold
+    knn.doBinarization = proto.doBinarization
+    knn.binarizationThreshold = proto.binarizationThreshold
+    knn.useSparseMemory = proto.useSparseMemory
+    knn.sparseThreshold = proto.sparseThreshold
+    knn.relativeThreshold = proto.relativeThreshold
+    knn.numWinners = proto.numWinners
+    if proto.numSVDSamples != -1:
+      knn.numSVDSamples = proto.numSVDSamples
+    else:
+      knn.numSVDSamples = None
+    if proto.numSVDDims == 0:
+      knn._adaptiveSVDDims = True
+      knn.numSVDDims = 'adaptive'
+    else:
+      knn._adaptiveSVDDims = False
+      if proto.numSVDDims == -1:
+        knn.numSVDDims = None
+      else:
+        knn.numSVDDims = proto.numSVDDims
+    if int(proto.fractionOfMax) != -1:
+      knn.fractionOfMax = proto.fractionOfMax
+    else:
+      knn.fractionOfMax = None
+    knn.replaceDuplicates = proto.replaceDuplicates
+    knn.maxStoredPatterns = proto.maxStoredPatterns
+
+    knn.clear()
+
+    knn._iterationIdx = proto.iterationIdx
+
+    memory = numpy.array(proto.memory)
+    numRows = memory.shape[0]
+    numCols = memory.shape[1]
+    if knn.useSparseMemory:
+      knn._Memory = NearestNeighbor(numRows, numCols)
+    else:
+      knn._Memory = numpy.zeros((numRows, numCols))
+    for i in xrange(numRows):
+      for j in xrange(numCols):
+        knn._Memory[i,j] = memory[i][j]
+
+    if not knn.useSparseMemory:
+      m = numpy.array(proto.m)
+      numRows = m.shape[0]
+      numCols = m.shape[1]
+      knn._M = numpy.zeros((numRows, numCols))
+      for i in xrange(numRows):
+        for j in xrange(numCols):
+          knn._M[i,j] = m[i][j]
+
+    knn._numPatterns = proto.numPatterns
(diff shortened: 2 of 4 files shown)


david-ragazzi:
CC: @chetan51 @oxtopus @scottpurdy

nupic/algorithms/KNNClassifier.py
```diff
@@ -830,9 +1026,11 @@ def computeSVD(self, numSVDSamples=None, finalize=True):
     else:
       self._a = self._Memory.toDense()[:self._numPatterns]
 
-    self._mean = numpy.mean(self._a, axis=0)
+    self._mean = numpy.mean(self._a, axis=0).astype('float64')
```
david-ragazzi: I had to force this datatype because some times these variables were float64, another times were float32 because changes in `_Memory` variable (which alternate  in nupic.array and NearestNeighbor during a same iteration). These different datatypes for `_mean`, `_s`, and `_vt`, causes the assert test for compare two instances fails.

nupic/test/test_framework_helpers.py
```diff
@@ -64,3 +66,50 @@ def testSwarm(self):
       pass
   """
   return getattr(pytest.mark, tag)
+
```
david-ragazzi: This test is same that this one: https://github.com/david-ragazzi/nupic/commit/d78651f689f343ec817611bc32e51274f57dfb88

tests/unit/nupic/algorithms/knn_classifier_test.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python
```
david-ragazzi: There is not any unit test for KnnClassifier, only an integration test for test it together with PCA.

nupic/algorithms/KNNClassifier.capnp
```diff
@@ -0,0 +1,41 @@
+@0x99e493bfb8fbad94;
+
+# Next ID: ???
```
david-ragazzi: I still don't know what I shoud put here, i.e. the ID.

chetan51: It should be the next ID available in the following struct. In this case, it would be 31.

nupic/algorithms/KNNClassifier.capnp
```diff
@@ -0,0 +1,41 @@
...
+struct KNNClassifierProto {
+  version @0 :UInt16;
+  verbosity @1 :UInt8;
```
chetan51: Maybe we should make both of these types `:UInt32`, to keep things consistent.

nupic/algorithms/KNNClassifier.capnp
```diff
@@ -0,0 +1,41 @@
...
+  numWinners @12 :UInt32;
+  numSVDSamples @13 :Int32;
+  numSVDDims @14 :Int32;
+  fractionOfMax @15 :Float32;
+  replaceDuplicates @16 :Bool;
+  maxStoredPatterns @17 :Int32;
+
+  iterationIdx @18 :Int32;
+
+  numPatterns @19 :Int32;
```
chetan51: Why isn't this `:UInt32`?

nupic/algorithms/KNNClassifier.capnp
```diff
@@ -0,0 +1,41 @@
...
+  memory @20 :List(List(Float32));
+  m @21 :List(List(Float32));
+  categoryList @22 :List(UInt32);
+  categoryRecencyList @23 :List(Int32);
+  partitionIdList @24 :List(Int32);
+  partitionIdArray @25 :List(Int32);
```
chetan51: Should the above 3 data types be arrays of `Int32` or `UInt32`?

nupic/algorithms/KNNClassifier.capnp
```diff
@@ -0,0 +1,41 @@
...
+
+  s @26 :List(Float64);
+  vt @27 :List(List(Float64));
+  mean @28 :List(Float64);
```
chetan51: Do we need 64 bits for these 3?

nupic/algorithms/KNNClassifier.py
```diff
@@ -143,6 +143,201 @@ def __init__(self,
...
+    if proto.numSVDDims == 0:
+      knn._adaptiveSVDDims = True
+      knn.numSVDDims = 'adaptive'
+    else:
+      knn._adaptiveSVDDims = False
+      if proto.numSVDDims == -1:
+        knn.numSVDDims = None
+      else:
+        knn.numSVDDims = proto.numSVDDims
+    if proto.fractionOfMax != -1.:
```
chetan51: Could there be a floating point precision bug here, wherein this statement always returns False?

nupic/algorithms/KNNClassifier.py
```diff
@@ -143,6 +143,201 @@ def __init__(self,
...
+      knn._M = numpy.zeros((numRows, numCols))
+      for i in xrange(numRows):
+        for j in xrange(numCols):
+          knn._M[i,j] = m[i][j]
+
+    knn._numPatterns = proto.numPatterns
+    knn._categoryList = list(proto.categoryList)
+    knn._partitionIdList = list(proto.partitionIdList)
+    if list(proto.partitionIdArray) != [-2]:
+      if list(proto.partitionIdArray) != [-1]:
```
chetan51: Combine these `if` statements.

chetan51:
I've done a cursory review. This is the first time I'm seeing the KNNClassifier code, so perhaps someone else should take a look as well.

nupic/algorithms/KNNClassifier.capnp
david-ragazzi: hmmm... thanks!

nupic/algorithms/KNNClassifier.capnp
david-ragazzi: Not problem to me. I like consistency!

nupic/algorithms/KNNClassifier.capnp
david-ragazzi: Because `numPatterns` can be `None` which is represented by `-1` (which mismatch `UInt32`)

nupic/algorithms/KNNClassifier.capnp
david-ragazzi: Because blocks like this: 

https://github.com/david-ragazzi/nupic/blob/add_capnp_to_KnnClassifier/nupic/algorithms/KNNClassifier.py#L307

`[-2]` means that the member (still) doesn't exist.
`[-1]` means that the member exists and its value is `None`.
`[]` means that the member exists and its value is empty list.
`[x]` means that the member exists and its value is list with *x* elements.

nupic/algorithms/KNNClassifier.capnp
david-ragazzi: > Do we need 64 bits for these 3?

During iteration these variables change its sizes depending the value of `useSparseMemory`: if true it uses SparseMatrix (returning int32 elements), if false it uses a simple numpy.array (returning int64 elements). Finally, `s`, `vt` and `mean` are results of operations performed with the structure defined by `useSparseMemory`.

nupic/algorithms/KNNClassifier.py
david-ragazzi: Not sure. But we could mitigate this risk replacing this to if int(proto.fractionOfMax) != -1:

nupic/algorithms/KNNClassifier.py
david-ragazzi: I'll do.. Not remember why I do this.. hehe..

nupic/algorithms/KNNClassifier.capnp
scottpurdy: I don't think we should force everything to be 32 bits. The verbosity makes sense as UInt8.

The version can be dropped for now since it isn't needed with capnp

nupic/algorithms/KNNClassifier.capnp
scottpurdy: Is `0` a valid value or could that be used as the special case?

tests/unit/nupic/algorithms/knn_classifier_test.py
```diff
@@ -0,0 +1,88 @@
...
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2014, Numenta, Inc.  Unless you have purchased from
```
scottpurdy: it isn't 2014

tests/unit/nupic/algorithms/knn_classifier_test.py
scottpurdy: @BoltzmannBrain - any thoughts on organizing tests?

BoltzmannBrain: Yes the kNN classifier needs unit tests, and we also need to [redo the integration tests](https://github.com/numenta/nupic/issues/2172). The current integration tests do run and test the kNN under different scenarios, not just w/ PCA @david-ragazzi, but they're poorly written/formatted.

nupic/algorithms/KNNClassifier.capnp
david-ragazzi: We need -1 to None because assert function can fail.

david-ragazzi:
@chetan51 @scottpurdy Changes implemented.

david-ragazzi:
I'm closing this PR in favor of  numenta/nupic#2186 incorporate these changes to decrease dependency.

scottpurdy:
@david-ragazzi - I don't think 2186 was what you meant, it is for the region and this was for the algorithm. So now I'm not sure where the PR for the algorithm is

PR #2169, opened 2015-06-01
kNN exact matching classifications

### LINKED ISSUE #2168: kNN classifier should allow for exact matching
Exact matching would be a useful parameter for the [kNN classifier](https://github.com/numenta/nupic/blob/master/nupic/algorithms/KNNClassifier.py); i.e. classifications where the distance is zero.

BoltzmannBrain:
Fixes #2168.
@subutai please review.

### DIFF
--- a/nupic/algorithms/KNNClassifier.py
+++ b/nupic/algorithms/KNNClassifier.py
@@ -575,10 +575,9 @@ class KNNClassifier(object):
       validVectorCount = len(self._categoryList) - self._categoryList.count(-1)
 
       # Loop through the indices of the nearest neighbors.
-      roundToZero = float(1)/len(inputPattern)
       if self.exact:
         # Is there an exact match in the distances?
-        exactMatches = numpy.where(dist<roundToZero)[0]
+        exactMatches = numpy.where(dist<0.00001)[0]
         if len(exactMatches) > 0:
           for i in exactMatches[:min(self.k, validVectorCount)]:
             inferenceResult[self._categoryList[i]] += 1.0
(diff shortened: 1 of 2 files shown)


subutai:
Could you add a test case for this option? The KNN tests are in `$NUPIC/tests/integration/nupic/algorithms/knn_classifier_test`

nupic/algorithms/KNNClassifier.py
```diff
@@ -565,27 +569,35 @@ def infer(self, inputPattern, computeScores=True,
...
+        sorted = dist.argsort()
+        for j in sorted[:min(self.k, validVectorCount)]:
+          inferenceResult[self._categoryList[j]] += 1.0
+
+      # Prepare inference results.
+      if inferenceResult.any():
+        winner = inferenceResult.argmax()
+        inferenceResult /= inferenceResult.sum()
+      else:
+        winner = None
```
subutai: You are changing the return semantics. Please update the comment for infer() with the meaning of None.  Also, add a test case that covers this scenario.

subutai:
:+1: Other than the two comments.

BoltzmannBrain:
@subutai please review the tests. For now I added tests for exact matching within the current [classifier_test.py](https://github.com/numenta/nupic/blob/master/tests/integration/nupic/algorithms/knn_classifier_test/classifier_test.py), but I think this file and [categories_test.py](https://github.com/numenta/nupic/blob/master/tests/integration/nupic/algorithms/knn_classifier_test/categories_test.py) should be redone to follow standard testing framework; as is it's difficult to follow what's going on. If you agree I'll go ahead and create an issue, although fixing it won't be a top priority of mine right now.

BoltzmannBrain:
@subutai please review. To test exact matching, the classifier is trained on `patternDict` patterns and then tested on `testDict` patterns, where corresponding patterns are different for 2% of their entries. That is, for this test where each pattern is 100 values, the test pattern is identical to the training pattern for 98 of those values. Then the expected classifications for the `testDict` are all None.

subutai:
:+1:

PR #2174, opened 2015-06-04
Updated inferBacktrack in TP.py to remove contradiction in example

akhilaananthram:
In TP.py, inferBacktrack, originally the doc string said that seq 1 would be selected.  Later it says that because Z is more steps back, it should be selected.  Z is seq 4.  I just corrected it so both spots say seq 4 should be selected.

@subutai


subutai:
@akhilaananthram This PR seems to have a bunch of other commits in there and doesn't reflect the title. Can you clean it up to include just the intended commit?

PR #2176, opened 2015-06-05
Updated inferBacktrack in TP.py to remove contradiction in example

akhilaananthram:
In TP.py, inferBacktrack, originally the doc string said that seq 1 would be selected. Later it says that because Z is more steps back, it should be selected. Z is seq 4. I just corrected it so both spots say seq 4 should be selected. I updated it to remove commits that weren't relevant.

@subutai


subutai:
:+1: There's no linked issue but it's a one character fix to a comment.

PR #2188, opened 2015-06-08
Fixes #2183  - Port capnp serialization to TPRegion class

### LINKED ISSUE #2183: Port capnp serialization to TMRegion class
See parent issue: https://github.com/numenta/nupic/issues/2158

david-ragazzi:
Fixes: https://github.com/numenta/nupic/issues/2183

### DIFF
--- a/nupic/regions/TPRegion.capnp
+++ b/nupic/regions/TPRegion.capnp
@@ -1,0 +1,25 @@
+@0xf0efb05e82c5561b;
+
+using import "../bindings/proto/TemporalMemoryV1.capnp".TemporalMemoryV1Proto;
+
+struct TemporalMemoryV1RegionProto {
+  temporalImp @0 :Text;
+  temporalInstance @1 :TemporalMemoryV1Proto;
+
+  columnCount @2 :UInt32;
+  inputWidth @3 :UInt32;
+  cellsPerColumn @4 :UInt32;
+  learningMode @5 :Bool;
+  inferenceMode @6 :Bool;
+  anomalyMode @7 :Bool;
+  topDownMode @8 :Bool;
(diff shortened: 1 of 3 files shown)


nupic/research/TP.py
```diff
@@ -3280,21 +3634,75 @@ def __init__(self, tp, isSequenceSeg):
...
-  def __ne__(self, s):
-    return not self == s
+  def __eq__(self, other):
+    membersToIgnore = ['tp']
+    for k, v1 in self.__dict__.iteritems():
+      if k in membersToIgnore:
+        pass
+      else:
+        if not k in other.__dict__:
+          print 'not found: ', k
```
david-ragazzi: I'll remove this `print`'s before merge it.

scottpurdy:
Please separate the region out into another PR

david-ragazzi:
> Please separate the region out into another PR

Done.

scottpurdy:
Why are you including the TemporalMemoryV1Proto? If we are using separate schema since they are sufficiently different then I think it will be easier to maintain to have completely separate schema files. That way we don't have to coordinate changes between two different algorithms.

rhyolight:
Marking as `in progress` until builds pass.

scottpurdy:
@david-ragazzi - Getting the C++ -> Python capnp stuff working is taking a while so I will try to review your region changes now.

The TPRegion supports multiple implementations currently. The cpp/py options will need TemporalMemoryV1Proto while the tm_py option is the new implementation and will have a different schema. Should the capnp schema for the region have a [union](https://capnproto.org/language.html#unions) or [AnyPointer](https://capnproto.org/language.html#dynamically-typed-fields) field to support the two different cases?

PR #2197, opened 2015-06-09
Fixes #2196 - Implement global assert to test original and deserialized class instances

### LINKED ISSUE #2196: Implement global assert to test original and deserialized class instances
A global assert to test original and deserialized class instances can simplify the serialization unit tests.

david-ragazzi:
Fixes https://github.com/numenta/nupic/issues/2196

### DIFF
--- a/nupic/test/test_framework_helpers.py
+++ b/nupic/test/test_framework_helpers.py
@@ -33,6 +33,8 @@ by multiple mainstream test frameworks, such as pytest and nose.
 
 
 # Our current test framework is pytest
+import numbers
+import numpy
 import pytest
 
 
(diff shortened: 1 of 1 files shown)


david-ragazzi:
@chetan51 @oxtopus @scottpurdy Please review and merge this ASAP. This is very important to serialization PRs.

scottpurdy:
@david-ragazzi - Did you consider setting `__eq__` on the classes so that you could just do `obj1 == obj2` or `self.assertEqual(obj1, obj2)`?

david-ragazzi:
> @david-ragazzi - Did you consider setting __eq__ on the classes so that you could just do obj1 == obj2 or self.assertEqual(obj1, obj2)?

Do you mean setting __eq__ which will call `assertInstancesAlmostEqual` (or other name like `checkInstancesAlmostEqual`) within it, right?

scottpurdy:
I'm hesitant to use something like this. I've tried similar things before and it seems like a hack. The only reason I am not strongly in favor of having each class explicitly check attributes is that it is difficult to maintain. @oxtopus how do you feel about something like this vs. explicitly checking equality on a per-class basis?

oxtopus:
@scottpurdy I agree.  I'd much rather see equality comparisons implemented in the classes themselves, as is typically convention in OOP and, while useful in some very specific use-cases, this change would add some maintenance overhead.

david-ragazzi:
Could you guys pick some example where I can see explicitly checking equality?

oxtopus:
https://docs.python.org/2/reference/datamodel.html#object.__eq__

scottpurdy:
@oxtopus - I think David suggested that he could still implement via `__eq__` but would use this function inside that method for the actual comparison. So the question for me is whether you explicitly check each attribute or you use a function like this to check them all.

david-ragazzi:
>  I've tried similar things before and it seems like a hack. 

I think you're right. In the end, probably we could have a bizarre function which a few people understand it.

PR #2206, opened 2015-06-11
Modified TPRegion to output the indices needed for the Union Pooler

### LINKED ISSUE #2081: TPRegion should output 2 new fields: active & predicted columns separately
Depends on https://github.com/numenta/nupic/issues/1450

It will be useful if the output of the TP region provided separate fields for the `active` & `predicted` columns; 

currently these are combined to the default output as defined by the TP outputType - either of them, or both. 

Use case: AnomalyRegion could depend only on TPRegion (and not SP, which is now required to get the active columns, [see](https://github.com/numenta/nupic/blob/master/examples/network/network_api_demo.py#L158) ), I wonder if the [lrnActiveCells](https://github.com/numenta/nupic/blob/master/nupic/regions/TPRegion.py#L576) is the same as activeOut suggested for this usecase (maybe not, as it's only used in `learn` phase)

TODO:
- [ ] TPRegion.output provides new activeOut and predictedOut fields
- [ ] also we should remove [anomalyScore](https://github.com/numenta/nupic/blob/master/nupic/regions/TPRegion.py#L568) from TPRegion, SPRegion output, and use AnomalyRegion only

akhilaananthram:
The Union Pooler needs to know the active indices from the output of the temporal memory and which ones were predicted and active.  Originally, TPRegion only outputted the active bits SDR.  Now, it outputs the indices of the active bits and the indices of the active bits that were predicted.

For numenta/nupic.research#69
Fixes #2081

@ryan @scottpurdy

### DIFF
--- a/nupic/regions/TPRegion.py
+++ b/nupic/regions/TPRegion.py
@@ -517,7 +517,8 @@ class TPRegion(PyRegion):
       size = activeLearnCells.shape[0] * activeLearnCells.shape[1]
       outputs['lrnActiveStateT'][:] = activeLearnCells.reshape(size)
 
-    if self.unionMode:
+    unionMode = getattr(self, "unionMode", False)
+    if unionMode:
       # Reshape so we are dealing with 1D arrays
       activeState = self._tfdr.getActiveState().reshape(-1).astype('float32')
       predictedState = self._tfdr.getPredictedState().reshape(-1).astype('float32')
--- a/nupic/regions/TPRegion.py
+++ b/nupic/regions/TPRegion.py
@@ -211,7 +211,7 @@ def _getAdditionalSpecs(temporalImp, kwargs={}):
       constraints='bool'),
 
     computePredictedActiveCellIndices=dict(
-      description='1 if an union input is computed',
+      description='1 if indices are needed',
       accessMode='Create',
       dataType='UInt32',
       count=1,
--- a/nupic/regions/TPRegion.py
+++ b/nupic/regions/TPRegion.py
@@ -518,13 +518,14 @@ class TPRegion(PyRegion):
       outputs['lrnActiveStateT'][:] = activeLearnCells.reshape(size)
 
     if self.unionMode:
-      activeState = self._tfdr.getActiveState()
-      predictedState = self._tfdr.getPredictedState()
+      # Reshape so we are dealing with 1D arrays
+      activeState = self._tfdr.getActiveState().reshape(-1).astype('float32')
+      predictedState = self._tfdr.getPredictedState().reshape(-1).astype('float32')
       activeIndices = set(numpy.where(activeState != 0)[0])
       predictedIndices= set(numpy.where(predictedState != 0)[0])
       predictedActiveIndices = activeIndices & predictedIndices
-      outputs["activeCellsIndices"] = numpy.array(list(activeIndices), dtype=bool)
-      outputs["predictedActiveCellsIndices"] = numpy.array(list(predictedActiveIndices), dtype=bool)
+      outputs["activeCellsIndices"] = numpy.array(list(activeIndices), dtype=int)
+      outputs["predictedActiveCellsIndices"] = numpy.array(list(predictedActiveIndices), dtype=int)
 
 
   #############################################################################
--- a/nupic/regions/TPRegion.py
+++ b/nupic/regions/TPRegion.py
@@ -210,7 +210,7 @@ def _getAdditionalSpecs(temporalImp, kwargs={}):
       count=1,
       constraints='bool'),
 
-    unionMode=dict(
+    computePredictedActiveCellIndices=dict(
       description='1 if an union input is computed',
       accessMode='Create',
       dataType='UInt32',
@@ -305,7 +305,7 @@ class TPRegion(PyRegion):
                cellsSavePath='',
                temporalImp=gDefaultTemporalImp,
                anomalyMode=False,
-               unionMode=False,
+               computePredictedActiveCellIndices=False,
 
                **kwargs):
 
@@ -322,7 +322,7 @@ class TPRegion(PyRegion):
     self.learningMode   = True      # Start out with learning enabled
     self.inferenceMode  = False
     self.anomalyMode    = anomalyMode
-    self.unionMode = unionMode
+    self.computePredictedActiveCellIndices = computePredictedActiveCellIndices
     self.topDownMode    = False
     self.columnCount    = columnCount
     self.inputWidth     = inputWidth
@@ -517,8 +517,8 @@ class TPRegion(PyRegion):
       size = activeLearnCells.shape[0] * activeLearnCells.shape[1]
       outputs['lrnActiveStateT'][:] = activeLearnCells.reshape(size)
 
-    unionMode = getattr(self, "unionMode", False)
-    if unionMode:
+    computePredictedActiveCellIndices = getattr(self, "computePredictedActiveCellIndices", False)
+    if computePredictedActiveCellIndices:
       # Reshape so we are dealing with 1D arrays
       activeState = self._tfdr.getActiveState().reshape(-1).astype('float32')
       predictedState = self._tfdr.getPredictedState().reshape(-1).astype('float32')
--- a/nupic/regions/TPRegion.py
+++ b/nupic/regions/TPRegion.py
@@ -526,6 +526,7 @@ class TPRegion(PyRegion):
       predictedActiveIndices = numpy.intersect1d(activeIndices, predictedIndices)
       outputs["activeCells"].fill(0)
       outputs["activeCells"][activeIndices] = 1
+      outputs["predictedActiveCells"].fill(0)
       outputs["predictedActiveCells"][predictedActiveIndices] = 1
 
 
--- a/nupic/regions/TPRegion.py
+++ b/nupic/regions/TPRegion.py
@@ -517,8 +517,7 @@ class TPRegion(PyRegion):
       size = activeLearnCells.shape[0] * activeLearnCells.shape[1]
       outputs['lrnActiveStateT'][:] = activeLearnCells.reshape(size)
 
-    computePredictedActiveCellIndices = getattr(self, "computePredictedActiveCellIndices", False)
-    if computePredictedActiveCellIndices:
+    if self.computePredictedActiveCellIndices:
       # Reshape so we are dealing with 1D arrays
       activeState = self._tfdr.getActiveState().reshape(-1).astype('float32')
       predictedState = self._tfdr.getPredictedState().reshape(-1).astype('float32')
@@ -776,6 +775,9 @@ class TPRegion(PyRegion):
 
     if not hasattr(self, 'storeDenseOutput'):
       self.storeDenseOutput = False
+
+    if not hasattr(self, 'computePredictedActiveCellIndices'):
+      self.computePredictedActiveCellIndices = False
 
     self.__dict__.update(state)
     self._loaded = True
--- a/nupic/regions/TPRegion.py
+++ b/nupic/regions/TPRegion.py
@@ -211,7 +211,7 @@ def _getAdditionalSpecs(temporalImp, kwargs={}):
       constraints='bool'),
 
     computePredictedActiveCellIndices=dict(
-      description='1 if indices are needed',
+      description='1 if active and predicted active indices should be computed',
       accessMode='Create',
       dataType='UInt32',
       count=1,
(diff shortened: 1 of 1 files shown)


nupic/regions/TPRegion.py
```diff
@@ -209,6 +210,14 @@ def getConstraints(arg):
       count=1,
       constraints='bool'),
 
+    unionMode=dict(
```
subutai: I wouldn't call this `unionMode` since the requirement is not specific to union pooling. I would call it something more descriptive and accurate like `computePredictedActiveCellIndices`.

nupic/regions/TPRegion.py
```diff
@@ -506,6 +517,17 @@ def _compute(self, inputs, outputs):
       size = activeLearnCells.shape[0] * activeLearnCells.shape[1]
       outputs['lrnActiveStateT'][:] = activeLearnCells.reshape(size)
 
+    computePredictedActiveCellIndices = getattr(self, "computePredictedActiveCellIndices", False)
```
scottpurdy: Is this for backwards compatibility? It might be better to make sure this attribute gets set in the `__setstate__` method and then elsewhere in the code you can assume it is set.

scottpurdy:
This looks good to me. Since this is going into nupic (rather than nupic.research), there needs to be proper rationale. I think these changes are generally useful for other regions to have access to.

It might be good to break this into two separate PRs, one for active cells, and one for the predicted-active cells.

nupic/regions/TPRegion.py
```diff
@@ -209,6 +210,14 @@ def getConstraints(arg):
       count=1,
       constraints='bool'),
 
+    computePredictedActiveCellIndices=dict(
+      description='1 if an union input is computed',
```
scottpurdy: This parameter isn't specific to union pooling

nupic/regions/TPRegion.py
```diff
@@ -506,6 +517,16 @@ def _compute(self, inputs, outputs):
...
       outputs['lrnActiveStateT'][:] = activeLearnCells.reshape(size)
 
+    if self.computePredictedActiveCellIndices:
+      # Reshape so we are dealing with 1D arrays
+      activeState = self._tfdr.getActiveState().reshape(-1).astype('float32')
+      predictedState = self._tfdr.getPredictedState().reshape(-1).astype('float32')
+      activeIndices = set(numpy.where(activeState != 0)[0])
+      predictedIndices= set(numpy.where(predictedState != 0)[0])
+      predictedActiveIndices = activeIndices & predictedIndices
+      outputs["activeCellsIndices"] = numpy.array(list(activeIndices), dtype=int)
```
scottpurdy: I think this should be something like this:

```
outputs["activeCellsIndices"][:] = activeIndices
```

akhilaananthram: If I do that then activeIndices would need to be the same length as outputs["activeCellsIndices"][:].  The length of outputs["activeCellsIndices"][:] is set as the maximum possible length. The length of activeIndices will vary.

breznak:
@akhilaananthram is this same as https://github.com/numenta/nupic/issues/2081 ?

akhilaananthram:
@breznak: yup! Thanks for catching that.

breznak:
yea, the main deal was to have separate active/predicted. so :+1: :+1:

nupic/regions/TPRegion.py
scottpurdy: You are overriding the array in `output["activeCellsIndices"]` with a copy of `activeIndices`, right? I think overriding it will cause issues, you should just be modifying the existing array.

scottpurdy: Maybe you want to do this?

```
outputs["activeCellsIndices"].fill(0)
outputs["activeCellsIndices"][activeIndices] = 1
```

akhilaananthram:
@scottpurdy: changed how i handled the output so i actually modify the existing array

nupic/regions/TPRegion.py
```diff
@@ -506,6 +517,17 @@ def _compute(self, inputs, outputs):
       size = activeLearnCells.shape[0] * activeLearnCells.shape[1]
       outputs['lrnActiveStateT'][:] = activeLearnCells.reshape(size)
 
+    if self.computePredictedActiveCellIndices:
+      # Reshape so we are dealing with 1D arrays
+      activeState = self._tfdr.getActiveState().reshape(-1).astype('float32')
+      predictedState = self._tfdr.getPredictedState().reshape(-1).astype('float32')
+      activeIndices = set(numpy.where(activeState != 0)[0])
+      predictedIndices= set(numpy.where(predictedState != 0)[0])
+      predictedActiveIndices = activeIndices & predictedIndices
```
scottpurdy: Instead of converting to sets and using the `&` operator, use `numpy.intersect1d`.

Plus then `activeIndices` and `predictedIndices` are numpy arrays which can be used as indices below without needed to convert to a Python list.

nupic/regions/TPRegion.py
```diff
@@ -209,6 +210,14 @@ def getConstraints(arg):
       count=1,
       constraints='bool'),
 
+    computePredictedActiveCellIndices=dict(
+      description='1 if indices are needed',
```
scottpurdy: Have a better description

scottpurdy:
Does it make sense to have both new outputs controlled by the same parameter? Is there a case where you would want one but not the other? And are there other cases where outputs are conditional or is it more common for the outputs to be set regardless of whether they are used? Is there a way to see during initialization of the region whether or not any links use the output rather than passing a constructor argument?

akhilaananthram:
I'm not sure if it makes sense to pair active and predicted+active always.  The only use case I know of requires both of them.  That is why I made them both dependent on computePredictedActiveCellIndices. There are cases when the outputs are conditional.  Most of TPRegion's outputs are conditional.  That is why I made this one conditional as well.  Right now, I don't think there is a way to see if any links use the output.

scottpurdy:
It looks like you can [get an output by name](https://github.com/numenta/nupic.core/blob/54faae374b409b8874feeeec40b2644eec6cddc1/src/nupic/engine/RegionImpl.hpp#L212) and then [check whether the output has outgoing links](https://github.com/numenta/nupic.core/blob/54faae374b409b8874feeeec40b2644eec6cddc1/src/nupic/engine/Output.hpp#L130). Can you try that in the initialize function rather than adding a parameter?

Or do you think the parameter is better?

akhilaananthram:
I think those methods are only accessible to the wrapper for a network region in nupic/engine/__init__.py, not to the actual region

scottpurdy:
`getOutput` is a protected method of RegionImpl so it is only accessible to RegionImpl subclasses (which is where we want to call it). And then `hasOutgoingLinks` on the Output instance is public.

nupic/regions/TPRegion.py
```diff
@@ -506,6 +517,17 @@ def _compute(self, inputs, outputs):
...
+    if self.computePredictedActiveCellIndices:
+      # Reshape so we are dealing with 1D arrays
+      activeState = self._tfdr.getActiveState().reshape(-1).astype('float32')
+      predictedState = self._tfdr.getPredictedState().reshape(-1).astype('float32')
+      activeIndices = numpy.where(activeState != 0)[0]
+      predictedIndices= numpy.where(predictedState != 0)[0]
+      predictedActiveIndices = numpy.intersect1d(activeIndices, predictedIndices)
+      outputs["activeCells"].fill(0)
+      outputs["activeCells"][activeIndices] = 1
+      outputs["predictedActiveCells"][predictedActiveIndices] = 1
```
scottpurdy: do you need to zero out `outputs["predictedActiveCells"]` first?

scottpurdy:
:+1: after the one additional comment

nupic/regions/TPRegion.py
akhilaananthram: oops. yeah

nupic/regions/TPRegion.py
```diff
@@ -506,6 +517,18 @@ def _compute(self, inputs, outputs):
       size = activeLearnCells.shape[0] * activeLearnCells.shape[1]
       outputs['lrnActiveStateT'][:] = activeLearnCells.reshape(size)
```
breznak: @akhilaananthram is this the same field semantically? If so, we should remove either of them: `lrnActiveStateT vs. activeCells`. Or what is the difference?
https://github.com/numenta/nupic/pull/2206/files#diff-7cf0d542d5dacba939c229573fc267e5R528

nupic/regions/TPRegion.py
```diff
@@ -506,6 +517,18 @@ def _compute(self, inputs, outputs):
...
+      # Reshape so we are dealing with 1D arrays
+      activeState = self._tfdr.getActiveState().reshape(-1).astype('float32')
+      predictedState = self._tfdr.getPredictedState().reshape(-1).astype('float32')
+      activeIndices = numpy.where(activeState != 0)[0]
+      predictedIndices= numpy.where(predictedState != 0)[0]
+      predictedActiveIndices = numpy.intersect1d(activeIndices, predictedIndices)
+      outputs["activeCells"].fill(0)
+      outputs["activeCells"][activeIndices] = 1
+      outputs["predictedActiveCells"].fill(0)
+      outputs["predictedActiveCells"][predictedActiveIndices] = 1
```
breznak: I can compute that from these fields too, but my original intention was to provide 2 separate fields (`activeCells`, `predictedCells`), wouldn't that make more sense?

nupic/regions/TPRegion.py
akhilaananthram: You are right. They are the same.

nupic/regions/TPRegion.py
akhilaananthram: So replace "predictedActiveCells" with "predictedCells"?

cogmission: Should be "successfullyPredictedCells" to make it absolutely distinct from:

1. Predicted Cells - Cells currently predicted in the current input cycle.
2. Previously Predicted Cells - Cells predicted in the previous input cycle
(t -1)
3. Successfully Predicted Cells - Cells predicted in the previous input
cycle that **also** appear in the active input field. (as distinct from
previously predicted cells which are not successful (a.k.a. not present as
activations in current input field)

On Mon, Jun 29, 2015 at 11:18 AM, Akhila <notifications@github.com> wrote:

> In nupic/regions/TPRegion.py
> <https://github.com/numenta/nupic/pull/2206#discussion_r33481503>:
>
> > @@ -506,6 +517,18 @@ def _compute(self, inputs, outputs):
> >        size = activeLearnCells.shape[0] * activeLearnCells.shape[1]
> >        outputs['lrnActiveStateT'][:] = activeLearnCells.reshape(size)
> >
> > +    if self.computePredictedActiveCellIndices:
> > +      # Reshape so we are dealing with 1D arrays
> > +      activeState = self._tfdr.getActiveState().reshape(-1).astype('float32')
> > +      predictedState = self._tfdr.getPredictedState().reshape(-1).astype('float32')
> > +      activeIndices = numpy.where(activeState != 0)[0]
> > +      predictedIndices= numpy.where(predictedState != 0)[0]
> > +      predictedActiveIndices = numpy.intersect1d(activeIndices, predictedIndices)
> > +      outputs["activeCells"].fill(0)
> > +      outputs["activeCells"][activeIndices] = 1
> > +      outputs["predictedActiveCells"].fill(0)
> > +      outputs["predictedActiveCells"][predictedActiveIndices] = 1
>
> So replace "predictedActiveCells" with "predictedCells"?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/numenta/nupic/pull/2206/files#r33481503>.
>



-- 
*With kind regards,*

David Ray
Java Solutions Architect

*Cortical.io <http://cortical.io/>*
Sponsor of:  HTM.java <https://github.com/numenta/htm.java>

d.ray@cortical.io
http://cortical.io

akhilaananthram: do we need to change the name from predictedActiveCells to successfullyPredictedCells?

cogmission: I think "predictedActiveCells" is ambiguous at best, and totally obscuring
at worst - and is no where near as clear as successfullyPredictedCells. But
that's ultimately up to @ScottPurdy... I did however make this change in
HTM.java ;-)

On Mon, Jun 29, 2015 at 11:57 AM, Akhila <notifications@github.com> wrote:

> In nupic/regions/TPRegion.py
> <https://github.com/numenta/nupic/pull/2206#discussion_r33485701>:
>
> > @@ -506,6 +517,18 @@ def _compute(self, inputs, outputs):
> >        size = activeLearnCells.shape[0] * activeLearnCells.shape[1]
> >        outputs['lrnActiveStateT'][:] = activeLearnCells.reshape(size)
> >
> > +    if self.computePredictedActiveCellIndices:
> > +      # Reshape so we are dealing with 1D arrays
> > +      activeState = self._tfdr.getActiveState().reshape(-1).astype('float32')
> > +      predictedState = self._tfdr.getPredictedState().reshape(-1).astype('float32')
> > +      activeIndices = numpy.where(activeState != 0)[0]
> > +      predictedIndices= numpy.where(predictedState != 0)[0]
> > +      predictedActiveIndices = numpy.intersect1d(activeIndices, predictedIndices)
> > +      outputs["activeCells"].fill(0)
> > +      outputs["activeCells"][activeIndices] = 1
> > +      outputs["predictedActiveCells"].fill(0)
> > +      outputs["predictedActiveCells"][predictedActiveIndices] = 1
>
> do we need to change the name from predictedActiveCells to
> successfullyPredictedCells?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/numenta/nupic/pull/2206/files#r33485701>.
>



-- 
*With kind regards,*

David Ray
Java Solutions Architect

*Cortical.io <http://cortical.io/>*
Sponsor of:  HTM.java <https://github.com/numenta/htm.java>

d.ray@cortical.io
http://cortical.io

rhyolight: It is usually best to choose verbose clarity over terse ambiguity.

---------
Matt Taylor
OS Community Flag-Bearer
Numenta

PR #2219, opened 2015-06-15
Allows users to register Python regions by passing a pointer to the class

### LINKED ISSUE #2211: Allow users to use custom regions where the class name does not match the module name.
This means users should be able to register a region by passing a reference to its class.  Then, when they want to add the region, they simply say the class name.

akhilaananthram:
By passing a pointer to the class, the user can create a region who's name doesn't have to match the file name.

There are also changes to nupic.core for this: https://github.com/numenta/nupic.core/pull/466

Fixes #2211

@scottpurdy


extensions/cpp_region/PyRegion.cpp
```diff
@@ -860,11 +874,14 @@ void PyRegion::compute()
...
 // by RegionImplFactory.
 //
-void PyRegion::createSpec(const char * nodeType, Spec & ns)
+void PyRegion::createSpec(const char * nodeType, Spec & ns, const char* className)
 {
   // Get the Python class object
-  std::string className = Path::getExtension(nodeType);
-  py::Class nodeClass(nodeType, className);
+  std::string realClassName(className);
+  if (realClassName.empty())
```
scottpurdy: use braces even though there is only one statement

extensions/cpp_region/PyRegion.cpp
```diff
@@ -93,7 +93,7 @@ extern "C"
   // createPyNode is used by the MultinodeFactory to create a C++ PyNode instance
   // That references a Python instance. The function tries to create a NuPIC 2.0
   // Py node first and if it fails it tries to create a NuPIC 1.x Py node
-  NTA_EXPORT void * NTA_createPyNode(const char * module, void * nodeParams, void * region, void ** exception)
+  NTA_EXPORT void * NTA_createPyNode(const char * module, void * nodeParams, void * region, void ** exception, const char* className)
```
scottpurdy: need a line break

examples/network/custom_region_demo.py
```diff
@@ -90,8 +90,11 @@ def createNetwork(dataSource):
   # In this demo, we have modified it using sys.path.append since we need it to
   # have an effect on this program.
   sys.path.append(os.path.dirname(os.path.abspath(__file__)))
-  # Add custom region package to network
-  Network.registerRegionPackage("custom_region")
+  
+  custom_region = __import__("custom_region", fromlist=["IdentityRegion"])
```
scottpurdy: How about we change the filename to be different than the class name to make sure this works? And use a regular import statement.

```
# In import section
from custom_region import identity_region

Network.registionRegion(identity_region.IdentityRegion)
```

nupic/engine/__init__.py
```diff
@@ -730,12 +730,12 @@ def inspect(self):
...
-  def registerRegionPackage(package):
+  def registerRegion(regionClass):
     """
-    Adds the package to the list of packages the network can access for regions
-    package: name of Python package, that is reachable through the PYTHONPATH
+    Adds the module and class name for the region to the list of classes the network can use
+    regionClass: a pointer to a subclass of PyRegion
     """
-    engine.Network.registerPyRegionPackage(package)
+    engine.Network.registerPyRegion(regionClass.__module__, regionClass.__name__)
```
scottpurdy: I believe that `__name__` is the full import path, not just the class name.

scottpurdy: Oohhh nvm, that is for the module, not the class

scottpurdy:
:+1:

akhilaananthram:
@scottpurdy responded to review

examples/network/custom_region_demo.py
akhilaananthram: I didn't want to include it at the top incase a user tries to run the example from a different directory since custom_region is only added to the python path in the demo.  I replaced it with a regular import though.

scottpurdy:
:+1:

extensions/cpp_region/PyRegion.cpp
```diff
@@ -282,14 +300,19 @@ static void prepareCreationParams(const ValueMap & vm, py::Dict & d)
   }
 };
 
-PyRegion::PyRegion(const char * module, const ValueMap & nodeParams, Region * region) : 
+PyRegion::PyRegion(const char * module, const ValueMap & nodeParams, Region * region, const char* className) :
```
scottpurdy: Break this line to keep under 80 characters

nupic/engine/__init__.py
```diff
@@ -29,6 +29,22 @@
 
 basicTypes = ['Byte', 'Int16', 'UInt16', 'Int32', 'UInt32', 'Int64', 'UInt64', 'Real32', 'Real64', 'Handle']
 
+pyRegions = {"nupic.regions.AnomalyRegion": "AnomalyRegion",
+           "nupic.regions.CLAClassifierRegion": "CLAClassifierRegion",
```
scottpurdy: indentation

scottpurdy: and this doesn't really need to be a dictionary since you just want to iterate over the elements. You could make it a tuple of tuples which would be immutable:

```
pyRegions = (
    ("nupic.regions.AnomalyRegion", "AnomalyRegion"),
    ...
)
```

and later:

```
for module, className in pyRegions:
```

scottpurdy:
:+1: but tests failing

PR #2253, opened 2015-06-25
Updates nupic.core and adds function definitions for read/write in PyRegion

### LINKED ISSUE #2252: Implement read/write capnp methods in PyRegion
See parent #2241 

This is the initial step to get PyRegion in sync with RegionImpl. The implementation of passing the proto struct to Python will have to wait for #2242

scottpurdy:
fixes #2252

@chetan51 please review

### DIFF
--- a/.nupic_modules
+++ b/.nupic_modules
@@ -1,3 +1,3 @@
 # Default nupic.core dependencies (override in optional .nupic_config)
 NUPIC_CORE_REMOTE = 'git://github.com/numenta/nupic.core.git'
-NUPIC_CORE_COMMITISH = 'f55119a2f255fa4961b9c44ecb5261076bb4db0a'
+NUPIC_CORE_COMMITISH = 'cb55ebd9633923040176db0c041945f190d69800'
--- a/extensions/cpp_region/PyRegion.cpp
+++ b/extensions/cpp_region/PyRegion.cpp
@@ -25,6 +25,9 @@
 #include <iostream>
 #include <sstream>
 #include <memory>
+
+#include <capnp/any.h>
+
 #include <nupic/engine/Spec.hpp>
 #include <nupic/engine/Region.hpp>
 #include <nupic/engine/Input.hpp>
@@ -425,8 +428,15 @@ void PyRegion::deserialize(BundleIO& bundle)
   
 }
 
+void PyRegion::write(capnp::AnyPointer::Builder& proto) const
+{
+  NTA_THROW << "Unimplemented method PyRegion::write.";
+}
 
-
+void PyRegion::read(capnp::AnyPointer::Reader& proto)
+{
+  NTA_THROW << "Unimplemented method PyRegion::write.";
+}
 
 const Spec & PyRegion::getSpec()
 {
--- a/extensions/cpp_region/PyRegion.cpp
+++ b/extensions/cpp_region/PyRegion.cpp
@@ -435,7 +435,7 @@ void PyRegion::write(capnp::AnyPointer::Builder& proto) const
 
 void PyRegion::read(capnp::AnyPointer::Reader& proto)
 {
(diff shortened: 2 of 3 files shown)


scottpurdy:
@vitaly-krugl - please review. This change is necessary to update the nupic.core SHA in nupic

vitaly-krugl:
Reviewing...

extensions/cpp_region/PyRegion.hpp
```diff
@@ -55,6 +57,9 @@ namespace nupic
     void serialize(BundleIO& bundle);
     void deserialize(BundleIO& bundle);
 
+    virtual void write(capnp::AnyPointer::Builder& proto) const override;
```
vitaly-krugl: Some comments tying these methods to serialization/deserialization and describing the intended strategy would really help. I see serialize/deserialize as well as read/write here, and am confused about their relationship. Which ones would I use? Which ones would I need to override? Does one set of them call the other set?

scottpurdy: Good idea. I don't want to get into too much detail since there are a million places where this exists. Ultimately uses should understand the difference at the high level since I don't think we can have a full explanation at every single entry point to serialization.

extensions/cpp_region/PyRegion.hpp
```diff
@@ -52,9 +54,14 @@ namespace nupic
     PyRegion(const char * module, BundleIO& bundle, Region * region, const char* className="");
     virtual ~PyRegion();
 
+    // Manual serialization methods
     void serialize(BundleIO& bundle);
     void deserialize(BundleIO& bundle);
 
+    // Capnp serialization methods - not yet implemented for PyRegions
```
vitaly-krugl: When reading this code, I still wouldn't know my way through this. When/why might I need/want to use the "Manual serialization methods" vs. the "Capnp serialization methods"? Is one set intended to obsolete the other? Where would I learn any of this?

extensions/cpp_region/PyRegion.cpp
```diff
@@ -425,8 +428,15 @@ void PyRegion::deserialize(BundleIO& bundle)
   
 }
 
+void PyRegion::write(capnp::AnyPointer::Builder& proto) const
```
vitaly-krugl: Should this also have `override` at the end?

extensions/cpp_region/PyRegion.cpp
```diff
@@ -425,8 +428,15 @@ void PyRegion::deserialize(BundleIO& bundle)
...
+{
+  NTA_THROW << "Unimplemented method PyRegion::write.";
+}
 
-
+void PyRegion::read(capnp::AnyPointer::Reader& proto)
```
vitaly-krugl: Should this also have `override` at the end?

.nupic_modules
```diff
@@ -1,3 +1,3 @@
 # Default nupic.core dependencies (override in optional .nupic_config)
 NUPIC_CORE_REMOTE = 'git://github.com/numenta/nupic.core.git'
-NUPIC_CORE_COMMITISH = 'f55119a2f255fa4961b9c44ecb5261076bb4db0a'
\ No newline at end of file
+NUPIC_CORE_COMMITISH = 'cb55ebd9633923040176db0c041945f190d69800'
```
vitaly-krugl: cb55ebd9633923040176db0c041945f190d69800  exists in nupic.core, so I am assuming that it's the one that you intended.

vitaly-krugl:
@scottpurdy: I'm done with code review.

extensions/cpp_region/PyRegion.cpp
scottpurdy: No, the spec states that `override` should be used exactly once and I have to use it in the header.

vitaly-krugl:
LGTM

PR #2263, opened 2015-06-28
Clamodel predicted field check

### LINKED ISSUE #2208: Better exception needed for improperly set predictedField
When ```model.enableInference({'predictedField': fieldName})``` is improperly set there should be a better/more descriptive exception raised than the one below:

``` 
File "/Users/jweiss/Library/Python/2.7/lib/python/site-packages/nupic-0.3.0.dev0-py2.7-macosx-10.10-intel.egg/nupic/frameworks/opf/clamodel.py", line 396, in run
  File "/Users/jweiss/Library/Python/2.7/lib/python/site-packages/nupic-0.3.0.dev0-py2.7-macosx-10.10-intel.egg/nupic/frameworks/opf/clamodel.py", line 534, in _multiStepCompute
    rawInput=rawInput)
  File "/Users/jweiss/Library/Python/2.7/lib/python/site-packages/nupic-0.3.0.dev0-py2.7-macosx-10.10-intel.egg/nupic/frameworks/opf/clamodel.py", line 720, in _handleCLAClassifierMultiStep
    self._classifierInputEncoder = encoderList[self._predictedFieldIdx]
TypeError: list indices must be integers, not NoneType
```

breznak:
more robust check for predicted field in CLAModel, adds unittest
fixes #2208


breznak:
@jaredweiss @rhyolight please review. This has already been partially fixed for anomaly model, now should be solid for all models.

nupic/frameworks/opf/clamodel.py
```diff
@@ -1486,4 +1482,79 @@ def getNextRecordDict(self):
 
 
 
-
+########################
+# example model confuguration: used in tests, examples, etc
+# use: from nupic.frameworks.opf.clamodel import modelConfig
+modelConfig = (
```
rhyolight: I really don't like having this config in the production class. I liked it much better in the test. Please move it back?

rhyolight:
@jaredweiss Can you review? I had one comment, but otherwise, LGTM.

scottpurdy:
This changes the semantics. Before, the check would always be done. With this change, it will only be done for multistep models. Was this intended?

nupic/frameworks/opf/clamodel.py
```diff
@@ -671,6 +666,12 @@ def _handleCLAClassifierMultiStep(self, patternNZ,
         "No predicted field was enabled! Did you call enableInference()?"
       )
     self._predictedFieldName = predictedFieldName
+    # check if predicted field was correctly specified
+    if not self._predictedFieldName in self._input:
```
scottpurdy: should this be `rawInput` instead of `self._input`? Also, please captilalize first letter of comment and use proper grammar. Something like this:

`# Check if the input includes the predicted field.`

breznak: > should this be rawInput instead of self._input? 

they have the same semantics, I just find `self._input` better as it's used in more places in the model.

breznak:
> This changes the semantics. Before, the check would always be done. With this change, it will only be done for multistep models. Was this intended?

@scottpurdy good catch, thanks! Definitely not intended. Before it was run for `Multistep` models, now I've changed to run the check in `run()` for all models and improved relevant set of `predictedFieldName`.

nupic/frameworks/opf/clamodel.py
scottpurdy: If an instance attribute is passed as a function parameter then you should always use the passed in parameter because it is much easier to test the function and would be confusing if you pass a parameter in but it doesn't get used.

nupic/frameworks/opf/clamodel.py
```diff
@@ -368,7 +368,13 @@ def run(self, inputRecord):
     results.inferences = {}
     self._input = inputRecord
 
-    # -------------------------------------------------------------------------
+    # Check if the input includes the predicted field.
+    if not self._predictedFieldName in self._input:
```
scottpurdy: `if X not in Y:`

nupic/frameworks/opf/clamodel.py
```diff
@@ -664,13 +663,10 @@ def _handleCLAClassifierMultiStep(self, patternNZ,
                   None.
     rawInput:   The raw input to the sensor, as a dict.
     """
-    inferenceArgs = self.getInferenceArgs()
-    predictedFieldName = inferenceArgs.get('predictedField', None)
-    if predictedFieldName is None:
+    if self._predictedFieldName is None:
```
scottpurdy: You are removing the code that sets `self._predictedFieldName` so where does it get set now?

nupic/frameworks/opf/model.py
```diff
@@ -160,6 +160,9 @@ def enableInference(self, inferenceArgs=None):
     """
     self.__inferenceEnabled = True
     self.__inferenceArgs = inferenceArgs
+    # Update predicted field
```
scottpurdy: This comment doesn't seem to add anything

nupic/frameworks/opf/clamodel.py
scottpurdy: Nevermind, I see in the base class now.

tests/unit/nupic/frameworks/opf/clamodel_test.py
```diff
@@ -82,7 +83,104 @@ def testTemporalAnomalyModelFactory(self):
...
+       '_timestampRecordIdx': None,
+       u'c0': datetime.datetime(2013, 12, 7, 0, 0),
+       u'c1': 7.0}
+    ]
+
+    model = ModelFactory.create(modelConfig=modelConfig)
+    model.enableLearning()
+    model.enableInference(inferenceArgs)
+
+    for row in data:
```
scottpurdy: Do the second and third data rows add any test coverage or does all of the testing value come from the first one?

tests/unit/nupic/frameworks/opf/clamodel_test.py
```diff
@@ -82,7 +83,104 @@ def testTemporalAnomalyModelFactory(self):
...
+
+    # unknown predicted field - TemporalAnomaly model
+    inferenceArgsWrong = copy.copy(inferenceArgs)
+    inferenceArgsWrong['predictedField']='mispeltField' #make a typo
+
+    model = ModelFactory.create(modelConfig=modelConfig)
+    model.enableLearning()
+    model.enableInference(inferenceArgsWrong)
+
+    expMsg="Expected predicted field 'mispeltField' in input row, but was not found! Raw input is: {'_timestamp': datetime.datetime(2013, 12, 5, 0, 0), '_category': None, '_sequenceId': 0, u'c1': 5.0, u'c0': datetime.datetime(2013, 12, 5, 0, 0), '_timestampRecordIdx': None, '_reset': 0}"
```
scottpurdy: Need to break this onto multiple lines:

```
exceptionMessage = (
    "first line of text "
    "second line of text "
    "raw input is: %r"
) % (data)
```

tests/unit/nupic/frameworks/opf/clamodel_test.py
```diff
@@ -82,7 +83,104 @@ def testTemporalAnomalyModelFactory(self):
...
+
+    for row in data:
+      with self.assertRaises(ValueError) as ve:
+        result = model.run(row)
+      self.assertEqual(str(ve.exception), expMsg) #should fail
+
+
+    # unknown predicted field - TemporalMultiStep model
+    modelConfig['modelParams']['inferenceType']='TemporalMultiStep' # change to MultiStep model
```
scottpurdy: The trailing comment doens't make sense. And the comment on the line before also doesn't make sense.

tests/unit/nupic/frameworks/opf/clamodel_test.py
```diff
@@ -82,7 +83,104 @@ def testTemporalAnomalyModelFactory(self):
...
+
+    model = ModelFactory.create(modelConfig=modelConfig)
+    model.enableLearning()
+    model.enableInference(inferenceArgsWrong)
+
+    for row in data:
+      with self.assertRaises(ValueError) as ve:
+        result = model.run(row)
+      self.assertEqual(str(ve.exception), expMsg) #should fail
```
scottpurdy: The trailing comment says this line should fail but it is outside the `assertRaises` context.

tests/unit/nupic/frameworks/opf/clamodel_test.py
breznak: probably not in this test, but I just reuse the data from previous tests where it has a meaning.

rhyolight:
Whoops.

breznak:
@rhyolight do you know whom to ask about recent changes to clamodel/pyregion/recordsensor about this error? https://travis-ci.org/numenta/nupic/jobs/78262076#L918

rhyolight:
Sorry, no. I'll have to pass that question to @scottpurdy or @chetan51.

scottpurdy:
@marionleborgne  - does that error look related to the network categories change?

scottpurdy:
@marionleborgne and I took a look and don't see how this error is coming from your changes. I just tried rebuilding. If the problem persists then we can look again

nupic/frameworks/opf/clamodel.py
```diff
@@ -703,18 +699,13 @@ def _handleCLAClassifierMultiStep(self, patternNZ,
...
       self._numFields = len(encoderList)
 
       # This is getting index of predicted field if being fed to CLA.
       fieldNames = sensor.getSelf().encoder.getScalarNames()
-      if predictedFieldName in fieldNames:
-        self._predictedFieldIdx = fieldNames.index(predictedFieldName)
+      if self._predictedFieldName in fieldNames:
+        self._predictedFieldIdx = fieldNames.index(self._predictedFieldName)
       else:
         # Predicted field was not fed into the network, only to the classifier
```
breznak: does it make sense to run in this option, or should we raise an exception? As https://github.com/numenta/nupic/pull/2263/files#diff-5bf13635ee2cded8c4bfe689ea9a297dR728 fails and that encoder is used in most of the function here...

rhyolight:
@breznak Please give @scottpurdy a couple of days to review. We are all in internal training sessions outside the office.

tests/unit/nupic/frameworks/opf/clamodel_test.py
```diff
@@ -82,7 +83,103 @@ def testTemporalAnomalyModelFactory(self):
...
+            "Raw input is: %r" % (data))
+
+    for row in data:
+      with self.assertRaises(ValueError) as ve:
+        result = model.run(row)
+        self.assertEqual(str(ve.exception), expMsg) #should fail
+
+
+    # test unknown predicted field check on TemporalMultiStep model
+    modelConfig['modelParams']['inferenceType']='TemporalMultiStep'
```
scottpurdy: space before and after `=` and use double quotes

nupic/frameworks/opf/clamodel.py
```diff
@@ -726,15 +717,15 @@ def _handleCLAClassifierMultiStep(self, patternNZ,
         encoderList = sensor.getSelf().disabledEncoder.getEncoderList()
       else:
         encoderList = []
-      if len(encoderList) >= 1:
+
+      if encoderList: # not empty
```
scottpurdy: Remove the comment. If the code isn't clear enough then change it back to checking the length.

nupic/frameworks/opf/model.py
```diff
@@ -160,6 +160,8 @@ def enableInference(self, inferenceArgs=None):
     """
     self.__inferenceEnabled = True
     self.__inferenceArgs = inferenceArgs
+    self._predictedFieldName = self.__inferenceArgs.get('predictedField', None)
```
scottpurdy: double quotes

tests/unit/nupic/frameworks/opf/clamodel_test.py
```diff
@@ -82,7 +83,103 @@ def testTemporalAnomalyModelFactory(self):
...
+                     u'predictedField': u'c1',
+                     u'predictionSteps': [1]}
+
+    data = [
+      {'_category': [None],
+       '_reset': 0,
+       '_sequenceId': 0,
+       '_timestamp': datetime.datetime(2013, 12, 5, 0, 0),
+       '_timestampRecordIdx': None,
+       u'c0': datetime.datetime(2013, 12, 5, 0, 0),
```
scottpurdy: these don't need to be unicode, just use double quote strings everywhere for consistency

scottpurdy:
done with review.

breznak:
@scottpurdy fixed according to your review.

rhyolight:
Marked in progress because of failing builds.

breznak:
trying to resume the work...

PR #2269, opened 2015-06-29
Fixes bug in mmGetCellActivityPlot

ywcui1990:
@chetan51 Please take a look when you get a chance

This PR is meant to fix a bug in Line 421-424 of temporal_memory_monitor_mixin.py This piece of code covert traces that contain ConnectionsCell to int. However, the traces could also contain int  in many cases, and this conversion caused an error ('int' object has no attribute 'idx'). 

I tried to fix this by add an additional logic in the getCellIndices function in temporal_memory.py


chetan51:
@ywcui1990 Do you know when these traces would contains `int`s? That's where we should address this issue. The point of `getCellIndex` is to be specific to the implementation of temporal memory so that we don't have to check the type of the indices at runtime.

Issue #2277, opened 2015-06-30
TM implementation 'tm_py' broken 


marionleborgne:
When changing the TM implementation from 'cpp' to 'tm_py' (see MODEL_PARAMS['modelParams']['tpParams']['temporalImp']) I get the following error:

Traceback (most recent call last):
  File "/Users/mleborgne/_git/nupic.research/classification/classify.py", line 147, in <module>
    trainAndClassify(TRAINING_SET_SIZE, model, dataPath, resultsPath)
  File "/Users/mleborgne/_git/nupic.research/classification/classify.py", line 78, in trainAndClassify
    result = model.run(modelInput)
  File "/Users/mleborgne/_git/nupic/nupic/frameworks/opf/clamodel.py", line 386, in run
    self._tpCompute()
  File "/Users/mleborgne/_git/nupic/nupic/frameworks/opf/clamodel.py", line 486, in _tpCompute
    tp.compute()
  File "/Users/mleborgne/_git/nupic/nupic/engine/__init__.py", line 443, in compute
    return self._region.compute()
  File "/Users/mleborgne/_git/nupic/nupic/bindings/engine_internal.py", line 1299, in compute
    return _engine_internal.Region_compute(self)
  File "/Users/mleborgne/_git/nupic/nupic/regions/TPRegion.py", line 450, in compute
    self._compute(inputs, outputs)
  File "/Users/mleborgne/_git/nupic/nupic/regions/TPRegion.py", line 481, in _compute
    tpOutput = self._tfdr.compute(buInputVector, self.learningMode, self.inferenceMode)
  File "/Users/mleborgne/_git/nupic/nupic/research/TP_shim.py", line 82, in compute
    numberOfCells = self.connections.numberOfCells()
AttributeError: 'Connections' object has no attribute 'numberOfCells'

marionleborgne:
cc @chetan51

PR #2281, opened 2015-06-30
Fixes broken TP Shim

chetan51:
Fixes https://github.com/numenta/nupic/issues/2277

### DIFF
--- a/nupic/research/TP_shim.py
+++ b/nupic/research/TP_shim.py
@@ -79,7 +79,7 @@ class TPShim(TemporalMemory):
     """
     super(TPShim, self).compute(set(bottomUpInput.nonzero()[0]),
                                             learn=enableLearn)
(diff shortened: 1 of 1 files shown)


chetan51:
This fixes all prediction models, but anomaly detection models are still broken (due to a different issue).

chetan51:
@scottpurdy

chetan51:
cc: @marionleborgne

scottpurdy:
Do we want to switch from a getter to direct access? Does this work for C++ version of `Connections`?

chetan51:
@scottpurdy Good point and catch. I've updated the fix to work for both versions of `Connections`.

chetan51:
@scottpurdy Bump

scottpurdy:
:+1:

Issue #1360, opened 2014-09-22
OPF cannot construct growing-sized CLA 


breznak:
It's possible to construct a TP as a fixed, or growing size model

The default is growing size, which is possible to initialize `TP()` (did not check if other methods fail to run later with it).

When I set this in `permutations.py` in swarming, I get:

```
Results from all experiments:
----------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mmm/nupic/nupic-source/scripts/run_swarm.py", line 187, in <module>
    runPermutations(sys.argv[1:])
  File "/home/mmm/nupic/nupic-source/scripts/run_swarm.py", line 181, in runPermutations
    fileArgPath, optionsDict, outputLabel, permWorkDir)
  File "/home/mmm/nupic/nupic-source/nupic/swarming/permutations_runner.py", line 346, in runWithPermutationsScript
    return _runAction(runOptions)
  File "/home/mmm/nupic/nupic-source/nupic/swarming/permutations_runner.py", line 217, in _runAction
    returnValue = _runHyperSearch(runOptions)
  File "/home/mmm/nupic/nupic-source/nupic/swarming/permutations_runner.py", line 160, in _runHyperSearch
    metricsKeys=search.getDiscoveredMetricsKeys())
  File "/home/mmm/nupic/nupic-source/nupic/swarming/permutations_runner.py", line 822, in generateReport
    raise Exception(jobInfo.workerCompletionMsg)
Exception: E10002: Exiting due to receiving too many models failing from exceptions (6 out of 6). 
Model Exception: Exception occurred while running model 77094: OverflowError("in method 'Cells4_setMaxSegmentsPerCell', argument 2 of type 'int'",) (<type 'exceptions.OverflowError'>)
Traceback (most recent call last):
  File "/home/mmm/nupic/nupic-source/nupic/swarming/utils.py", line 430, in runModelGivenBaseAndParams
    (completionReason, completionMsg) = runner.run()
  File "/home/mmm/nupic/nupic-source/nupic/swarming/ModelRunner.py", line 245, in run
    self._model = ModelFactory.create(modelDescription)
  File "/home/mmm/nupic/nupic-source/nupic/frameworks/opf/modelfactory.py", line 80, in create
    return modelClass(**modelConfig['modelParams'])
  File "/home/mmm/nupic/nupic-source/nupic/frameworks/opf/clamodel.py", line 203, in __init__
    clParams, anomalyParams)
[...]

PR #2282, opened 2015-06-30
Fix TP non "fixed-sized" mode

breznak:
Fixes: #1360  

DON'T merge until both Py, C++, TP10X2 are resolved!
c++ sibling: https://github.com/numenta/nupic.core/pull/484

- [ ] revert `.nupic_modules` pointing to my local SHA
- [x] python non fixed-sized TP works
- [ ] fix binding from c++ in TP10X2

### DIFF
--- a/nupic/research/TP.py
+++ b/nupic/research/TP.py
@@ -85,8 +85,8 @@ class TP(ConsolePrinterMixin):
                maxLrnBacktrack=5,
                maxAge=100000,
                maxSeqLength=32,
-               maxSegmentsPerCell=-1,
-               maxSynapsesPerSegment=-1,
+               maxSegmentsPerCell=0,
+               maxSynapsesPerSegment=0,
                outputType='normal',
               ):
     """
--- a/nupic/research/TP.py
+++ b/nupic/research/TP.py
@@ -123,7 +123,8 @@ class TP(ConsolePrinterMixin):
                   is used to turn on "fixed size CLA" mode. When in effect,
                   globalDecay is not applicable and must be set to 0 and
                   maxAge must be set to 0. When this is used (> 0),
-                  maxSynapsesPerSegment must also be > 0.
+                  maxSynapsesPerSegment must also be > 0. For not fixed-sized HTM
+                  use 0.
 
     @param maxSynapsesPerSegment The maximum number of synapses allowed in a segment.
                   This is used to turn on "fixed size CLA" mode. When in effect,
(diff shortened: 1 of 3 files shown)


.nupic_modules
```diff
@@ -1,3 +1,3 @@
 # Default nupic.core dependencies (override in optional .nupic_config)
-NUPIC_CORE_REMOTE = 'git://github.com/numenta/nupic.core.git'
-NUPIC_CORE_COMMITISH = '140a176ef1f53bb5014e60cec5a26ecc0498d23e'
+NUPIC_CORE_REMOTE = 'git://github.com/breznak/nupic.core.git' # FIXME revert before merge!
```
breznak: this needs to be reverted before final merge!

breznak:
@oxtopus @scottpurdy @rhyolight please review. 

The Py part runs fine. C++ has error: 

```
  File "/home/mmm/devel/nupic/nupic-source/tests/unit/nupic/research/tp_test.py", line 106, in testCheckpointMiddleOfSequence
    tp1.compute(bottomUpInput, True, True)
  File "/home/mmm/devel/nupic/nupic-source/nupic/research/TP10X2.py", line 312, in compute
    y = self.cells4.compute(bottomUpInput, computeInfOutput, enableLearn)
  File "/home/mmm/devel/nupic/nupic-source/nupic/bindings/algorithms.py", line 2219, in compute
    return _algorithms.Cells4_compute(self, *args)
RuntimeError: ASSERTION FAILED: "segIdx < _segments.size()" 

```

rhyolight:
Should we still do this? The temporal memory algorithm replaces this.

breznak:
@rhyolight can we add 'wontfix' label on such closed PRs, so it's possible to revisit when needed? 

Actually, I have a proposal: can we merge the part "fixes python part of the issue", even if we don't have compatibility with c++ version (we don't have it anyway as it does not work), atleast it's working for python. 
Or we should evaluate the functionality (still useful?) and remove completely.

rhyolight:
I would like to move ahead with TM and get rid of the TP as soon as we can. @scottpurdy What do you think about it?

scottpurdy:
@rhyolight - Tons of work before we could consider that. AFAIK, TP is still better for time series problems.

rhyolight:
Ok then, we should reopen this!

rhyolight:
Closing due to lack of activity. Please re-open if progress is made.

PR #2286, opened 2015-06-30
Removed stripUnlearnedColumns-from-SPRegion

### LINKED ISSUE #2279: Remove stripUnlearnedColumns from compute in SPRegion
Finishes replacing outdated issue #1262 (followup to issue #2278)

```stripUnlearnedColumns``` does not need to be default behavior within compute. This will only affect the case of a spacial pooler region being run with learning mode disabled AFTER learning has occurred. 

After ```stripUnlearnedColumns``` is removed, the default behavior for this case will be for active, unlearned columns to remain in the output of the spacial pooler if learning is disabled. This does not affect the spacial pooler if learning is enabled or if it is being with only it's randomly initialized values (no learning enabled).

jaredweiss:
Fixes #2279

Removes `stripUnlearnedColumns` from `_doBottomUpCompute` in SP. 

`stripUnlearnedColumns` should only be used in special cases where it is specifically intended to be run on the output of the Spatial Pooler. It should not be running by default (in the current implementation of NuPIC it will only run by default in the case that learning is disabled AND the model has previously learned. This change removes it from ever running by default)

### DIFF
--- a/nupic/regions/SPRegion.py
+++ b/nupic/regions/SPRegion.py
@@ -590,13 +590,7 @@ class SPRegion(PyRegion):
     inputVector = numpy.array(rfInput[0]).astype('uint32')
     outputVector = numpy.zeros(self._sfdr.getNumColumns()).astype('uint32')
 
-    # Don't strip unlearned columns if learning is off and the SP hasn't
-    # learned anything yet. This acts as a random SP.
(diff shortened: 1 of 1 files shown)


jaredweiss:
@scottpurdy Please take a look at this

scottpurdy:
Have you validated that this doesn't affect hotgym?

jaredweiss:
Yes. Hotgym isn't affected

Issue #2260, opened 2015-06-28
AnomalyLikelihood with default timestamp behavior


breznak:
When time stamp is not specified to `AnomalyLikelihood.anomalyProbability()` it defaults to `None` which results in using current time! `datettime.now()`. 

That was done for convenience when you can't/don't want to pass the `timestamp` manually. On the other hand, passing `None`, one would expect it is ignored - which is not. As seen in : 
https://github.com/breznak/nupic/commit/0fc0acfdc8ae977f9d6f9f995cd046ead04021eb

Possible solution: 
- make 2 "default" choices, eg 
  * `"now"` uses current time (as used now) 
  * `None` would use iterations instead of times, so 
```
l1 = AnomalyLikelihood()
l2 = AnomalyLikelihood()

l1.anomalyProbability("something", 0.45)
l2.anomalyProbability("something", 0.45)

l1 == l2 # True
```

breznak:
@subutai ?

rhyolight:
This makes sense to me. I do not think the current time should be used if
no timestamp is sent. Each row must have a user-provided timestamp.

---------
Matt Taylor
OS Community Flag-Bearer
Numenta

PR #2288, opened 2015-07-01
fix likelihood default timestamp - use iteration step

breznak:
instead of datetime.now() for equality of 2 likelihood instances
called with same numbers withnout timestamps.

Fixes: #2260


breznak:
@rhyolight please merge

rhyolight:
Before merging, I want @scottpurdy or @subutai to give their opinion. I may be wrong.

scottpurdy:
Seems reasonable but I'll defer to @subutai in case he knows why we had it default to the current time.

breznak:
> That was done for convenience when you can't/don't want to pass the timestamp manually.  

When I added the functionality to call likelihood from Anomaly class, some records wouldn't pass timestamp, instead pass None, so we added this work-around.

rhyolight:
@subutai Your opinion is requested above.

Issue #2314, opened 2015-07-14
TemporalMemory.getCellIndex doesn't work correctly when running through OPF


chetan51:
When running the TM through the OPF, the function `getCellIndex` doesn't work correctly; there are errors stating that the `idx` property can't be accessed.

PR #2315, opened 2015-07-14
Fix: TemporalMemory.getCellIndex doesn't work correctly when running through OPF

chetan51:
Fixes https://github.com/numenta/nupic/issues/2314.


chetan51:
@scottpurdy

scottpurdy:
:+1:

Issue #2316, opened 2015-07-14
Remove unused synapses in Temporal Memory


chetan51:
Over time, the TM builds up many synapses with permanence 0. These can be deleted for cleanliness and performance.

relates to numenta/nupic.core#495

PR #2317, opened 2015-07-14
Remove unused synapses in Temporal Memory

chetan51:
Fixes https://github.com/numenta/nupic/issues/2316.

I confirmed that this improves performance when running through a large hotgym dataset.


chetan51:
@subutai @scottpurdy

chetan51:
Talked with @scottpurdy; will move this logic to TemporalMemory.

PR #2320, opened 2015-07-14
Remove unused synapses in Temporal Memory

chetan51:
Fixes https://github.com/numenta/nupic/issues/2316.

I verified that it improves accuracy and speed for the Python `TemporalMemory`, improves accuracy and has no impact on speed for the `FastTemporalMemory` (tested by running through a large hotgym dataset).

### DIFF
--- a/nupic/research/temporal_memory.py
+++ b/nupic/research/temporal_memory.py
@@ -597,6 +597,8 @@ class TemporalMemory(object):
     @param permanenceIncrement  (float)  Amount to increment active synapses
     @param permanenceDecrement  (float)  Amount to decrement inactive synapses
     """
+    # Need to copy synapses for segment set below because it will be modified
+    # during iteration by `destroySynapse`
     for synapse in set(connections.synapsesForSegment(segment)):
       synapseData = connections.dataForSynapse(synapse)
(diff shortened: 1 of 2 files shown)


chetan51:
@subutai @scottpurdy

nupic/research/temporal_memory.py
```diff
@@ -593,7 +597,7 @@ def adaptSegment(self, segment, activeSynapses, connections,
     @param permanenceIncrement  (float)  Amount to increment active synapses
     @param permanenceDecrement  (float)  Amount to decrement inactive synapses
     """
-    for synapse in connections.synapsesForSegment(segment):
```
scottpurdy: will `synapsesForSegment` really return the same synapse multiple times?

chetan51: No, it's because `destroySynapse` modifies the set of synapses while it's being iterating over, so this change is necessary to avoid an error.

scottpurdy: ahh makes sense. You might want to put a comment right before to indicate this

scottpurdy:
done with review, :+1: when you're ready

PR #2323, opened 2015-07-16
Adds hotgym regression test to ensure that prediction results don't c…

### LINKED ISSUE #2322: Create OPF regression test
We should have a test to ensure that any changes to OPF results are intentional. This can be a simple test that runs hotgym and enforces that the error metric matches. If it fails then the expected result must be explicitly updated to match the new result.

scottpurdy:
…hange unintentionally.

fixes #2322

@vitaly-krugl - please review

### DIFF
--- a/tests/integration/nupic/opf/hotgym_regression_test.py
+++ b/tests/integration/nupic/opf/hotgym_regression_test.py
@@ -40,6 +40,8 @@ class HotgymRegressionTest(unittest.TestCase):
   """Hotgym regression test to validate that predictions don't change."""
 
 
+  @unittest.skip("This currently fails on Travis. The issue is being tracked "
+                 "at: https://github.com/numenta/nupic/issues/2358")
   def testHotgymRegression(self):
     experimentDir = pkg_resources.resource_filename(
         "nupic",
--- a/tests/integration/nupic/opf/hotgym_regression_test.py
+++ b/tests/integration/nupic/opf/hotgym_regression_test.py
@@ -1,0 +1,77 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2015, Numenta, Inc.  Unless you have purchased from
+# Numenta, Inc. a separate commercial license for this software code, the
+# following terms and conditions apply:
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


tests/integration/nupic/opf/hotgym_regression_test.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class HotgymRegressionTest(unittest.TestCase):
+  """Hotgym regression test to validate that predictions don't change."""
+
+
+  def testHotgymRegression(self):
+    experimentDir = pkg_resources.resource_filename(
+        "nupic",
+        os.path.join("..", "examples", "opf", "experiments", "multistep",
```
vitaly-krugl: nit: `os.path.pardir` instead of `".."`

tests/integration/nupic/opf/hotgym_regression_test.py
```diff
@@ -0,0 +1,77 @@
...
+          resultsDir, "DefaultTask.TemporalMultiStep.predictionLog.csv")
+      with open(resultsPath) as f:
+        reader = csv.reader(f)
+        headers = reader.next()
+        self.assertEqual(headers[12],
+                         "multiStepBestPredictions:multiStep:errorMetric='aae':"
+                         "steps=1:window=1000:field=consumption")
+        lastRow = collections.deque(reader, 1)[0]
+
+      # Changes that affect prediction results will cause this test to failure.
```
vitaly-krugl: "failure" -> "fail"

tests/integration/nupic/opf/hotgym_regression_test.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+  def testHotgymRegression(self):
+    experimentDir = pkg_resources.resource_filename(
+        "nupic",
+        os.path.join("..", "examples", "opf", "experiments", "multistep",
+                     "hotgym")
+    )
+
+    resultsDir = os.path.join(experimentDir, "inference")
```
vitaly-krugl: If two tests both using the hotgym experment were to run concurrently, then they would step on each other in these directories.

vitaly-krugl:
Other than a few nits, LGTM

tests/integration/nupic/opf/hotgym_regression_test.py
scottpurdy: I'm not sure there is any way around this with the current experiment runner. Note that this path I construct is something created by the experiment runner. I create the path just so I can read the results and clean them up afterwards.

scottpurdy:
I will merge this when green and we can enable the test once we figure out #2328

PR #2330, opened 2015-07-17
Move registerBuildInRegion call in nupic.engine.__init__ to global scope

### LINKED ISSUE #2329: Move registerBuildInRegion call in nupic.engine.__init__ to global scope
`registerBuildInRegion` needs to be run before any of the methods in `nupic.engine.__init__` are run. Otherwise, if `registerPyRegion` or `unregisterPyRegion` is run before a network is initialized (as would happen if loading a serialized network with regions that conflict with the built ins), `registerBuildInRegion` would either break or not have the proper region in it's list of regions.

jaredweiss:
Fixes #2329

Also fixed the spelling of registerBuildInRegion to registerBuiltInRegion

@scottpurdy Please review


scottpurdy:
:+1:

PR #2337, opened 2015-07-22
Hyper search decoupling

mihail911:
Added a HyperSearch directory and moved two potentially deprecated .tpl files. Please review @scottpurdy  @oxtopus


scottpurdy:
Can you just move the entire experiment generator directory over?

mihail911:
With a bit more work, that should be possible. I will make those changes appropriately. Also do not merge this yet, as I just realized this causes some tests to fail.

PR #2339, opened 2015-07-23
FileRecordStream support multiple categories

### LINKED ISSUE #2335: FileRecordStream with multiple categories
Currently FileRecordStream only accepts one category label per data instance; [spec is here](https://github.com/numenta/nupic/blob/master/nupic/data/file_record_stream.py#L132). We would like the functionality for multiple category labels per row; i.e. multiple columns labeled with `C`.

BoltzmannBrain:
Fixes #2335


BoltzmannBrain:
@scottpurdy please take a look. The main change here is `FileRecordStream._categoryIdx` is now a list of ints instead of a single int. The affected regions have been modified accordingly, examples (e.g. examples/network/hierarchy_network_demo.py) run as expected, and tests pass (fails CI build b/c `Error: Failed to download resource "mysql"`)... What else needs to be done to ensure backwards-compatibility?

scottpurdy:
@BoltzmannBrain - is this ready for review? Would be nice to squash down to clear set of commits first

BoltzmannBrain:
@scottpurdy not yet ready; not passing opf tests. I'll let you know.

BoltzmannBrain:
@scottpurdy please review

scottpurdy:
@BoltzmannBrain - My initial take is that a new list field type would make sense for this. Otherwise the "string" type is interpreted different based on whether it is marked as a category or not. It would also be useful for the pass through encoders.

@chetan51 - thoughts?

chetan51:
@scottpurdy I agree.

BoltzmannBrain:
@scottpurdy please take a look now. Sorry but the (not)squashed commits, I'm unsure why it's not working correctly. Note the enum for list is already defined ([here](https://github.com/numenta/nupic/blob/master/nupic/data/fieldmeta.py#L92)).

nupic/data/utils.py
```diff
@@ -141,9 +132,21 @@ def parseSdr(s):
   return sdr
 
 
-
```
scottpurdy: three blank lines before top level functions

nupic/data/utils.py
```diff
@@ -141,9 +132,21 @@ def parseSdr(s):
...
   """
 
   return "".join(str(bit) for bit in sdr)
+
+
+def parseStringList(s):
+  """Parse a string of space-separated numbers, returning a Python list."""
+  assert isinstance(s, basestring)
+  ints = []
+  for val in s:
```
scottpurdy: how about this:

```
assert isinstance(s, basestring)
return s.split()
```

nupic/regions/RecordSensor.py
```diff
@@ -30,49 +30,35 @@ class RecordSensor(PyRegion):
   """
   A Record Sensor (RS) retrieves an information "record" and encodes
   it to be suitable as input to an HTM.
-
```
scottpurdy: Why are all these blank lines removed? Doesn't that make it harder to read?

scottpurdy:
I keep seeing first_order_0.csv being changed in PRs so I think some test modifies it. I'm not sure if we should remove it or what but for now please remove from this PR.

This PR is really big and has a handful of different changes:

- adding the "list" field type
- extending the "category" special flag to support both "int" and "list" field types and making corresponding network changes
- updating the knn classifier to support multiple input categories
- etc

Can you break this up into separate PRs for each? That probably sounds tedious but it will really speed up the review. With all of this together it is going to take many rounds of back and forth to get everything right.

BoltzmannBrain:
@scottpurdy okay, I'll split it up according to the subtasks now defined in #2335.

BoltzmannBrain:
Changes incorporated into PRs #2367, #2368, and #2370.

PR #2346, opened 2015-07-27
Allow Connections to be serialized.

### LINKED ISSUE #2347: Connections class cannot be pickled
The `SynapseData` namedtuple cannot be pickled.

scottpurdy:
fixes #2347

@chetan51 - What do you think?

@jaredweiss fyi

### DIFF
--- a/nupic/research/connections.py
+++ b/nupic/research/connections.py
@@ -230,8 +230,8 @@ class Connections(object):
 
     data = self._synapses[synapse]
     newData = SynapseData(data.segment,
-                               data.presynapticCell,
-                               permanence)
+                          data.presynapticCell,
+                          permanence)
     self._synapses[synapse] = newData
 
     # Update indexes
--- a/nupic/research/connections.py
+++ b/nupic/research/connections.py
@@ -412,4 +412,3 @@ class Connections(object):
     """
     if permanence < 0 or permanence > 1:
       raise ValueError("Invalid permanence")
-
--- a/nupic/research/connections.py
+++ b/nupic/research/connections.py
@@ -23,14 +23,28 @@ from collections import defaultdict, namedtuple
 
 
 
+class SynapseData(object):
+
+
+  __slots__ = ("segment", "presynapticCell", "permanence")
+
+
+  def __init__(self, segment, presynapticCell, permanence):
+    self.segment = segment
+    self.presynapticCell = presynapticCell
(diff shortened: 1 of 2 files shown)


nupic/research/connections.py
```diff
@@ -215,7 +229,7 @@ def updateSynapsePermanence(self, synapse, permanence):
     self._validatePermanence(permanence)
 
     data = self._synapses[synapse]
-    newData = self.SynapseData(data.segment,
+    newData = SynapseData(data.segment,
                                data.presynapticCell,
```
chetan51: Alignment

chetan51:
LGTM. There may be some code that expects `SynapseData` to be a tuple, but I doubt it.

nupic/research/connections.py
```diff
@@ -23,14 +23,28 @@
...
+
+
 class Connections(object):
   """
   Class to hold data representing the connectivity of a collection of cells.
   """
 
-  SynapseData = namedtuple("SynapseData", ["segment",
-                                           "presynapticCell",
-                                           "permanence"])
```
rcrowder: @chetan51 From my limited knowledge, would this affect nupic_studio. Issues revolving around "TypeError: 'SynapseData' object is not iterable' in this line https://github.com/nupic-community/nupic.studio/blob/master/nupic_studio/htm/node_region.py#L472 Be great if you could let me know if it is, and whether thee is an easy fix in nupic_studio for it?

chetan51: I don't know about nupic studio, but I would open an issue in nupic studio if this change breaks it.

rcrowder: This was raised in https://github.com/nupic-community/nupic.studio/issues/29#issuecomment-142504682 and recently updated with a new user facing this problem. I recreated this locally. Let me know if there is a simple Py change for this (potentially an isolated change just within nupic.studio to cope)?

rhyolight: I agree that the fix needs to be in NuPIC Studio. I'm not sure about how easy it will be to fix, though. As internal architectures change in NuPIC, this could continue to be a problem. It really won't be solved until we get to NuPIC 1.0 and an official public API.

rcrowder: Was an existing issue I was hoping someone would see a quick fix for; https://github.com/nupic-community/nupic.studio/issues/29

PR #2354, opened 2015-07-28
Makes nupic a namespace package that other projects can extend.

### LINKED ISSUE #2353: Make nupic a namespace package
Making nupic a namespace package will enable other projects to extend nupic. One example is nupic.vision, where we have moved the vision-related code. This setup enables NuPIC users that don't need vision-related code to have a smaller install with fewer dependencies while still giving the vision-related code nice import paths.

scottpurdy:
fixes #2353

@rhyolight - Does removing the version like this introduce issues? The documentation for namespace packages prohibits anything else from being in the `__init__.py` for a namespace package and [this documentation on version](https://www.python.org/dev/peps/pep-0396/) backs this up.

@oxtopus - please review

### DIFF
--- a/nupic/__init__.py
+++ b/nupic/__init__.py
@@ -1,6 +1,6 @@
 # ----------------------------------------------------------------------
 # Numenta Platform for Intelligent Computing (NuPIC)
-# Copyright (C) 2013, Numenta, Inc.  Unless you have an agreement
+# Copyright (C) 2013-2015, Numenta, Inc.  Unless you have an agreement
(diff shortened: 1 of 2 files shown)


rhyolight:
> Does removing the version like this introduce issues?

No, this is much better. It simplifies the [release script](https://github.com/rhyolight/nupic.devtools/blob/master/bin/release). Ping me when you merge this so I can update it, please.

scottpurdy:
@oxtopus please review. Needed for numenta/nupic.vision#32

oxtopus:
:+1:

PR #2377, opened 2015-08-05
Update codebase to be Python 3.x compatible, while maintaining Python 2.7.x support

oxtopus:
Note: This is not full support for Python 3.x, however, the _syntax_ is compatible with both 2.7.x and 3.x.

Most of the changes are from the `2to3` tool, with some manual fixups to ensure that the result is backwards compatible with 2.  For example, all `prints` are now `print()`'s, with the manual addition of `from __future__ import print_function`.  In other areas, `six` is used for portability.


nupic/regions/Spec.py
```diff
@@ -215,7 +215,7 @@ def items2dict(items):
       attributes. The entire items dict will become a dict of dicts (same keys).
       """
       d = {}
-      for k, v in items.items():
+      for k, v in list(items.items()):
```
vitaly-krugl: With the understanding that this PR is just proof of concept, I wanted to call attention to the extra, unnecessary overhead that the automatic conversion tool added to the code in cases like this one, which is a pretty common case. Since we're not modifiying the items dict inside the loop, the `list(...)` wrapper (instantiation and population of it) is completely unnecessary overhead.

vitaly-krugl: This addition of the unnecessary `list(...)` wrapper by the automatic conversion tool actually highlights the incorrect use of `items()` in our code base in a number of places. Since the body of the loop doesn't alter the `items` dict, the original call should have been `iteritems()`.

nupic/algorithms/CLAClassifier.py
```diff
@@ -558,13 +560,13 @@ def read(cls, proto):
     classifier._patternNZHistory = deque(maxlen=max(classifier.steps) + 1)
     patternNZHistoryProto = proto.patternNZHistory
     learnIteration = classifier._learnIteration - len(patternNZHistoryProto) + 1
-    for i in xrange(len(patternNZHistoryProto)):
+    for i in range(len(patternNZHistoryProto)):
```
vitaly-krugl: This is an example of an automatic change that is inappropriate in most circumstances for python 2.x:

1. it adds unnecessary overhead of creating a list on python 2.x
2. In the cases where the parameter to `xrange` was a really large number, this would generate a huge list on python 2.x

oxtopus: More appropriate change would have been to leave as-is and use `from six.moves import xrange` (see http://pythonhosted.org/six/#module-six.moves <http://pythonhosted.org/six/#module-six.moves>)

oxtopus: There are also alternative approaches to the original code to consider, too.  For example, I think the original `for i in xrange(len())` was the wrong approach.  It would have been better to use `enumerate()`, since `patternNZHistoryProto[i]` is used inside the loop.

nupic/data/jsonhelpers.py
```diff
@@ -78,7 +80,7 @@ def validate(value, **kwds):
           ValidationError when value fails json validation
   """
 
-  assert len(kwds.keys()) >= 1
+  assert len(list(kwds.keys())) >= 1
```
vitaly-krugl: This kind of conversion would have been more efficient as `assert kwds` or `assert len(kwds)` or `assert len(kwds) >=1`

nupic/data/record_stream.py
```diff
@@ -67,19 +67,19 @@ def getNextRecordDict(self):
...
-    
+
     if not values:
       return dict()
-    
+
     # Create the return dict
-    result = dict(zip(self.getFieldNames(), values))
-    
+    result = dict(list(zip(self.getFieldNames(), values)))
```
vitaly-krugl: I think that this would have still worked correctly as `result = dict(zip(self.getFieldNames(), values))`, without the unnecessary overhead of instantiating and populating the `list` object.

vitaly-krugl:
Realizing that this is a proof of concept, a spot-check of the changes made by the automatic conversion tool suggests that if left to its own devices, it would pollute the code base with many unnecessary and inefficient data transformations.

oxtopus:
@vitaly-krugl Understood, however, I draw a different conclusion.  I think the `2to3` conversion tool is doing the right thing assuming your intention is to A) switch to python 3 and B) preserve existing behavior.  Left to its own devices, yes, the modified code in Python 2.7 would be slower than the original in Python 2.7, but it would be on par with Python 3.x.  In most cases, the original code could be rewritten (or have been written from the start) such that generators and iterators are used in 2.7, and then when translated to 3, also use generators and iterators.  It's only a degradation because we're already using the slow version before the conversion AND staying in 2.x AFTER the conversion, and not necessarily the fault of the tool

Meanwhile, [2to3](https://docs.python.org/2/library/2to3.html) has some options for disabling some of the fixers.  For example, if we add `from future_builtins import zip`, then the `dict(zip(self.getFieldNames(), values))` example would not apply.  There are also lots of missed opportunities to use [six](http://pythonhosted.org/six/#module-six.moves).

Ultimately, I think the exercise was a success in that despite the large and storied code base, a translation is possible with minimal effort.  Right now, I'm interested in knowing whether or not these changes break existing 2.7 installations.  So far, I can only use my own anecdotal experience and tests, but I know that's not 100% coverage.  I'm hoping someone else will try it out and help expose holes in test coverage.

vitaly-krugl:
@oxtopus, I agree that as a proof of concept, this exercise is a success, especially if we wish to produce a separate one-off nupic just for the Windows Unicorn app. 

However, if we wish to modify the existing public nupic code base, I believe that we would need some additional intervention to avoid polluting the code with unnecessary overhead and complexity. For example:
* changes like `dict(list...` add unnecessary overhead in both python 3.x and 2.x.
* changes like `len(list(kwds.keys())) >= 1` make the code work on python 3.x, while adding unnecessary overhead on both 3.x and 2.x and making the code less clean/more complicated
* changes like `for i in xrange(len(patternNZHistoryProto))` -> `for i in range(len(patternNZHistoryProto))` preserve the same level of performance in 3.x, but add unnecessary overhead in 2.x
* It might be a good opportunity to fix up the other case where legacy code was using the wrong primitive or pattern, since the 2to3 tool's code changes do such a good job of bringing those faults to our attention.

oxtopus:
> - It might be a good opportunity to fix up the other case where legacy code was using the wrong primitive or pattern, since the 2to3 tool's code changes do such a good job of bringing those faults to our attention.

:+1: for fixing legacy codebase _before_ translation.

rhyolight:
@oxtopus So this PR is not actually meant to be merged?

oxtopus:
At least not until it has the "ready" label

hilojack:
nupic for python3 failed with unitest2.

> `unittest2` has the project name "unittest2py3k" but uses the same package name ("unittest2")

`pip3 install nupic
Collecting nupic
  Using cached nupic-0.5.4.tar.gz
......
Collecting unittest2==0.5.1 (from nupic)
  Using cached unittest2-0.5.1.zip
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/private/var/folders/73/7vxr7kzs09ndh3kwzw9zpdj80000gn/T/pip-build-ktr9f4zj/unittest2/setup.py", line 12, in <module>
        from unittest2 import __version__ as VERSION
      File "/private/var/folders/73/7vxr7kzs09ndh3kwzw9zpdj80000gn/T/pip-build-ktr9f4zj/unittest2/unittest2/__init__.py", line 40, in <module>
        from unittest2.collector import collector
      File "/private/var/folders/73/7vxr7kzs09ndh3kwzw9zpdj80000gn/T/pip-build-ktr9f4zj/unittest2/unittest2/collector.py", line 3, in <module>
        from unittest2.loader import defaultTestLoader
      File "/private/var/folders/73/7vxr7kzs09ndh3kwzw9zpdj80000gn/T/pip-build-ktr9f4zj/unittest2/unittest2/loader.py", line 92
        except Exception, e:
                        ^
    SyntaxError: invalid syntax

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/73/7vxr7kzs09ndh3kwzw9zpdj80000gn/T/pip-build-ktr9f4zj/unittest2/`

PR #2380, opened 2015-08-05
allows us to specify a nupic core branch

akhilaananthram:
This addresses:
File "/opt/numenta/products/infrastructure/infrastructure/utilities/nupic/build_commands.py", line 142, in getNuPICCoreDetails
    branch = core["NUPIC_CORE_BRANCH"] if not nupicCoreBranch else nupicCoreBranch
KeyError: 'NUPIC_CORE_BRANCH'

It is trying to get "NUPIC_CORE_BRANCH" from .nupic_modules

@jcasner

### DIFF
--- a/.nupic_modules
+++ b/.nupic_modules
@@ -1,3 +1,4 @@
 # Default nupic.core dependencies (override in optional .nupic_config)
 NUPIC_CORE_REMOTE = 'git://github.com/numenta/nupic.core.git'
 NUPIC_CORE_COMMITISH = 'eb0b420973f54536debd9d13334a88cdc05bf5c1'
+NUPIC_CORE_BRANCH = 'master'


jcasner:
:+1:

akhilaananthram:
@scottpurdy: please review.  This is for the nupic-pipeline.  If we add this to .nupic_modules, then we can test different branches of nupic.core.

scottpurdy:
That sounds like a problem with that pipeline. I don't have a problem with this but don't know the intent so will defer to @rhyolight

rhyolight:
:+1: but while you are at it, perhaps we should remove `NUPIC_CORE_REMOTE` if it is truly not being used. @oxtopus  what do you think?

oxtopus:
:-1: `NUPIC_CORE_COMMITISH` and `NUPIC_CORE_BRANCH` are technically redundant

rhyolight:
Oh yeah, that's true. Now this doesn't make sense to me. @akhilaananthram are you expecting the branch to be on your remote? Because the primary fork should only ever have one branch: `master`. 

If so, what you really want is to re-enable the usage of the `NUPIC_CORE_REMOTE` value in the pipelines so you can run SHAs that exist on different remote locations.

akhilaananthram:
Yes, that was for when the branch is on my remote.  NUPIC_CORE_REMOTE is enabled in the pipeline.  I'll remove the redundancy

Issue #2340, opened 2015-07-25
compute() function not implemented in CLAClassifierRegion.py


marionleborgne:
The `compute` [1] function is not implemented. Consequently - after creating a Network instance with a CLAClassifierRegion - `network.run()` won't feed data to the CLAClassifier correctly, and we have to use the `customComputeFunction` [2] in an outer loop [3].

[1] https://github.com/numenta/nupic/blob/master/nupic/regions/CLAClassifierRegion.py#L222-L232
[2] https://github.com/numenta/nupic/blob/master/nupic/regions/CLAClassifierRegion.py#L235-L263
[3] https://github.com/numenta/nupic.research/blob/ab3bc24e1523d604723bb872ff2e6ac2b0dc8f53/classification/classify_active_cells.py#L157-L163

rhyolight:
@scottpurdy How important is this issue?

scottpurdy:
I'd say fairly high. It falls in line with the roadmap priorities around flexible network API. That said, it is more of a code hygiene issue than functionality so may still be a P3. But we should tackle it in the near term.

PR #2415, opened 2015-08-11
[nupic issue 2340] fix compute() function in CLACLassifierRegion

marionleborgne:
@scottpurdy: fixes https://github.com/numenta/nupic/issues/2340

This fixes the `compute()` function in `CLAClassifierRegion.py` in the case of *one-step prediction*. Warning: fix doesn't work for multi-step prediction. To be able to output classification results for multi step prediction, we would need to be able to output a dict: `{stepNum: <classification_results> }` where `<classification_results>` is an array. That not possible with the way the network API is setup today: it only outputs arrays.

If we were to stick to one-step prediction with the CLA classifier, that fix would work and we would be able to use the `compute()` function. However, this is not the case and we most likely want to keep using multi-step predictions -- this is why the `customCompute()` function was used in the first place.

That might be a good argument to create this new classification-specific region (where there would be no notion of not multi-step inferences)


scottpurdy:
@marionleborgne - How about we have two outputs for the region:

- actualValues - The actual value associated with each bucket
- probabilities - The probability (as a fraction 0.0 to 1.0) of each bucket for the first prediction step, followed by the probability for each bucket for the second prediction step, etc. So the size of this will be `(numPredictionSteps * numBuckets)`. This is the same as using [numpy flatten](http://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.flatten.html) to turn a multidimensional array into one dimension

Does this work? The nice thing is if you have just one prediction step then you don't have to do any logic to see which probabilities correspond to which prediction steps.

marionleborgne:
@scottpurdy: It's a good idea! I'm going to try that.

marionleborgne:
@scottpurdy Implemented your idea. I tested it locally and the `compute` function now works even for multi-steps prediction :grin: The regions outputs `probabilities` (size `numCategories * numSteps`), `actualValues` (size `numCategories`) and classification results (size `numSteps`). The last one is the winning bucket for each step.

marionleborgne:
@scottpurdy FYI, I also added an additional functionality: you now can train the CLA classifier on multi-class input.

nupic/regions/CLAClassifierRegion.py
```diff
@@ -263,6 +355,30 @@ def customCompute(self, recordNum, patternNZ, classification):
...
+    return self._outputValues[outputName]
+
+
+  def getOutputElementCount(self, name):
+    """Returns the width of dataOut."""
+   
+    if name == "classificationResults":
+      return len(self.stepsList)
+    elif name == "probabilities":
+      if not self.numCategories:
```
scottpurdy: when would this be true? Is this for backwards compatibility? In that case, it should be updated in the deserialization to upgrade old version. Same for "actualValues" as well

scottpurdy: If these are true for any model when the parameter is not specified then this is probably fine

marionleborgne: Mm. Good point. The exceptions would never be raised since numCatgories is now a required param

scottpurdy:
Will this code work with old saved models? And what happens if people call `customCompute` after `run`, like in most uses currently? Will it double compute? Can we make it so `customCompute` is a noop?

Please also fix lines that are over 80 characters

nupic/algorithms/CLAClassifier.py
```diff
@@ -382,6 +381,11 @@ def compute(self, recordNum, patternNZ, classification, learn, infer):
     # Store pattern in our history
     self._patternNZHistory.append((self._learnIteration, patternNZ))
 
+    # To allow multi-class classification, we need to be able to run leaning
```
oxtopus: typo: "learning"

nupic/regions/CLAClassifierRegion.py
```diff
@@ -224,20 +278,58 @@ def compute(self, inputs, outputs):
...
 
-    pass
+    # An input can potentially belong to multiple categories. 
+    # If a category value is < 0, it means that the input does not belong to that category.
+    categories = []
+    for category in inputs['categoryIn']:
+      # if a category value <0, then it means 
+      # the input record does not belong to that category.
+      if category >= 0:
+        categories.append(category)
```
oxtopus: Could be rewritten as a list comprehension:

`categories = [category for category in inputs["categoryIn"] if category >= 0]`

marionleborgne:
@scottpurdy if people use customCompute after run, it will double compute. I think we should make customCompute a noop. but they would still have to make sure their classification network is compatible with this new version of the CLAClassifier, as it now requires a `numCategories` param

marionleborgne:
@scottpurdy @oxtopus thanks for the comments. I have made the changes. let me know if you have any additional feedback. There is only this question about what to do with the `customCompute` function that remains. otherwise it seems like this fixes the issue.

scottpurdy:
Can you give the numCategories parameter a default of 1? And make sure old checkpoints will get updated accordingly?

nupic/regions/CLAClassifierRegion.py
```diff
@@ -256,16 +340,36 @@ def customCompute(self, recordNum, patternNZ, classification):
...
+    elif name == "probabilities":
+      return len(self.stepsList) * self.numCategories
+    elif name == "actualValues":
+      return self.numCategories
+    else:
+      raise _UnknownOutput("Unknown output {}.".format(name))
 
-if __name__=='__main__':
+
+if __name__ == '__main__':
```
scottpurdy: double quotes

scottpurdy:
Our coding guidelines require three blank lines before top level class and function definitions and two before methods.

nupic/algorithms/CLAClassifier.py
```diff
@@ -131,13 +128,14 @@ def store(self, iteration, bucketIdx):
...
-                                  (iteration - self._lastTotalUpdate))
+             (iteration - self._lastTotalUpdate))
     if denom > 0:
       dcNew = dc + (self._classifier.alpha / denom)
 
     # This is to prevent errors associated with inf rescale if too large
     if denom == 0 or dcNew > DUTY_CYCLE_UPDATE_INTERVAL:
-      exp =  (1.0 - self._classifier.alpha) ** (iteration-self._lastTotalUpdate)
+      exp = (1.0 - self._classifier.alpha) ** (
+      iteration - self._lastTotalUpdate)
```
scottpurdy: fix indentation

nupic/algorithms/CLAClassifier.py
```diff
@@ -553,13 +548,15 @@ def read(cls, proto):
     classifier.alpha = proto.alpha
     classifier.actValueAlpha = proto.actValueAlpha
     classifier._learnIteration = proto.learnIteration
-    classifier._recordNumMinusLearnIteration = proto.recordNumMinusLearnIteration
+    classifier._recordNumMinusLearnIteration = \
```
scottpurdy: Don't use backslashes. Use parens for implicit line continuations. Details here:
https://www.python.org/dev/peps/pep-0008/#maximum-line-length

nupic/regions/CLAClassifierRegion.py
```diff
@@ -29,6 +29,13 @@
 from nupic.algorithms.cla_classifier_factory import CLAClassifierFactory
 
 
+class _NumCatgoriesNotSpecified(Exception):
```
scottpurdy: this is spelled wrong and also not used so remove it

nupic/regions/CLAClassifierRegion.py
```diff
@@ -256,16 +340,36 @@ def customCompute(self, recordNum, patternNZ, classification):
...
+    """Returns the width of dataOut."""
 
+    if name == "categoriesOut":
+      return len(self.stepsList)
+    elif name == "probabilities":
+      return len(self.stepsList) * self.numCategories
+    elif name == "actualValues":
+      return self.numCategories
+    else:
+      raise _UnknownOutput("Unknown output {}.".format(name))
```
scottpurdy: I'd just use `ValueError` rather than defining a new exception type. The purpose of defining your own is to give fine grained ability to catch exceptions but `ValueError` is plenty specific. Describe the exception thrown in the doc string

nupic/regions/CLAClassifierRegion.py
```diff
@@ -256,16 +340,36 @@ def customCompute(self, recordNum, patternNZ, classification):
...
 
+  def getOutputValues(self, outputName):
+    """
+    Return the dictionary of output values. Note that these are normal Python
+    lists, rather than numpy arrays. This is to support lists with mixed scalars
+    and strings, as in the case of records with categorical variables
+    """
+    return self._outputValues[outputName]
+
+  def getOutputElementCount(self, name):
```
scottpurdy: Here you use `name` but in the function above you use `outputName`. If these refer to the same thing then choose one and be consistent.

nupic/regions/CLAClassifierRegion.py
```diff
@@ -214,30 +258,70 @@ def setParameter(self, name, index, value):
...
+                                  learn=self.learningMode,
+                                  infer=False)
+
+    outputs['actualValues'] = clResults["actualValues"]
+    for step in self.stepsList:
+      stepIndex = self.stepsList.index(step)
+      categoryOut = clResults["actualValues"][clResults[step].argmax()]
+      outputs['categoriesOut'][stepIndex] = categoryOut
+
+      # Flaten the rest of the output. For example:
```
scottpurdy: I like this comment. It kind of suggests that yo ucould move this logic into a function though which would make it more testable.

marionleborgne:
@scottpurdy i'm done addressing the PR feedback. On Monday, I'd like to discuss your idea of setting the default value of `numCategories` set to 1. I am not sure that would work. Have a good weekend!

nupic/regions/CLAClassifierRegion.py
scottpurdy: The comment also has a typo s/Flaten/Flatten

scottpurdy:
@marionleborgne - Please update the description and include an explanation of the plan for backwards compatibility. Plus this will support multi-step prediction still, right?

And it looks like the customCompute still makes a call to `compute` which will now be a duplicate call, right?

marionleborgne:
@scottpurdy I just made the changes to support backward compatibility. Here are the changes:
* default value of `numCategories` is set to `1000`, so support a potentially large number of output categories (i.e. inference results of the CLAClassifier). Right now the parameter `numCategories` is required. It's unclear to me if it should be or not. If it's required, it's probably going to impact the ability to run old models checkpoints, right? It might be better to have this param as optional for now.
* updated `customCompute()` so that it does only inference, and not learning. I have extracted the `infer`  part in `CLAClassfier.py` and I am calling this new `infer` function in `customCompute` now. That should prevent the model from learning two times if `customCompute` is called while `compute` has also be called wile running the model. `customCompute` will keep outputting the inference results output it was outputting before. 

I think this should fix the backward compatibility issue. Let me know what you think of that.

scottpurdy:
This code looks good. Can you create a model checkpoint from the previous codebase and validate that you can load and get the same results from both old and new code?

marionleborgne:
yes. do you ming pointing me to an example / doc of model checkpointing with the network API? thanks!

scottpurdy:
@marionleborgne - You could use the example linked below. It uses the CLAModel, which internally has a Network instance. If you change the example to save the model at then end (`model.save("dir/to/save/in")`) then you can create a seprate script to load it:

```
from nupic.frameworks.opf.modelfactory import ModelFactory

model = ModelFactory.loadFromCheckpoint("...")
result = model.run({"field1": data, ...})
# Check result matches what it was in previous code
```

I might have some imports or function names wrong so let me know if you have any trouble getting that to work.

Here is what you can modify to save the model:
https://github.com/numenta/nupic/blob/master/examples/opf/clients/hotgym/simple/hotgym.py

marionleborgne:
@scottpurdy So I think this backward compatibility issue works now.

I performed the following test:
1. created and trained a model on the hotgym data, with the old CLA classifier implementation in `master`. Saved the model.
2. Changed to `fix-nupic-issue-2340 ` branch with the new CLA classifier implementation. Loaded the model checkpoint, and and ran the model.

The test failed at step 2, at first. So I added a check in `getOutputElementCount` https://github.com/numenta/nupic/commit/b19098a1b965d71ebf5bf0bd0797c6896be8929c#diff-cbc2aaee76e33738f3c0def55d30b4b4R370

This solved the backward compatibility issue that I had with the old model. After that the test passed: the old model checkpoint was loaded and could be run even with the new implementation of the CLAclassifier.

marionleborgne:
There is one remaining problem though: all works fine if I use the `'implementation': 'py'` in the classifier model params. But if I set the implementation to `cpp` then I can run the model created from the old model checkpoint: 

```
  File "/Users/mleborgne/_git/nupic/examples/opf/clients/hotgym/simple/hotgym.py", line 106, in <module>
    runHotgym()
  File "/Users/mleborgne/_git/nupic/examples/opf/clients/hotgym/simple/hotgym.py", line 89, in runHotgym
    result = model.run(modelInput)
  File "/Users/mleborgne/_git/nupic/nupic/frameworks/opf/clamodel.py", line 395, in run
  File "/Users/mleborgne/_git/nupic/nupic/frameworks/opf/clamodel.py", line 533, in _multiStepCompute
    rawInput=rawInput)
  File "/Users/mleborgne/_git/nupic/nupic/frameworks/opf/clamodel.py", line 767, in _handleCLAClassifierMultiStep
    classification=classificationIn)
  File "/Users/mleborgne/_git/nupic/nupic/regions/CLAClassifierRegion.py", line 354, in customCompute
    return self._claClassifier.infer(patternNZ, classification)
AttributeError: 'FastCLAClassifier' object has no attribute 'infer'
```

`customCompute` is calling that `infer` method, that I added to separate learning from inference - in order to no-double compute when `customCompute` is called. I am not sure how to propagate that change (the new `infer` method) to the `cpp` implementation. Any idea @scottpurdy ?

marionleborgne:
@scottpurdy I ended up adding an `infer` method in `nupic.bindings.algorithms.py`, in the class `FastCLAClassifier`:

```
    def infer(self, recordNum, patternNZ, classification):
      learn = False
      infer = True
      return self.compute(recordNum, patternNZ, classification, learn, infer)
```

This seems to fix the issue. 

However, I see that `nupic/bindings/*.py` is the `.gitignore` so I did commit that change yet. Let me know how you would like to proceed.

marionleborgne:
@scottpurdy OK so I tested it with the hotgym example, and it works. I am getting consistent accuracy results when I am loading an old model checkpoint with the new CLAClassifier implementation (the one with a working `compute()` method)

marionleborgne:
@scottpurdy provided you are happy with the code style, then I think this PR is ready to be merged :grinning:

nupic/regions/CLAClassifierRegion.py
```diff
@@ -223,26 +271,84 @@ def compute(self, inputs, outputs):
...
-    fit the standard "vector of reals" used by the engine. Instead, call
-    the customCompute() method directly
     """
 
-    pass
-
+    # This flag helps to prevent double-computation, in case the deprecated 
+    # customCompute() method is being called in addition to compute() called 
+    # when network.run() is called
+    if not self._computeFlag:
```
scottpurdy: no need for conditional, just set it to true

scottpurdy:
Could you include some simple tests for the two different use cases of the classifier? There may already be tests somewhere.

Done with review

marionleborgne:
OK. does that work if I commit a old pickled model + expected results along with the test? I'm going to load that model and check that the results i'm getting with the new impl are consistent with the old model results.

marionleborgne:
@scottpurdy The test is passing.

marionleborgne:
@BoltzmannBrain FYI

scottpurdy:
@marionleborgne - this needs to say `fixes #2340` or similar somewhere in the description to pass the fixes issue validator

marionleborgne:
fixes #2340

PR #2420, opened 2015-08-11
Split out the Record-dict encoding logic from RecordStreamIface into a separate class.

vitaly-krugl:
NOTE: this is a preview PR; it doesn't include tests, yet, and no issue has been filed.

Split out the Record-dict encoding logic from `RecordStreamIface` into a separate class so that it may be used by clamodel apps that don't need the rest of `RecordStreamIface`'s abstract-base-class baggage.

Pass fieldmetainfo list to RecordStreamIface instead of querying the derived class every time these static values are needed.
Remove unnecessary implementations of several methods from derived StreamReader class, since an equivalent implementation already exists in `RecordStreamIface`.

Replace use of hard-coded special attribute values with constants from `FieldMetaSpecial`.

Resolved most pylint issues in affected modules.

Fixed getFields validation tests: assert comparison of sequences instead of expecting the values wrapped in a list type explicitly.

### DIFF
--- a/nupic/data/file_record_stream.py
+++ b/nupic/data/file_record_stream.py
@@ -63,10 +63,9 @@ contents using a for loop:
 
 import os
 import csv
-import copy
 import json
 
-from nupic.data.fieldmeta import FieldMetaInfo
+from nupic.data.fieldmeta import FieldMetaInfo, FieldMetaSpecial
 from nupic.data import SENTINEL_VALUE_FOR_MISSING_DATA
 from nupic.data.record_stream import RecordStreamIface
 from nupic.data.utils import (intOrNone, floatOrNone, parseBool, parseTimestamp,
@@ -78,7 +77,7 @@ from nupic.data.utils import (intOrNone, floatOrNone, parseBool, parseTimestamp,
 class FileRecordStream(RecordStreamIface):
   """ CSV file based RecordStream implementation
   """
-  
+
   # Private: number of header rows (field names, types, special)
   _NUM_HEADER_ROWS = 3
 
@@ -91,7 +90,7 @@ class FileRecordStream(RecordStreamIface):
 
   def __init__(self, streamID, write=False, fields=None, missingValues=None,
                bookmark=None, includeMS=True, firstRecord=None):
-    """ 
+    """
     streamID:
         CSV file name, input or output
     write:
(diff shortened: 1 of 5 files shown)


vitaly-krugl:
NOTE for reviewers: my IDE eliminated a bunch of trailing white spaces in a couple of files. Use this URL to ignore the trailing whtespace changes in the diffs: https://github.com/numenta/nupic/pull/2420/files?w=1

nupic/data/file_record_stream.py
```diff
@@ -579,31 +595,15 @@ def isCompleted(self):
...
+
 
   def setCompleted(self, completed=True):
     """ Marks the stream completed (True or False)
     """
     # CSV file is always considered completed, nothing to do
     return
-  
-  
-  def getFieldNames(self):
```
vitaly-krugl: Removed in favor of the ones in the base class that provide equivalent functionality.

vitaly-krugl:
CC @scottpurdy

scottpurdy:
@vitaly-krugl - High level looks good but I need to review more thoroughly. One question first: is adding the L special field necessary for this change or is it just another thing that you think should be added?

nupic/data/record_stream.py
```diff
@@ -25,182 +25,272 @@
...
-              'months': number of months in aggregation period
-              'seconds': number of seconds in aggregation period (as a float)
+
+class RecordStreamIface(object):
+  """This is the interface for the record input/output storage classes."""
+
+  __metaclass__ = ABCMeta
+
+
+  def __init__(self, fields, aggregationPeriod=None):
```
vitaly-krugl: Note that the args in the `RecordStreamIface` constitute a change in its interface versus `def __init__(self)`. The legacy code was a bit backwards in that it depended on abstract methods to query this **static** information on demand each time it was needed (and it needed these values) instead of simply getting it via constructor as is typical.

vitaly-krugl:
@scottpurdy: regarding
> High level looks good but I need to review more thoroughly. One question first: is adding the L special field necessary for this change or is it just another thing that you think should be added?

I noticed that `RecordStream` and `FileRecordStream` are hard-coding the attributes instead of making use of making use of the constants in `FieldMetaSpecial`. Support for `L` (learning attribute) had been hard-coded in those classes without updating updating `FieldMetaSpecial`. Circumventing the intended classes constants is a bad situation, so I set good precedent by updating `FieldMetaSpecial` to include the new attribute and modifying the code to use `FieldMetaSpecial` constants instead of hard-coded values. The new `RecordDictEncoder` relies on `FieldMetaSpecial`, so I also felt that it was in-scope to make that change.

**NOTE** that a similar problem exists with the use of hard-coded types instead of `FieldMetaType`-based constants in `FileRecordStream` and elsewhere. That one had diverged even more than `FieldMetaSpecial`, including new types and support for both upper-case and lower-case variants. So, I ended up leaving the field type code alone, as it was more involved to rein in and completely out of scope of the change that I needed to make.

vitaly-krugl:
Closing this PR. Will created a new PR with a more limited scope of changes.

PR #2422, opened 2015-08-11
moved sortedJSONdumps out of nupic into swarming

### LINKED ISSUE #2421: Migrate Sorted JSON Dump function
The 'sortedJSONDumpS' function is only used within swarming, so it should be moved out of nupic and into the appropriate dir.

mihail911:
Fixes #2421 
@scottpurdy


scottpurdy:
The issue says the function is only used by swarming so should the original be deleted?

mihail911:
For some reason my editor showed that the original had been deleted. I will double-check the commit changes more closely next time. @scottpurdy

scottpurdy:
:+1:

PR #2429, opened 2015-08-13
remove swarming matchPatterns dependency on NuPIC

### LINKED ISSUE #2428: Resolve matchPatterns Swarming Dependency
Remove the swarming dependency on 'matchPatterns' within NuPIC.

mihail911:
Fixes #2428 
@scottpurdy @oxtopus

### DIFF
--- a/nupic/swarming/hypersearch/utils.py
+++ b/nupic/swarming/hypersearch/utils.py
@@ -39,11 +39,6 @@ import validictory
 from nupic.database.ClientJobsDAO import (
     ClientJobsDAO, InvalidConnectionException)
 
-# TODO: Note the function 'rUpdate' is also duplicated in the
-# nupic.data.dictutils module -- we will eventually want to change this
-# TODO: 'ValidationError', 'validate', 'loadJSONValueFromFile' duplicated in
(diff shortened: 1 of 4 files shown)


nupic/swarming/hypersearch/utils.py
```diff
@@ -753,3 +754,25 @@ def sortedJSONDumpS(obj):
 
   else:
     return json.dumps(obj)
+
+
+
+def matchPatterns(patterns, keys):
```
oxtopus: Is this duplicated from somewhere else?  Please include a TODO or something in the docstring that indicates that it was copied from nupic so that it may be refactored.

Otherwise, I have some feedback for how this ought to change.  For example, I don't believe the docstring doesn't matches the doxygen format, or sphinx (commonly used for documenting python code).  The fact that it returns `None` or a `list()` (which may be empty) doesn't seem right.

oxtopus: I see it above, now.  I think it would be better if the `TODO` were closer to the actual function.

mihail911:
Moved TODO closer to function as requested @oxtopus

PR #2432, opened 2015-08-13
Input Model Record Encoder

vitaly-krugl:
1. Broke out model record encoding logic from RecordStreamIface into ModelRecordEncoder class.
2. Removed the 'address' field type hack that was left over as the legacy from Precog prep.
3. Added missing attributes and validation checkers in nupic.data.fieldmeta.
4. Converted stream modules to use nupic.data.fieldmeta types/constants instead of hard-coding those values.
5. Implemented unit tests for ModelRecordEncoder.
6. Implemented unit tests for RecordStreamIface.getNextRecordDict and RecordStreamIface.rewind
7. Implemented unit tests for ValueError exception from FieldMetaInfo.
8. Implemented unit tests for FieldMetaSpecial.isValid and FieldMetaType.isValid
9. Fixed a bunch of pylint findings in the affected modules.

### DIFF
--- a/nupic/data/fieldmeta.py
+++ b/nupic/data/fieldmeta.py
@@ -1,6 +1,6 @@
 # ----------------------------------------------------------------------
 # Numenta Platform for Intelligent Computing (NuPIC)
-# Copyright (C) 2013, Numenta, Inc.  Unless you have an agreement
+# Copyright (C) 2013-15, Numenta, Inc.  Unless you have an agreement
 # with Numenta, Inc., for a separate license for this software code, the
 # following terms and conditions apply:
 #
@@ -30,6 +30,8 @@ from collections import namedtuple
 
 FieldMetaInfoBase = namedtuple('FieldMetaInfoBase', ['name', 'type', 'special'])
 
+
+
 class FieldMetaInfo(FieldMetaInfoBase):
   """
   This class acts as a container of meta-data for a single field (column) of
@@ -57,6 +59,27 @@ class FieldMetaInfo(FieldMetaInfoBase):
   3.
   """
 
+
+  def __init__(self,
+               name,
+               type,  # pylint: disable=W0622
+               special):
+    """
+    :param str name: field name
+    :param str type: one of the values from FieldMetaType
+    :param str special: one of the values from FieldMetaSpecial
+    :raises ValueError: if type or special arg values are invalid
+    """
+
+    if not FieldMetaType.isValid(type):
+      raise ValueError('Unexpected field type %r' % (type,))
+
+    if not FieldMetaSpecial.isValid(special):
+      raise ValueError('Unexpected field special attribute %r' % (special,))
+
+    super(FieldMetaInfo, self).__init__(name, type, special)
+
+
   @staticmethod
   def createFromFileFieldElement(fieldInfoTuple):
     """ Creates a FieldMetaInfo instance from an element of the File.fields list
@@ -76,7 +99,7 @@ class FieldMetaInfo(FieldMetaInfoBase):
     Returns:  A list of FieldMetaInfo elements corresponding to the given
               'fields' list.
     """
-    return map(lambda x: cls.createFromFileFieldElement(x), fields)
+    return [cls.createFromFileFieldElement(f) for f in fields]
 
 
 
@@ -90,6 +113,22 @@ class FieldMetaType(object):
   float = 'float'
   boolean = 'bool'
   list = 'list'
+  sdr = 'sdr'  # sparse distributed representation
+
+  _ALL = (string, datetime, integer, float, boolean, list, sdr)
+
+
+  @classmethod
+  def isValid(cls, fieldDataType):
+    """Check a candidate value whether it's one of the valid field data types
+
+    :param str fieldDataType: candidate field data type
+    :returns: True if the candidate value is a legitimate field data type value;
+      False if not
+    :rtype: bool
+    """
+    return fieldDataType in cls._ALL
+
 
 
 class FieldMetaSpecial(object):
@@ -101,3 +140,18 @@ class FieldMetaSpecial(object):
   sequence = 'S'
   timestamp = 'T'
   category = 'C'
+  learning = 'L'
+
+  _ALL = (none, reset, sequence, timestamp, category, learning,)
+
+
+  @classmethod
+  def isValid(cls, attr):
+    """Check a candidate value whether it's one of the valid attributes
+
+    :param str attr: candidate value
+    :returns: True if the candidate value is a legitimate "special" field
+      attribute; False if not
+    :rtype: bool
+    """
+    return attr in cls._ALL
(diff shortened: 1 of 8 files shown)


vitaly-krugl:
@scottpurdy: please take a look

nupic/data/stream_reader.py
```diff
@@ -472,38 +475,6 @@ def getBookmark(self):
     return self._aggBookmark
 
 
-  def getResetFieldIdx(self):
```
vitaly-krugl: These get*FieldIdx() methods are already implemented in the base class.

vitaly-krugl:
Fixes #2433

nupic/data/fieldmeta.py
```diff
@@ -97,7 +136,22 @@ class FieldMetaSpecial(object):
   Public values for the "special" field attribute
   """
   none = ''
-  reset = 'R'
-  sequence = 'S'
+  reset = 'R'     # designates the field as a reset flag
```
scottpurdy: these comments don't seem to clarfiy anything

nupic/data/fieldmeta.py
```diff
@@ -97,7 +136,22 @@ class FieldMetaSpecial(object):
...
+  sequence = 'S'  # designates the field as a sequenceId
   timestamp = 'T'
   category = 'C'
+  learning = 'L'
+
+  _ALL = (none, reset, sequence, timestamp, category, learning,)
+
+
+  @classmethod
+  def isValid(cls, attr):
```
scottpurdy: these functions are nice

nupic/data/stream_reader.py
```diff
@@ -24,20 +24,15 @@
 import logging
 import tempfile
 
-from pkg_resources import resource_filename
+import pkg_resources
 
 from nupic.data.aggregator import Aggregator
-from nupic.data.fieldmeta import FieldMetaInfo
+from nupic.data.fieldmeta import FieldMetaInfo, FieldMetaType,FieldMetaSpecial
```
scottpurdy: missing space after last comma

tests/unit/nupic/data/record_stream_test.py
```diff
@@ -0,0 +1,502 @@
+#!/usr/bin/env python
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2013, Numenta, Inc.  Unless you have an agreement
```
scottpurdy: update years in headers

scottpurdy:
at some point we should update `CLAModel.run` documentation to make the input format more clear. This PR looks good though

nupic/data/fieldmeta.py
```diff
@@ -57,6 +59,27 @@ class FieldMetaInfo(FieldMetaInfoBase):
   3.
   """
 
+
+  def __init__(self,
+               name,
+               type,  # pylint: disable=W0622
```
oxtopus: `type` is built-in.

nupic/data/fieldmeta.py
```diff
@@ -90,6 +113,22 @@ class FieldMetaType(object):
...
+  sdr = 'sdr'  # sparse distributed representation
+
+  _ALL = (string, datetime, integer, float, boolean, list, sdr)
+
+
+  @classmethod
+  def isValid(cls, fieldDataType):
+    """Check a candidate value whether it's one of the valid field data types
+
+    :param fieldDataType attr: candidate field data type
```
oxtopus: Something fishy about this doc format.

nupic/data/record_stream.py
```diff
@@ -25,178 +25,271 @@
...
+def _getFieldIndexBySpecial(fields, special):
+  """ Return index of the field matching the field meta special value.
+  :param fields: sequence of nupic.data.fieldmeta.FieldMetaInfo objects
+    representing the fields of a stream
+  :param special: one of the special field attribute values from
+    nupic.data.fieldmeta.FieldMetaSpecial
+  :returns: first zero-based index of the field tagged with the target field
+    meta special attribute; None if no such field
+  """
+  for i, field in enumerate(fields):
```
oxtopus: Maybe a good idea to check that `fields` is in fact a sequence that can be indexed (like a `list` or `tuple`).  A subtle bug could be introduced if someone accidentally passes a `dict` or `set`.

nupic/data/record_stream.py
```diff
@@ -25,178 +25,271 @@
...
+    hasReset = self._resetFieldIndex is not None
+    hasSequenceId = self._sequenceFieldIndex is not None
     if hasReset and not hasSequenceId:
       # Reset only
       if result['_reset']:
         try:
           self._sequenceId += 1
         except:
-          import pdb; pdb.set_trace() 
+          import pdb; pdb.set_trace()
```
oxtopus: :confounded: Let's get rid of this!

nupic/data/fieldmeta.py
vitaly-krugl: @oxtopus: unfortunately, the `type` arg was used in the legacy implementation, and changing it would be out of scope for this task. Take a look at the base class definition above it `FieldMetaInfoBase = namedtuple('FieldMetaInfoBase', ['name', 'type', 'special'])`. This may have been done by me (?) as I was learning the pythonic ways and before I knew better.

nupic/data/fieldmeta.py
vitaly-krugl: :) you're right.

nupic/data/record_stream.py
vitaly-krugl: @oxtopus: What happened to duck-typing, anyway? :) This is a private method. There are multiple places where the fields are passed and used, so this is probably not the ideal place for it. Public methods are usually where we might have checks like that.

nupic/data/record_stream.py
vitaly-krugl: Might as well get rid of it now. No idea why we have this `import pdb; pdb.set_trace()` business here, anyway.

nupic/data/record_stream.py
oxtopus: This is a private method, but there are public methods that use this and blindly pass the value -- `ModelRecordEncoder.__init__()` for example.  Regardless, someone may intentionally make a change that results in this function being passed a different type, regardless of its visibility.

vitaly-krugl: @oxtopus, I added a check to make sure that `fields` arg evaluates to true, which covers the non-empty case. I leave the rest up to duck typing. I am hesitant about the check that you suggested, as it is conventional in python to use the "duck typing" style. By explicitly checking for `list` and `tuple`, we're reducing the generality of the implementation, since anything that implements the sequence protocol should be an acceptable container in this case. Second, you would need a pretty unlikely confluence of getting the wrong container and having "special" as the attribute of each of its elements. I don't believe that we have a problem here that needs to be solved.

PR #2435, opened 2015-08-13
Create wheel for nupic.bindings during installation

### LINKED ISSUE #2413: Master build deployment failure, missing "nupic.bindings"
See: https://travis-ci.org/numenta/nupic/jobs/75138342#L6712

```
Downloading/unpacking nupic.bindings==0.1.0 (from nupic==0.2.7.dev0)
  Could not find any downloads that satisfy the requirement nupic.bindings==0.1.0 (from nupic==0.2.7.dev0)
Cleaning up...
No distributions at all found for nupic.bindings==0.1.0 (from nupic==0.2.7.dev0)
Storing complete log in /home/travis/.pip/pip.log
The command ". ./ci/travis/before_deploy.sh" failed and exited with 1 during .
```

CC @akhilaananthram @scottpurdy

akhilaananthram:
This allows us to deploy a nupic wheel and the wheels for the dependencies.

Fixes #2413

@scottpurdy @rhyolight

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -385,7 +385,6 @@ if __name__ == "__main__":
     copyProtoFiles(nupicCoreReleaseDir)
 
     requirements, dependencies = findRequirements(nupicCoreReleaseDir)
-    print dependencies
     setup(
       name="nupic",
       version=getVersion(),
--- a/setup.py
+++ b/setup.py
@@ -9,6 +9,7 @@ import tarfile
 import urllib2
 
 from setuptools import setup, find_packages, Extension
+from wheel import egg2wheel
 
 """
 This file builds and installs the NuPIC binaries.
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -9,6 +9,7 @@
 import urllib2
 
 from setuptools import setup, find_packages, Extension
+from wheel import egg2wheel
```
scottpurdy: Can we import this where it is used? Otherwise if the user doesn't have wheel installed and just wants to install NuPIC then we will get an error.

akhilaananthram:
@scottpurdy @oxtopus 
I split "pip wheel . " into multiple commands to make the wheels that we need.  I also added a command to make the nupic egg.

scottpurdy:
:+1:

PR #2439, opened 2015-08-14
Fixed fast_temporal_memory cellsForColumn calculation

### LINKED ISSUE #2440: Fixed fast_temporal_memory cellsForColumn calculation

jaredweiss:
@scottpurdy @chetan51 

fixes #2440 

This code is failing for me w/ the `getCellIndex` call. Column is an `numpy.int64` not a nupic `Cell`, so getCellIndex won't actually work on this. The docstring describes column as an int anyway, so the `fast_temporal_memory` implementation shouldn't try to use `getCellIndex` which expects a `Cell`.


chetan51:
:+1:

PR #2445, opened 2015-08-18
Formalize indent-after-paren=2 in pylintrc per nupic python coding guidelines of 2 spaces per indent.

vitaly-krugl:


vitaly-krugl:
Please review

CC @scottpurdy

PR #2469, opened 2015-08-21
Oo

oxtopus:
Creating this PR so I can give meaningful feedback on the changes in the OO proposal.  Not meant to merge, but have a way to comment, and discuss specific lines of code.


nupic/research/spatial_pooler.py
```diff
@@ -50,8 +51,8 @@ class SpatialPooler(object):
   """
 
   def __init__(self,
+               layer=Layer(columnDimensions=(64,64)),
```
oxtopus: This is a dangerous convention.  You should only use immutable types as defaults in function definitions.  Here's a concrete example to demonstrate why:

```python
>>> class MutableObject(object):
...   def __init__(self, foo=None):
...     self.foo = foo
... 
>>> class Adapter(object):
...   def __init__(self, mutableObj=MutableObject(1)):
...     self.mutableObj = mutableObj
... 
>>> obj1 = Adapter()
>>> obj2 = Adapter()
>>> 
>>> obj1.mutableObj.foo = "Foo"
>>> print obj2.mutableObj.foo
Foo
>>> obj2.mutableObj.foo = "Bar"
>>> print obj1.mutableObj.foo
Bar
```

In other words, `obj1.mutableObj` and `obj2.mutableObj` are the same instance of `MutableObject`, despite being attributes of different instances of `Adapter`.  This is because the object is created at time of definition, rather than instantiation.

Using my example, better to have something like the following to handle default values:

```python
class Adapter(object):
  def __init__(self, mutableObj=None):
    self.mutableObj = mutableObj or MutableObject(1)
```

nupic/network/synapse.py
```diff
@@ -0,0 +1,38 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+
+
+class Synapse(object):
```
oxtopus: With thousands of columns, several cells per column, and thousands of synapses per cell, there is potential for all of the instances of `Synapse` to occupy way more memory than is necessary than if a single contiguous range of memory, or another efficient data structure like a sparse matrix.

nupic/research/spatial_pooler.py
```diff
@@ -997,59 +959,97 @@ def _raisePermanenceToThreshold(self, perm, mask):
...
+      "to the input size. [len(synapses) < self._stimulusThreshold]")
 
-    numpy.clip(perm, self._synPermMin, self._synPermMax, out=perm)
+    numConnected = sum(1 for synapse in synapses if synapse.permanence >= self._synPermConnected)
     while True:
-      numConnected = numpy.nonzero(perm > self._synPermConnected)[0].size
       if numConnected >= self._stimulusThreshold:
         return
-      perm[mask] += self._synPermBelowStimulusInc
+      numConnected = 0
```
oxtopus: Why is `numConnected` being re-calculated here and below, compared to above?

nupic/research/temporal_memory.py
```diff
@@ -53,30 +53,24 @@ def __init__(self,
...
-    # Error checking
-    if not len(columnDimensions):
-      raise ValueError("Number of column dimensions must be greater than 0")
-
-    if not cellsPerColumn > 0:
-      raise ValueError("Number of cells per column must be greater than 0")
 
     # TODO: Validate all parameters (and add validation tests)
+    if layer is None:
+      raise ValueError("Layer must be an instance of Layer class")
```
oxtopus: Message is "Layer must be an instance of Layer class", but not asserting on type.  It's possible for `layer` to be something else (not a `Layer`) and pass this check.

nupic/network/cell.py
```diff
@@ -0,0 +1,56 @@
...
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+from nupic.network.segment import Segment
+
+
+
+class Cell(object):
```
oxtopus: Classes like this should assign `__slots__` for improved memory.

nupic/network/synapse.py
cogmission: Austin, I understand your point and share your interest in this concern. Please see the email I forwarded you entitled "Not to beat a dead horse..." 

The current data structure implemented creates a synapse on every mutation which is way more significant than the implications of using a reusable and mutable Synapse object?

oxtopus: Expanding on my contrived example, consider the following python code:

```python
import sys



ITERATIONS = 10000000



class MutableObject(object):
  def __init__(self, foo=None):
    self.foo = foo



class SlottedMutableObject(MutableObject):
  __slots__ = ("foo",)



def create():
  for index in xrange(ITERATIONS):
    obj = MutableObject("bar")



def mutate(objects = {}):
  for index in xrange(ITERATIONS):
    obj = objects.setdefault(index, MutableObject("bar"))



def mutateSlotted(objects = {}):
  for index in xrange(ITERATIONS):
    obj = objects.setdefault(index, SlottedMutableObject("bar"))



if __name__ == "__main__":
  selection = sys.argv[1]

  if selection == "create":
    create()
  elif selection == "mutate":
    mutate()
  elif selection == "mutate-slotted":
    mutateSlotted()
```

`create()` creates 10,000,000 `MutableObject` instances, `mutate()` instantiates and caches them, `mutateSlotted()` does the same with a subclass that defines `__slots__`.  When I run each of those with `cProfile`, here's what I get:

```
Austins-MacBook-Pro:Desktop amarshall$ python -m cProfile mutant.py create
         10000005 function calls in 3.851 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    3.851    3.851 mutant.py:1(<module>)
 10000000    1.145    0.000    1.145    0.000 mutant.py:10(__init__)
        1    0.000    0.000    0.000    0.000 mutant.py:15(SlottedMutableObject)
        1    2.706    2.706    3.851    3.851 mutant.py:20(create)
        1    0.000    0.000    0.000    0.000 mutant.py:9(MutableObject)
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}


Austins-MacBook-Pro:Desktop amarshall$ python -m cProfile mutant.py mutate
         20000005 function calls in 15.278 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000   15.278   15.278 mutant.py:1(<module>)
 10000000    5.746    0.000    5.746    0.000 mutant.py:10(__init__)
        1    0.000    0.000    0.000    0.000 mutant.py:15(SlottedMutableObject)
        1    8.372    8.372   15.278   15.278 mutant.py:26(mutate)
        1    0.000    0.000    0.000    0.000 mutant.py:9(MutableObject)
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
 10000000    1.160    0.000    1.160    0.000 {method 'setdefault' of 'dict' objects}


Austins-MacBook-Pro:Desktop amarshall$ python -m cProfile mutant.py mutate-slotted
         20000005 function calls in 10.649 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000   10.649   10.649 mutant.py:1(<module>)
 10000000    0.986    0.000    0.986    0.000 mutant.py:10(__init__)
        1    0.000    0.000    0.000    0.000 mutant.py:15(SlottedMutableObject)
        1    8.568    8.568   10.649   10.649 mutant.py:32(mutateSlotted)
        1    0.000    0.000    0.000    0.000 mutant.py:9(MutableObject)
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
 10000000    1.095    0.000    1.095    0.000 {method 'setdefault' of 'dict' objects}
```

Notice that `create()` is ~5x faster than `mutate()`.  That's just for the _creation_ of the objects.  It is faster to do a lookup in a `dict` than it is to create an object, however, but there's memory overhead in retaining the `dict` in which the objects are cached and as the results above show, there's performance overhead in caching the objects in a dict.  When I add `print sys.getsizeof(objects)` to the end of `mutate()` and `mutateSlotted()` I get 402653464 bytes.

Are we willing to take the memory hit?  Probably not, but it would be cool if it were an option so that the user can make that choice according to their own usage pattern.

Something else to consider, any memory overhead that would need to be translated to a persisted checkpoint could potentially slow things down in a production setting, such as is the case in `htmengine` where models are checkpointed and loaded from disk regularly.

cogmission: 1. Why is it the first example does a create of 10M and the others do a mutate of 20M?
2. Also I think the following code might be rehashing or indexing which needs to be taken into account as opposed to merely mutating  permanence value?

``` Python
def updateSynapsePermanence(self, synapse, permanence):
    """
    Updates the permanence for a synapse.

    @param synapse    (int)   Synapse index
    @param permanence (float) New permanence
    """
    self._validatePermanence(permanence)

    data = self._synapses[synapse]
    newData = SynapseData(data.segment,   <--------------------   HERE
                          data.presynapticCell,
                          permanence)
    self._synapses[synapse] = newData

    # Update indexes
    self._synapsesForPresynapticCell[newData.presynapticCell][synapse] = newData
```

oxtopus: > Why is it the first example does a create of 10M and the others do a mutate of 20M?

In each case, I'm doing exactly 10M iterations.  However, cProfile is reporting on total number of function calls.  The `dict` lookup adds an extra function call, which contributes to the slowdown.

cogmission: Not sure how fast your fingers work :-) You might have answered this before I asked the follow up question about the Synapse in the actual code? 

I wonder what your example would do for an object accessed by:
```Python
self._synapses[synapse].permanence = xx.xx
```

rather than a dispatched method call like setdefault() in the above examples? I bet with 10M (as opposed to 20M) calls and assignments like the above it would be faster than create?

oxtopus: If I replace `obj = objects.setdefault(index, MutableObject("bar"))` with `objects[index] = MutableObject("bar")`, it shaves ~2 seconds off the total, bringing execution down to 13.204 seconds.  In a realistic caching scenario, you would need to do some sort of existence check anyway, hence the `setdefault()`.

```
Austins-MacBook-Pro:Desktop amarshall$ python -m cProfile mutant.py mutate
         10000005 function calls in 13.204 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000   13.204   13.204 mutant.py:1(<module>)
 10000000    6.351    0.000    6.351    0.000 mutant.py:10(__init__)
        1    0.000    0.000    0.000    0.000 mutant.py:15(SlottedMutableObject)
        1    6.853    6.853   13.203   13.203 mutant.py:26(mutate)
        1    0.000    0.000    0.000    0.000 mutant.py:9(MutableObject)
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
```

cogmission: In this case you are setting a numeric permanence value (a primitive data type which by default has to exist, no? At least it does in Java (no such thing as a null or empty int).  I guess Python is much different from other languages that array access is not much faster than an object creation paired with a method call that does null checking ;-)  weird...

Does this mean I should close that Issue too? #2479

oxtopus: In python and elsewhere there are similar abstractions, `list()` index access is generally faster than `dict()` key lookup, however, `list()` assumes a contiguous range whereas `dict()` can be random and sparse.  The trade-off is in memory.

You _can_ get better performance, as you suggest, by retaining and mutating an object, but you have to pre-allocate those objects AND cache them, which have a significant overhead.  Meanwhile in my example, in which the cache `dict` adds 400 Mb of memory overhead, if I needed to persist the object and reload it, that adds a substantial amount of overhead in writing to, and reading from disk.  The runtime benefits are a moot point if I have to load and save the cache object frequently, as we do in our own production settings.  The added memory footprint also limits concurrency.

It does seem redundant, however, to access `self._synapses[synapse]` and replace it with a new instance when it could possibly be changed in-place.  I'll comment in the issue with my thoughts.

cogmission: Austin, thanks for taking the time to explain the innards of Python to me - I appreciate it!

rhyolight: > innards of Python

python innards

david-ragazzi: >> With thousands of columns, several cells per column, and thousands of synapses per cell, there is potential for all of the instances of Synapse to occupy way more memory than is necessary than if a single contiguous range of memory, or another efficient data structure like a sparse matrix.

Once Python is to be readable and not for a production environment, I don't see problem in keep such change. Furthermore, TemporalMemory doesn't use `single contiguous range of memory, or another efficient data structure like a sparse matrix`.

nupic/research/spatial_pooler.py
david-ragazzi: >Why is numConnected being re-calculated here and below, compared to above?

Because according to current implementation, before permanences can be incremented with `self._synPermBelowStimulusInc` constant, first we have to check numConnected fits criteria. We could remove the first counting operation, but this would require we use a loop counter to check `synapse.permanence >= self._synPermConnected` only after 1 iteration:

```
i = 0
while True:
  if numConnected >= self._stimulusThreshold:
    return
  numConnected = 0
  for synapse in synapses:
    if i > 0:
      synapse.permanence += self._synPermBelowStimulusInc
    if synapse.permanence >= self._synPermConnected:
      numConnected += 1
  i += 1
```

nupic/network/cell.py
david-ragazzi: Yeah, this can be useful.

nupic/research/temporal_memory.py
david-ragazzi: Agreed.

oxtopus:
Closing until further notice.  Thanks, @david-ragazzi and @cogmission!

PR #2472, opened 2015-08-21
update pytest Version requirement 

### LINKED ISSUE #2473: Wrong pytest requirement version
As filed in the mailing list we need the version 2.5.1 but only instal 2.4.2

passiweinberger:
Fixes #2473


passiweinberger:
fixes  #2473

rhyolight:
@oxtopus Ok with you?

oxtopus:
I'm always ok with keeping requirements.txt up to date, but was the impetus for this, exactly?

rhyolight:
There was a report [on the mailing list](http://lists.numenta.org/pipermail/nupic_lists.numenta.org/2015-August/011512.html) that @passiweinberger helped debug. The solution was to update pytest.

rhyolight:
@oxtopus I'm sorry, I just realized you didn't actually give a thumbs up. :/

oxtopus:
> @oxtopus I'm sorry, I just realized you didn't actually give a thumbs up.

:rage1:

I'd like to understand why, though, since it's not a problem anywhere else.  We can't just adjust version numbers willy nilly to accommodate individual environments.

oxtopus:
For example, after reading the mailing list issue is that today, or at least at the time of the post, the version that is installed via `sudo apt-get install python-pytest` is 2.5.1, which will not always be the case.  Meanwhile, the _correct_ and supported means fore installing nupic requirements is via `python setup.py` and/or `pip install -r extensions/common/requirements.txt`, which specifies the correct version.

:-1: for this, as it's now in conflict with other downstream projects -- see https://github.com/numenta/numenta-apps/blob/master/nta.utils/requirements.txt#L10

passiweinberger:
Do the nta.utils tests work with 2.5.1 as well on iOS? 
If so, that may be updated as well? Or shouldn't it?

passiweinberger:
#2477 for a rollback though.... :)

rhyolight:
@oxtopus You are correct, this should not have been merged. My apologies.

PR #2477, opened 2015-08-22
Rollback of req change

passiweinberger:
As @oxtopus indicated the conflict...


oxtopus:
I think the original change is ok.  Instead, we'll just need to follow up with the downstream project(s) to match.

My concern is the motivation behind the change, since it was created because the user followed instructions for installing `pytest` that weren't consistent with the standard nupic instructions.  The version installed by apt is specific to the user's environment, and not directed by `requirements.txt`.  Not everyone uses ubuntu, or apt, and if they did, there's still the possibility for the versions to conflict.

passiweinberger:
Ok I see,
I tried to update to 2.5.1 and run the test and they worked, so keeping
up-to-date requirements is always a good idea, isn't it?

rhyolight:
> keeping
> up-to-date requirements is always a good idea, isn't it?

Sure until it's not. In this case because of conflicting existing versions. Some packages might require one version of a dependency and another package needs a different version. The Python packaging structure doesn't accommodate that very well.

passiweinberger:
This was for you to undo the merge, though :)
On Aug 25, 2015 7:05 PM, passiweinberger@gmail.com wrote:

Okay, Thanks :)
On Aug 25, 2015 5:13 PM, "Matthew Taylor" <notifications@github.com> wrote:

Closed #2477 <https://github.com/numenta/nupic/pull/2477>.

—
Reply to this email directly or view it on GitHub
<https://github.com/numenta/nupic/pull/2477#event-391773572>.

passiweinberger:
Okay, Thanks :)

Issue #2492, opened 2015-08-25
Allow specifying MonitoredTemporalMemory as TM implementation through OPF


chetan51:
Should be able to set `temporalImp='monitored_tm_py'` through the OPF.

PR #2495, opened 2015-08-25
Allow specifying MonitoredTemporalMemory as TM implementation through OPF

chetan51:
Fixes https://github.com/numenta/nupic/issues/2492.

Note that this is a stop-gap until https://github.com/numenta/nupic.research/pull/247 is finished.

### DIFF
--- a/nupic/research/TP_shim.py
+++ b/nupic/research/TP_shim.py
@@ -27,6 +27,9 @@ for use with OPF.
 import numpy
 
 from nupic.research.temporal_memory import TemporalMemory
+from nupic.research.monitor_mixin.temporal_memory_monitor_mixin import (
+  TemporalMemoryMonitorMixin)
+class MonitoredTemporalMemory(TemporalMemoryMonitorMixin, TemporalMemory): pass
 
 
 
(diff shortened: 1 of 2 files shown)


chetan51:
@scottpurdy

nupic/research/TP_shim.py
```diff
@@ -27,6 +27,9 @@
 import numpy
 
 from nupic.research.temporal_memory import TemporalMemory
+from nupic.research.monitor_mixin.temporal_memory_monitor_mixin import (
+  TemporalMemoryMonitorMixin)
+class MonitoredTemporalMemory(TemporalMemoryMonitorMixin, TemporalMemory): pass
```
scottpurdy: Why not define this in some place where it can be reused?

scottpurdy:
:+1:

nupic/research/TP_shim.py
chetan51: Good idea, I can put it in `nupic.research.monitor_mixin.temporal_memory_monitor_mixin`. I'll do that in a separate PR.

chetan51: Actually, on second thought, this isn't worth it because it's a mixin class, so you should be able to mix it in to any `TemporalMemory` instance, either `TemporalMemory` (python), `FastTemporalMemory`, or `TemporalMemory` (C++). I think it's better to let the client of the mixin make the decision on which one to use.

PR #2501, opened 2015-08-26
Coordinate encoder speedup using cache

### LINKED ISSUE #2150: Speed up CoordinateEncoder for streaming data
hey all, 
it would be awesome if we could speed up coordinate encoder something like `60x+` :man_with_turban: 

Use case: 
I have real world date where `3+s` ~ `600k+ data points`, but just processing with coordinate encoder takes `180+s`. 

Here's the profile:
```
        128703788 function calls (128701858 primitive calls) in 189.398 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.497    0.497  189.399  189.399 retina.py:1(<module>)
   663257    0.393    0.000  182.238    0.000 retina.py:170(run)
   389924    2.433    0.000  181.845    0.000 retina.py:47(encode)
   389924    1.597    0.000  178.010    0.000 retina.py:25(encodeIntoArray)
   389924    7.379    0.000  175.725    0.000 coordinate.py:100(encodeIntoArray)
  3509316    4.033    0.000  106.439    0.000 coordinate.py:112(<lambda>)
  3509316    9.117    0.000  102.406    0.000 coordinate.py:174(_bitForCoordinate)
  7018632   21.687    0.000   98.697    0.000 coordinate.py:151(_hashCoordinate)
  7019198   13.218    0.000   66.038    0.000 {method 'join' of 'str' objects}
   389924    6.516    0.000   56.352    0.000 coordinate.py:134(_topWCoordinates)
 21055896   22.783    0.000   52.819    0.000 coordinate.py:154(<genexpr>)
  3509316    7.557    0.000   46.932    0.000 coordinate.py:160(_orderForCoordinate)
  7018632    7.489    0.000   30.036    0.000 numeric.py:1681(array_str)
  7018635   14.164    0.000   27.245    0.000 math.py:5978(__init__)
  7018632   18.718    0.000   22.547    0.000 arrayprint.py:343(array2string)
  7018635   13.082    0.000   13.082    0.000 {_math.new_Random}
  7018633    6.228    0.000    6.228    0.000 {_hashlib.openssl_md5}
        1    0.012    0.012    5.005    5.005 spatial_pooler.py:52(__init__)
```

Limitations:
The task has some preconditions, which hopefully could be exploited! 
- it's `2D`, ranges on axis are small (1..127)

What do you think, should I try, or start from somewhere else completely?

breznak:
caches encodings for Coord encoder to make it usable for streaming data (high throughput). 
I've decided not to cache only the hashes but went further and cache whole encodings for a given input. 
This is handeled through `cacheSize` param which defaults to 0 = off. 

I;ve tried `OrderedDict` or even better `cachetools` (LeastFrequentlyUsed, LeastRecentlyUsed caches) to better handle full cache (removal), but they don't cut it. (well, still faster than current, but cca 2x slower then this suggested dict-only approach), on the other hand, I have to remove "random" elements when purging cache. 

Results: 
`   ncalls  tottime  percall  cumtime  percall filename:lineno(function)`
new:  ` 63314    0.444    0.000    0.490    0.000 coordinate.py:114(encodeIntoArray)` ~63000x
old:  `14702    0.809    0.000   31.369    0.002 coordinate.py:114(encodeIntoArray)`   ~14000x

For serialization, I did not bother with storing a dict (maybe possible?), but just drop it, as it's only an optimization.  

Fixes #2150

### DIFF
--- a/nupic/encoders/coordinate.py
+++ b/nupic/encoders/coordinate.py
@@ -101,10 +101,14 @@ class CoordinateEncoder(Encoder):
 
   
   @cache(maxsize=100, typed=False)
-  def encode(self, coordinate, radius, **kwds):
+  def encode(self, coordinate, radius, cacheSizeName='cacheSize', **kwds):
     """
     @param coordinate - a tuple (is hashable) of each axis, eg (x,y,z)
     @param radius - float
+    @param cacheSizeName - string (opt), points to the attribute of the object
+                         (CoordinateEncoder().cacheSize) which holds value for
+                         the size of the cache. 
+                         See nupic.utils.lru_cache and tests for explanation.
     """
     out = numpy.zeros(self.n)
     coordArr = numpy.array(coordinate)
--- a/nupic/encoders/coordinate.py
+++ b/nupic/encoders/coordinate.py
@@ -19,6 +19,7 @@
 # http://numenta.org/licenses/
 # ----------------------------------------------------------------------
 
+from fastcache import clru_cache as cache
 import hashlib
 import itertools
 
@@ -57,6 +58,7 @@ class CoordinateEncoder(Encoder):
                w=21,
                n=1000,
                name=None,
+               cacheSize=10000,
                verbosity=0):
     """
     See `nupic.encoders.base.Encoder` for more information.
@@ -76,6 +78,7 @@ class CoordinateEncoder(Encoder):
     self.n = n
     self.verbosity = verbosity
     self.encoders = None
+    self.cacheSize = cacheSize
 
     if name is None:
       name = "[%s:%s]" % (self.n, self.w)
@@ -96,6 +99,20 @@ class CoordinateEncoder(Encoder):
     """See `nupic.encoders.base.Encoder` for more information."""
     return numpy.array([0]*len(inputData))
 
+  
+  @cache(maxsize=100, typed=False)
+  def encode(self, coordinate, radius):
+    """
+    @param coordinate - a tuple (is hashable) of each axis, eg (x,y,z)
+    @param radius - float
+    """
+    out = numpy.zeros(self.n)
+    coordArr = numpy.array(coordinate)
+    # make ndarray read-only so it can be hashed
+    coordArr.flags.writeable = False
+    self.encodeIntoArray((coordArr, radius), out)
+    return out
+    
 
   def encodeIntoArray(self, inputData, output):
     """
@@ -189,6 +206,7 @@ class CoordinateEncoder(Encoder):
     print "CoordinateEncoder:"
     print "  w:   %d" % self.w
     print "  n:   %d" % self.n
+    print self.encode.cache_info()
 
 
   @classmethod
@@ -198,6 +216,7 @@ class CoordinateEncoder(Encoder):
     encoder.n = proto.n
     encoder.verbosity = proto.verbosity
     encoder.name = proto.name
+    encoder.cacheSize = proto.cacheSize
     return encoder
 
 
@@ -206,3 +225,4 @@ class CoordinateEncoder(Encoder):
     proto.n = self.n
     proto.verbosity = self.verbosity
     proto.name = self.name
+    proto.cacheSize = self.cacheSize
--- a/nupic/encoders/coordinate.py
+++ b/nupic/encoders/coordinate.py
@@ -165,15 +165,6 @@ class CoordinateEncoder(Encoder):
     return coordinates[indices]
 
 
-  @staticmethod
-  def _hashCoordinate(coordinate):
-    """Hash a coordinate to a 64 bit integer."""
-    coordinateStr = ",".join(str(v) for v in coordinate)
-    # Compute the hash and convert to 64 bit int.
-    hash = int(int(hashlib.md5(coordinateStr).hexdigest(), 16) % (2 ** 64))
-    return hash
-
-
   @classmethod
   def _orderForCoordinate(cls, coordinate):
     """
@@ -183,7 +174,7 @@ class CoordinateEncoder(Encoder):
     @return (float) A value in the interval [0, 1), representing the
                     order of the coordinate
     """
-    seed = cls._hashCoordinate(coordinate)
+    seed = hash(coordinate.data)
     rng = Random(seed)
     return rng.getReal64()
 
@@ -197,7 +188,7 @@ class CoordinateEncoder(Encoder):
     @param n (int) The number of available bits in the SDR
     @return (int) The index to a bit in the SDR
     """
-    seed = cls._hashCoordinate(coordinate)
+    seed = hash(coordinate.data)
     rng = Random(seed)
     return rng.getUInt32(n)
 
--- a/nupic/encoders/coordinate.py
+++ b/nupic/encoders/coordinate.py
@@ -19,7 +19,7 @@
 # http://numenta.org/licenses/
 # ----------------------------------------------------------------------
 
-from fastcache import clru_cache as cache
+from nupic.utils import lru_cache as cache
 import hashlib
 import itertools
 
--- a/nupic/encoders/coordinate.py
+++ b/nupic/encoders/coordinate.py
@@ -165,6 +165,15 @@ class CoordinateEncoder(Encoder):
     return coordinates[indices]
 
 
+  @staticmethod
+  def _hashCoordinate(coordinate):
+    """Hash a coordinate to a 64 bit integer."""
+    coordinateStr = ",".join(str(v) for v in coordinate)
+    # Compute the hash and convert to 64 bit int.
+    hash = int(int(hashlib.md5(coordinateStr).hexdigest(), 16) % (2 ** 64))
+    return hash
+
+
   @classmethod
   def _orderForCoordinate(cls, coordinate):
     """
@@ -174,7 +183,7 @@ class CoordinateEncoder(Encoder):
     @return (float) A value in the interval [0, 1), representing the
                     order of the coordinate
     """
-    seed = hash(coordinate.data)
+    seed = cls._hashCoordinate(coordinate)
     rng = Random(seed)
     return rng.getReal64()
 
@@ -188,7 +197,7 @@ class CoordinateEncoder(Encoder):
     @param n (int) The number of available bits in the SDR
     @return (int) The index to a bit in the SDR
     """
-    seed = hash(coordinate.data)
+    seed = cls._hashCoordinate(coordinate)
     rng = Random(seed)
     return rng.getUInt32(n)
 
--- a/nupic/encoders/coordinate.py
+++ b/nupic/encoders/coordinate.py
@@ -101,7 +101,7 @@ class CoordinateEncoder(Encoder):
 
   
   @cache(maxsize=100, typed=False)
-  def encode(self, coordinate, radius):
+  def encode(self, coordinate, radius, **kwds):
     """
     @param coordinate - a tuple (is hashable) of each axis, eg (x,y,z)
     @param radius - float
--- a/nupic/encoders/coordinate.py
+++ b/nupic/encoders/coordinate.py
@@ -100,15 +100,11 @@ class CoordinateEncoder(Encoder):
     return numpy.array([0]*len(inputData))
 
   
-  @cache(maxsize=100, typed=False)
-  def encode(self, coordinate, radius, cacheSizeName='cacheSize', **kwds):
+  @cache(sizeref='cacheSize', typed=False)
+  def encode(self, coordinate, radius, **kwds):
     """
     @param coordinate - a tuple (is hashable) of each axis, eg (x,y,z)
     @param radius - float
-    @param cacheSizeName - string (opt), points to the attribute of the object
-                         (CoordinateEncoder().cacheSize) which holds value for
-                         the size of the cache. 
-                         See nupic.utils.lru_cache and tests for explanation.
     """
     out = numpy.zeros(self.n)
     coordArr = numpy.array(coordinate)
--- a/nupic/utils.py
+++ b/nupic/utils.py
@@ -259,79 +259,100 @@ def lru_cache(maxsize=100, typed=False, sizeref=None):
         properties = dict()
         properties['maxsize'] = maxsize         # make maxsize updatable non-locally
         properties['init'] = True               # do _setSize
-                  
-        if properties['maxsize'] == 0:
+                 
+        ### cache wrappers
+        # exported methods optimized for specific cache settings
+        def wrapperNoCache(*args, **kwds):
+            # no caching, just do a statistics update after a successful call
+            result = user_function(*args, **kwds)
+            stats[MISSES] += 1
+            return result
 
-            def wrapper(*args, **kwds):
-                # no caching, just do a statistics update after a successful call
-                result = user_function(*args, **kwds)
-                stats[MISSES] += 1
+
+        def wrapperUnlimitedCache(*args, **kwds):
+            # simple caching without ordering or size limit
+            key = make_key(args, kwds, typed)
+            result = cache_get(key, root)   # root used here as a unique not-found sentinel
+            if result is not root:
+                stats[HITS] += 1
                 return result
+            result = user_function(*args, **kwds)
+            cache[key] = result
+            stats[MISSES] += 1
+            return result
 
-        elif properties['maxsize'] is None:
 
-            def wrapper(*args, **kwds):
-                # simple caching without ordering or size limit
-                key = make_key(args, kwds, typed)
-                result = cache_get(key, root)   # root used here as a unique not-found sentinel
-                if result is not root:
+        def wrapperLimitedCache(*args, **kwds):
+            # size limited caching that tracks accesses by recency
+            key = make_key(args, kwds, typed) if kwds or typed else args
+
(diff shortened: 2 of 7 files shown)


breznak:
@chetan51 @scottpurdy please review

breznak:
@rhyolight some strange errs from Travis: 
https://travis-ci.org/numenta/nupic/jobs/77311196#L314
https://travis-ci.org/numenta/nupic/jobs/77311196#L1140
etc ..

breznak:
At PR logic: ...hashing an array, I know :shame: hold on pls :)

breznak:
... pretends like nothing happened :stuck_out_tongue:

rhyolight:
> @rhyolight some strange errs from Travis: 
> https://travis-ci.org/numenta/nupic/jobs/77311196#L314
> https://travis-ci.org/numenta/nupic/jobs/77311196#L1140
> etc ..

Yes, strange. I restarted all the builds.

scottpurdy:
You could create a decorator that users could use wherever they want. For instance, in client code they could do something like this:

```
from nupic.encoders import SomeEncoder
from nupic.utils import cache

class CachedSomeEncoder(SomeEncoder):
  @cache(size=1000)
  def encode(self, value):
    super(self, CachedSomeEncoder).encode(value)
```

This keeps the opinion of whether caching should be done out of the core implementations and gives users flexibility to apply caching to any deterministic functions they want.

breznak:
> You could create a decorator that users could use wherever they want.

This is a nice approach, and is exactly how `cachetools` work (but they
don't cut it for speed for this purpose). On the other hand, my cache
implementation uses a stupid replacement mechanism (compared to theirs
LRU,LFU,...), so I'm not sure if it's worth to write a decorator for this
cache, or just leave it a part of CoordinateEncoder which is really slow.
Maybe we could identify non-learning parts in the system where caching
would make sense? (And leave it for a separate issue?)

chetan51:
I prefer @scottpurdy's decorator solution. The problem I have with this PR is that it significantly complicated the implementation code for Coordinate encoder. If we can keep the cache code out of it, I would be happier.

breznak:
Agreed the decorator would be nice, problem is that `ndarray` is unhashable, so I had to add encode() to overcome that. But encode() is imho useful for this encoder anyway and all encoders must implement it. 
The solution with `fastcache` is even faster.

nupic/encoders/coordinate.py
```diff
@@ -96,6 +104,20 @@ def getScalars(self, inputData):
     """See `nupic.encoders.base.Encoder` for more information."""
     return numpy.array([0]*len(inputData))
 
+  
+  @cache(maxsize=100, typed=False)
```
breznak: I'd like to change this value to `self.cacheSize` but can't access instance (or class, global maybe yes) members. Is it possoble to achieve without a new wrapper around this wrapper?

breznak:
@chetan51 @scottpurdy got it! 
Implemented a LRU cache as a decorator, taken from `functools.lru_cache` backport, so it's compatible with python 2.6+ & 3.0+. They have  a C version underway, so in future we could seemlessly benefit from that. 
+ I've implemented some improvements to be able to set the cache's size param at runtime, those are submitted for discussion in upstream: https://bugs.python.org/issue24969

Includes tests for `nupic.utils.lru_cache` and Coordinate encoder which uses the caching. 

(if you are ok with this, suggested next steps (not in this PR):
* separate nupic.utils to a directory with files in it, eg nupic.utils.cache, ...
* apply cache where possible- some methods in SP, ...

chetan51:
@breznak Doesn't the OPF use `encodeIntoArray` directly? In which case it wouldn't benefit from this cache optimization?

breznak:
> Doesn't the OPF use encodeIntoArray directly? In which case it wouldn't
benefit from this cache optimization?

Hmm, likely true. I can apply the decorator to encodeIntoArray() instead,
in which case I'll waste a few assignments for those who use encode().
Btw, shouldn't OPF call encode() instead?

chetan51:
@scottpurdy What do you suggest?

breznak:
A hack would be to cache both, which is contradictory, but OPF would gain,
and encode() would only suffer from double-buffering.

scottpurdy:
It looks like we will be moving away from `encodeIntoArray` in favor of standardizing on `encode` that returns a new array. Not sure what is best right now, or if it matters what you do now.

breznak:
I'd suggest leaving encodeIntoArray() out for now, the reason it's not cached is that its arg is a `ndarray` which is unhashable by default. I need to wrap the call to encodeIntoArray (therefore encode()) and do the work there (eg set `ndarray.flags.writeable = False` and hash `ndarray.data` instead). 

I may be able to add additional wrapper that will handle the ndarray, but I'd like to leave that only if we find another usecases for `@cache` that deal with ndarrays.

breznak:
Ok, the cache is ready. Results for Coordinate encoder are quite nice, you can check with 
`python -m cProfile --sort cumtime scripts/profiling/enc_profile.py ` 
(may comment out the other encoders there in the for-loop)

What I do not understand is: 

```
CacheInfo(hits=94861, misses=5139, maxsize=500, currsize=500)
None
         5333857 function calls (5331961 primitive calls) in 6.432 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.001    0.001    6.432    6.432 enc_profile.py:25(<module>)
       10    0.066    0.007    6.208    0.621 enc_profile.py:33(profileEnc)
   100000    0.113    0.000    6.140    0.000 utils.py:350(wrapper)
   100000    0.196    0.000    6.006    0.000 utils.py:292(wrapperLimitedCache)
     5139    0.024    0.000    5.365    0.001 coordinate.py:103(encode)
     5139    0.154    0.000    5.307    0.001 coordinate.py:117(encodeIntoArray)
   107919    0.101    0.000    3.110    0.000 coordinate.py:129(<lambda>)
   107919    0.262    0.000    3.010    0.000 coordinate.py:191(_bitForCoordinate)
   236394    0.700    0.000    2.997    0.000 coordinate.py:168(_hashCoordinate)
   236820    0.410    0.000    1.956    0.000 {method 'join' of 'str' objects}
     5139    0.159    0.000    1.922    0.000 coordinate.py:151(_topWCoordinates)
   128475    0.262    0.000    1.707    0.000 coordinate.py:177(_orderForCoordinate)
   709182    0.674    0.000    1.546    0.000 coordinate.py:171(<genexpr>)
   236404    0.467    0.000    0.986    0.000 math.py:5978(__init__)
```

* Odd 1: data generated from 0..maxV, cache size = maxV, why do I get cache misses? about 2% - ~5000 misses in 100k runs. Same behavior confirmed with an alternative cache (so not a flaw in my wrapper), increasing cachesize to maxV*1.33 does not help (if it were problems with pruning hashtable in cache). What can this be? A flaw in the cache, collisions in the hashing alg, or in cProfile? I've verified the cache and it seems to return values correctly. 

* Odd 2: with 5k misses (ok), there're 5k calls to encode()/encodeIntoArray(), but then why are there 100k+ calls to the support functions - _hashCoordinate, _bitForCoordinate, ...etc?

nupic/encoders/coordinate.py
```diff
@@ -96,6 +99,20 @@ def getScalars(self, inputData):
     """See `nupic.encoders.base.Encoder` for more information."""
     return numpy.array([0]*len(inputData))
 
+  
+  @cache(sizeref='cacheSize', typed=False)
+  def encode(self, coordinate, radius, **kwds):
```
chetan51: Please rename `kwds` to `kwargs` (the standard name for it)

tests/unit/nupic/encoders/coordinate_test.py
```diff
@@ -303,6 +303,22 @@ def testReadWrite(self):
     self.assertTrue(np.array_equal(output1, output2))
 
 
+  def testCachedEncoding(self):
+    eDefault = CoordinateEncoder()
+    eCached  = CoordinateEncoder(cacheSize=9)
+    iters = 10
+    for _ in xrange(iters):
+      (x,y) = np.random.randint(0, 2, 2)
```
chetan51: Is this test deterministic?

tests/unit/nupic/encoders/coordinate_test.py
```diff
@@ -303,6 +303,22 @@ def testReadWrite(self):
...
+    eCached  = CoordinateEncoder(cacheSize=9)
+    iters = 10
+    for _ in xrange(iters):
+      (x,y) = np.random.randint(0, 2, 2) 
+
+      res1 = encode(eDefault, np.array([x,y]), 2) # uses encodeIntoArray() directly
+      res2 = eCached.encode((x,y), 2) # calls encode() with cache
+
+      self.assertTrue(np.array_equal(res1, res2), "Cached and non-cached encodings differ! %r %r" % (res1, res2))
+    print eCached.dump()
```
chetan51: Please remove print statements from tests

tests/unit/nupic/utils_test.py
```diff
@@ -183,7 +185,73 @@ def testEquals(self):
...
+        return x**2
+
+    a = A(42)
+
+    # method 1: maxsize from method's arg, this is an override
+    for d in xrange(10):
+      a.foo(d, cacheSize="5")
+    mx = int(a.foo.cache_info().maxsize)
+    self.assertEqual(mx, 5, "Cache size from method's arg 'cacheSize' was set to 5, but is %r" % (mx))
+    print a.foo.cache_info()
```
chetan51: Remove print statements from tests here and below

nupic/utils.py
```diff
@@ -127,3 +127,258 @@ def write(self, proto):
     proto.windowSize = self.windowSize
     proto.slidingWindow = self.slidingWindow
     proto.total = self.total
+
+
+
+#######################################################
+# code from functools.lru_cache for Python 3.3, backports from ActiveState:
```
chetan51: @rhyolight Is this code okay to include in our codebase (license-wise)?

nupic/utils.py
```diff
@@ -127,3 +127,258 @@ def write(self, proto):
...
+    __slots__ = 'hashvalue'
+
+    def __init__(self, tup, hash=hash):
+        self[:] = tup
+        self.hashvalue = hash(tup)
+
+    def __hash__(self):
+        return self.hashvalue
+
+def _make_key(args, kwds, typed,
```
chetan51: Rename to `_makeKey`

nupic/utils.py
```diff
@@ -127,3 +127,258 @@ def write(self, proto):
...
+    print "setting maxsize to object's member %s = %i" % (sizeref, mxRef)
+    properties['maxsize'] = mxRef
+  elif mxRef is None or mxArg is None:
+    print "setting unlimited cache (None)"
+    properties['maxsize'] = None
+  else:
+    raise ValueError("lru_cache: No arg or object's member of name %s, as specified in 'sizeref'" % (sizeref))
+
+
+def lru_cache(maxsize=100, typed=False, sizeref=None):
```
chetan51: Rename to `lruCache`

chetan51:
Done with review.

scottpurdy:
This implementation adds the cache for all users of the CoordinateEncoder. Why not just let the clients determine if they want it (with a subclass that applies the decorator)? Plus if the client is using a network they would probably want to subclass the region instead and apply something similar there. Actually if there was a LearningEncoder subclass of Encoder then the RecordSensor could have an option that would enable caching for all non-learning encoders. Applying the cache when needed also removes the need for the `sizeref` hack that you added. If you go this route it would still be nice to have example code showing caching being used and an option added to the RecordSensor.

One case where the cache would not be good for the CoordinateEncoder is when it is set up to have fine-grained resolution with a large radius. In these cases, the input coordinates may very rarely match exactly and the large radius ensures that the encodings still have overlap for similar, but not exact, coordinates. In this case the caching mechanism would just slow the function down. I don't know if normal use cases would have high cache hit rates or not but perhaps @chetan51 has an intuition for it?

And instead of copying code into our repo we should probably just add a dependency on one of the library recommendations here:
http://stackoverflow.com/a/9674327

tests/unit/nupic/encoders/coordinate_test.py
breznak: nope, I will look to add a seed, or better use Random from nupic.core

nupic/utils.py
breznak: I can do that but I'd rather leave as-is, as it's part of the library

breznak:
hi @scottpurdy ,

> Why not just let the clients determine if they want it (with a subclass that applies the decorator)?

(this looks like a perfect use for an `ICached` interface, but that's for another time) I can do a subclass instead, but I see no point in it: now you can control the use of the cache with a param `cacheSize`, we can set the `default to 0` if you care. The overhead of a method decorated with a disabled cache is minimal. But I'd prefer the optimization be opt-out if it works, so it's transparent to a user who cares nothing about Coord's details. 

> a LearningEncoder subclass of Encoder then the RecordSensor could have an option that would enable caching for all non-learning encoders.

hey, great idea! I don't want to do that with the initial PR on introduction of a cache (I even consider if to remove the Coord example, and just give cache and its tests). but we must not forget it for later improvements. (Note, cache has problems with encodeIntoArrays() in some encoders that pass ndarray, I've added functionality to overcome this..the generic caching might use some more work)

> If you go this route it would still be nice to have example code showing caching being used and an option added to the RecordSensor.

I'd rather not focus on it now

>  Applying the cache when needed also removes the need for the sizeref hack that you added. 

I wish, but I don't see how? You only know maxsize at runtime
```
class CachedCoordinateEncoder(CoordinateEncoder):

  @cache(maxsize=???)
  def encode():
    return super().encode()
```

> we should probably just add a dependency on one of the library recommendations

I've tried most of the libraries/recipes, went with functools' version as its future compatible, has promised c++; I may try to rework to write my wrapper around the decorator again, as a simple import would be cleaner. Just import now does not work because I've added functionality & hacks not present in the original.

nupic/utils.py
rhyolight: Yes, because `https://code.activestate.com/recipes/578078/` states the content is MIT.

scottpurdy:
> I wish, but I don't see how? You only know maxsize at runtime

That is exactly why you should leave it to clients to create the subclass and add caching. This allows us to use an existing backport and we don't have to add the `sizeref` functionality. It also means that we don't have to add caching-specific state to the serialization. And it gives the clients the option to apply caching at whichever level makes sense (encoder vs region).

@chetan51 - I've left my opinion but will leave the decisions/review up to you.

breznak:
> That is exactly why you should leave it to clients to create the subclass and add caching. 

@scottpurdy do you think it's useless then? Here are points I care for:
 * ease of use - there's a difference in creating an instance with a param `e = CoordinateEncoder(cacheSize=1000)`, and actually editing the code and building from source. 
* cache can be used with the binary package
* it's our job to figure which methods are best to be wrapped, we can't do that w/o runtime param. 

Alternative: atleast use some reasonable cache size: it's a LRU, so maxsize 1000 will do a pretty decent job.

chetan51:
I think caching will improve performance in most use-cases for the Coordinate Encoder. I also prefer ease of use, so I'm fine with it being a parameter, but I wish there was a way to keep it out of the serialization state (and also ensure that it's easy to add caching to the other encoders). Any ideas on how to achieve both of those goals @breznak?

chetan51:
@breznak Could you try caching the functions `_orderForCoordinate` and `_bitForCoordinate` instead, and see whether you still get good speedup? Those functions are going to be called with the same arguments more often than `encode` will be.

breznak:
>  I wish there was a way to keep it out of the serialization state

cache is dropped at serialization, that makes sense. Only the parameter of
the object - cacheSize is stored.

> and also ensure that it's easy to add caching to the other encoders

the wrapper is generic, depends what we find - eg "learning encoders"
probably would not be cached, but we can still gain on some sub-routines
that get called many times.

> Could you try caching the functions _orderForCoordinate and
_bitForCoordinate instead, and see whether you still get good speedup?
Those functions are going to be called with the same arguments more often
than encode will be.

Yes! This is a good idea to solve encodeIntoAtray caching! Esp. since the
concern of many misses due to a few bits flipped. It makes  sense to cache
the highest-level functions, so the rest is not even called, so I'd suggest
both: cached encode() for those who call it - it will be the most
effective. And cached _orderForCoord, _bit4Coord to gains for callers of
encodeIntoArray.

Ok, as it looks the PR has a future, I'll try to rewrite the wrapper and
separate it from the cache, which we can get from pip. Hold on 1 day please.

breznak:
>  try caching the functions _orderForCoordinate and _bitForCoordinate
 instead

the decorator is having some troubles with static/class-methods, is there a
motivation why these in the encoder can't be instance methods?

chetan51:
> the decorator is having some troubles with static/class-methods, is there a
motivation why these in the encoder can't be instance methods?

Because they didn't need to be. If you need them to be, you can change them.

rhyolight:
@breznak said (12 days ago):
> Ok, as it looks the PR has a future, I'll try to rewrite the wrapper and
> separate it from the cache, which we can get from pip. Hold on 1 day please.

I'm labelling as "in progress" now.

PR #2507, opened 2015-08-26
Fixes #2478 Bug in temporal memory orphan decay test

### LINKED ISSUE #2478: Test H10 in TM integration test is invalid
There is an error in the loop on the second trial which caused the output to have the desired stats. The test fails when the loop is corrected. 

Secondly, the setting of "predictedSegmentDecrement" to 0.004 has no effect when the loop is corrected - indicating that there may need to be some algorithm enhancement in order to make the parameter have the desired effect. (Other than lowering the permanence of the Synapse resulting in eventual removal of the Synapse - which works).

**Line 532: extensive_temporal_memory_test.py**
``` Python

for _ in xrange(10):
      self.feedTM(sequenceNoisy[i])
```

**Should be:**

``` Python
for i in xrange(10):
      self.feedTM(sequenceNoisy[i])
```

Where the underscore ("_") should be an "i" so that all the sequences are actually entered.

This is the test description:

> Orphan Decay mechanism reduce predicted inactive cells (extra predictions).
    Test feeds in noisy sequences (X = 0.05) to TM with and without orphan decay.
    TM with orphan decay should has many fewer predicted inactive columns.
    Parameters the same as B11, and sequences like H9.

When the loop is "corrected" you will see by setting the "predictedSegmentDecrement" variable to "0.000" that it has no effect on the "predictedInactiveColumns" mean or sum. I had to set it to "0.999" in order for their to be noticeable reduction (by only half) - which I think may need some care and loving in order to have a more pronounced effect.

This took me **3 Days** of intense debugging before I began to suspect the problem wasn't with the code I was porting to. It is such a "subtle" change that I didn't even notice it!

ywcui1990:
Fixes #2478 

@cogmission @chetan51 @rhyolight   Please review

@cogmission I found that predictedSegmentDecrement=0.004 is too small to have a measurable effect on the number of predicted inactive segment. I increased this parameter to 0.04 in the test. I also added the additional check to make sure there is the same number of predicted active columns with and without orphan decay.

Here is the result I got on this test. You can see that the only difference caused by orphan decay is a smaller number of predicted->inactive columns/cells

image

Instead of using a higher value of predictedSegmentDecrement, I could also increase the size of the training data to get a measurable effect. Here is the result with 50 set of random sequences and predictedSegmentDecrement=0.01

image

In practice, I think the default value of predictedSegmentDecrement should be small to allow multiple predictions. It will have an effect on reducing extra predictions with large enough datasets.
\

### DIFF
--- a/tests/integration/nupic/algorithms/extensive_temporal_memory_test.py
+++ b/tests/integration/nupic/algorithms/extensive_temporal_memory_test.py
@@ -505,41 +505,58 @@ class ExtensiveTemporalMemoryTest(AbstractTemporalMemoryTest):
       self.tm.mmGetTraceUnpredictedActiveColumns())
     self.assertTrue(unpredictedActiveColumnsMetric.mean < 3)
 
+
   def testH10(self):
     """Orphan Decay mechanism reduce predicted inactive cells (extra predictions).
     Test feeds in noisy sequences (X = 0.05) to TM with and without orphan decay.
     TM with orphan decay should has many fewer predicted inactive columns.
     Parameters the same as B11, and sequences like H9."""
+
+    # train TM on noisy sequences with orphan decay turned off
     self.init({"cellsPerColumn": 4,
                "activationThreshold": 8})
 
     numbers = self.sequenceMachine.generateNumbers(2, 20, (10, 15))
     sequence = self.sequenceMachine.generateFromNumbers(numbers)
+
     sequenceNoisy = dict()
     for i in xrange(10):
       sequenceNoisy[i] = self.sequenceMachine.addSpatialNoise(sequence, 0.05)
       self.feedTM(sequenceNoisy[i])
+    self.tm.mmClearHistory()
 
     self._testTM(sequence)
+
     predictedInactiveColumnsMetric = self.tm.mmGetMetricFromTrace(
       self.tm.mmGetTracePredictedInactiveColumns())
-    predictedInactiveColumnsMean1 = predictedInactiveColumnsMetric.mean
+    predictedActiveColumnsMetric = self.tm.mmGetMetricFromTrace(
+      self.tm.mmGetTracePredictedActiveColumns())
 
+    predictedInactiveColumnsMeanNoOrphanDecay = predictedInactiveColumnsMetric.mean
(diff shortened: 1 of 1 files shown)


cogmission:
@ywcui1990 Your my hero! Awesome! :-)

I will do the same for HTM.java provided everyone else agrees!

rhyolight:
Thanks, Yuiwei. If @cogmission is happy, I'm happy. But @chetan51 should be the final reviewer.

tests/integration/nupic/algorithms/extensive_temporal_memory_test.py
```diff
@@ -505,41 +505,58 @@ def testH9(self):
...
     predictedInactiveColumnsMetric = self.tm.mmGetMetricFromTrace(
       self.tm.mmGetTracePredictedInactiveColumns())
-    predictedInactiveColumnsMean2 = predictedInactiveColumnsMetric.mean
+    predictedActiveColumnsMetric = self.tm.mmGetMetricFromTrace(
+      self.tm.mmGetTracePredictedActiveColumns())
 
-    self.assertGreater(predictedInactiveColumnsMean1, 0)
-    self.assertGreater(predictedInactiveColumnsMean1, predictedInactiveColumnsMean2)
+    predictedInactiveColumnsMeanOrphanDecay = predictedInactiveColumnsMetric.mean
+    predictedActiveColumnsMeanOrphanDecay = predictedActiveColumnsMetric.mean
```
chetan51: I'm confused; why are you replacing these values with new ones here?

ywcui1990: I didn't. These are two new values for TM with orphan decay. The variables created in line 535-536 are for TM without orphan decay.

chetan51: Ah, I misread that, sorry.

subutai:
> @subutai I guess your original "default" estimate should be changed?

I'd like to understand this better first so I will discuss with @ywcui1990. We don''t want to set default parameters based on specific artificial experiments, but based on what we expect in real world scenarios.

PR #2526, opened 2015-08-31
Minor cleanup in spatial pooler init

### LINKED ISSUE #2542: Fixup assert usage in SpatialPooler
`assert` is used in SpatialPooler constructor to validate user input.  `assert`'s are ignored when python interpreter is invoked with `-O` argument.  Meanwhile, it's generally bad practice to use `assert` for user input.  Instead, `assert` should be used to validate internal contracts to indicate that the code did something unexpected, independent of user input.

oxtopus:
Fixes #2542 

- Refer to sequences rather than lists in docstring
- Use first-class exceptions in lieu of `assert`'s, which may not apply if python is running in optimized mode
- Removed useless comments and extraneous whitespace

### DIFF
--- a/nupic/research/spatial_pooler.py
+++ b/nupic/research/spatial_pooler.py
@@ -34,6 +34,12 @@ uintType = "uint32"
 VERSION = 2
 
 
+
+class InvalidSPParamUserValueError(ValueError):
+  pass
+
+
+
 class SpatialPooler(object):
   """
   This class implements the spatial pooler. It is in charge of handling the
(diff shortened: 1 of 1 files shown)


breznak:
> Use first-class exceptions in lieu of assert's, which may not apply if python is running in optimized mode

Wasn't that the intended use with asserts? Just to make 200% sure in normal/debug runs, and then for hard-core speed just drop the checks?

nupic/research/spatial_pooler.py
```diff
@@ -34,6 +34,12 @@
 VERSION = 2
 
 
+
+class InvalidSPParamUserValueError(ValueError):
```
scottpurdy: "InvalidSPParamError" seems just as descriptive but shorter.

scottpurdy:
@breznak - I agree with the change to drop asserts in this case. they are in the constructor, not some tight loop that gets called a million times, and we would want them to be raised. In an application where the client can specify params, you need to perform these checks even in production.

The `assert` functionality is nice in cases where you are checking algorithm invariants that would only fail if there was a real bug in the code. For anything that could be caused by user input you shouldn't use an `assert` in my opinion.

scottpurdy:
:+1:

PR #2535, opened 2015-09-02
Replace saving of all historical data points with a sliding window in AnomalyLikelihood class

vitaly-krugl:
Replace saving of all historical data points with a sliding window in AnomalyLikelihood class.

NOTE: This improved memory and performance a ton! `for i in xrange(1000000): a.anomalyProbability(1.0, 0.9)` went from multiple hours to just under 10 minutes on my MacBookPro.

Also, added a new method AnomalyLikelihood.forceModelRefresh: "Force refresh of the anomaly likelihood model's state"

CC @scottpurdy, @subutai, @BoltzmannBrain 

@jcasner, @rhyolight: can we pull this into 0.2.12? https://github.com/numenta/numenta-apps/pull/277 depends on it.

### DIFF
--- a/nupic/algorithms/anomaly_likelihood.py
+++ b/nupic/algorithms/anomaly_likelihood.py
@@ -152,6 +152,12 @@ class AnomalyLikelihood(object):
   def _calcSkipRecords(numIngested, windowSize, learningPeriod):
     """Return the value of skipRecords for passing to estimateAnomalyLikelihoods
 
+    If `windowSize` is very large (bigger than the amount of data) then this
+    could just return `learningPeriod`. But when some values have fallen out of
+    the historical sliding window of anomaly records, then we have to take those
+    into account as well so we return the `learningPeriod` minus the number
+    shifted out.
+
     :param int numIngested: number of data points that have been added to the
       sliding window of historical data points.
     :param int windowSize: size of sliding window of historical data points.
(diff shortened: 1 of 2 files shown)


nupic/algorithms/anomaly_likelihood.py
```diff
@@ -134,39 +148,65 @@ def computeLogLikelihood(likelihood):
     return math.log(1.0000000001 - likelihood) / -23.02585084720009
 
 
+  @staticmethod
+  def _calcSkipRecords(numIngested, windowSize, learningPeriod):
+    """Return the value of skipRecords for passing to estimateAnomalyLikelihoods
```
scottpurdy: This doc string could be more clean. Something like this:

```
"""Return the value of skipRecords for passing to estimateAnomalyLikelihoods.

If windowSize was very large (bigger than the amount of data)
then this could just return `learningPeriod`. But when some values
have fallen out of the historical window of anomaly scores then
we have to take those into account as well so we return the
`learningPeriod` minus the number shifted out.

...
```

scottpurdy:
:+1: aside from the doc string suggestion on calculating the skip records number. Thanks for including test cases that cover the new functionality.

vitaly-krugl:
Thanks @scottpurdy! I updated the `AnomalyLikelihood._calcSkipRecords` docstring per your suggestion. Merging...

rhyolight:
> @jcasner, @rhyolight: can we pull this into 0.2.12? numenta/numenta-apps#277 depends on it.

@vitaly-krugl Sure thing. I have released 0.2.12 yet. Waiting on a nod from @jcasner.

breznak:
Awesome! thanks @vitaly-krugl ! :+1:

vitaly-krugl:
@breznak: thanks for the feedback. There is a follow-up change pending at https://github.com/numenta/nupic/pull/2544

PR #2544, opened 2015-09-02
Change default values of AnomalyLikelihood constructor args to match the empirically-optimized values in numenta-apps/htmengine's AnomalyLikelihoodHelper

vitaly-krugl:
Changed default values of AnomalyLikelihood constructor args to match the empirically-optimized values in numenta-apps/htmengine's AnomalyLikelihoodHelper.

Added reestimationPeriod arg to AnomalyLikelihood constructor.

Added logic in AnomalyLikelihood.anomalyProbability() to force greedy refresh of the
distribution when anomaly probability > 0.99 to match the empirically-optimized logic
in numenta-apps/htmengine's AnomalyLikelihoodHelper.

Implemented additional unit tests for AnomalyLikelihood class.

Deleted the recently-added AnomalyLikelihood.forceModelStateRefresh() method.

Fixed pylint findings in anomaly_likelihood_test.py.

@scottpurdy: please review

Please make sure this change gets into the nupic release of v0.2.13, since it is needed by https://github.com/numenta/numenta-apps/pull/277.

CC @rhyolight @jcasner

### DIFF
--- a/nupic/algorithms/anomaly_likelihood.py
+++ b/nupic/algorithms/anomaly_likelihood.py
@@ -73,10 +73,14 @@ class AnomalyLikelihood(object):
 
 
   def __init__(self,
-               claLearningPeriod=300,
-               estimationSamples=300,
-               estimationWindowSize=8640):
+               claLearningPeriod=288,
+               estimationSamples=100,
+               estimationWindowSize=8640,
+               reestimationPeriod=10):
     """
+    NOTE: Anomaly likelihood scores are reported at a flat 0.5 for
+    claLearningPeriod + estimationSamples iterations.
+
     :param claLearningPeriod: the number of iterations required for the CLA to
     learn the basic patterns in the dataset and for the anomaly score to 'settle
     down'. The default is based on empirical observations but in reality this
@@ -84,18 +88,20 @@ class AnomalyLikelihood(object):
     is that real anomalies might get ignored and not flagged.
 
     :param estimationSamples: the number of reasonable anomaly scores required
-    for the initial estimate of the Gaussian. The default of 300 records is
+    for the initial estimate of the Gaussian. The default of 100 records is
     reasonable - we just need sufficient samples to get a decent estimate for
     the Gaussian. It's unlikely you will need to tune this since the Gaussian is
-    re-estimated every 100 iterations.
+    re-estimated every 10 iterations by default.
 
     :param int estimationWindowSize: size of sliding window of historical
-      data points to maintain for periodic reestimation of the Gaussian. Note:
-      the default of 8640 is based on a month's worth of history at 5-minute
-      intervals.
+    data points to maintain for periodic reestimation of the Gaussian. Note:
+    the default of 8640 is based on a month's worth of history at 5-minute
+    intervals.
 
-    Anomaly likelihood scores are reported at a flat 0.5 for claLearningPeriod +
-    estimationSamples iterations.
+    :param int reestimationPeriod: How often we re-estimate the Gaussian
+    distribution. The ideal is to re-estimate every iteration but this is a
+    performance hit. In general the system is not very sensitive to this number
+    as long as it is small relative to the total number of records processed.
     """
     if estimationWindowSize < estimationSamples:
       raise ValueError("estimationSamples exceeds estimationWindowSize")
@@ -106,11 +112,7 @@ class AnomalyLikelihood(object):
     self._probationaryPeriod = claLearningPeriod + estimationSamples
     self._claLearningPeriod = claLearningPeriod
 
-    # How often we re-estimate the Gaussian distribution. The ideal is to
-    # re-estimate every iteration but this is a performance hit. In general the
-    # system is not very sensitive to this number as long as it is small
-    # relative to the total number of records processed.
-    self._reestimationPeriod = 100
+    self._reestimationPeriod = reestimationPeriod
 
 
   def __eq__(self, o):
@@ -166,12 +168,7 @@ class AnomalyLikelihood(object):
       'settle down'.
     """
     numShiftedOut = max(0, numIngested - windowSize)
-    return max(0, learningPeriod - numShiftedOut)
-
-
-  def forceModelRefresh(self):
-    """Force refresh of the anomaly likelihood model's state"""
-    self._distribution = None
+    return min(numIngested, max(0, learningPeriod - numShiftedOut))
 
 
   def anomalyProbability(self, value, anomalyScore, timestamp=None):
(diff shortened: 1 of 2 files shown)


nupic/algorithms/anomaly_likelihood.py
```diff
@@ -204,18 +201,22 @@ def anomalyProbability(self, value, anomalyScore, timestamp=None):
...
+        [dataPoint],
+        self._distribution)
 
-      likelihoods, _, self._distribution = (
-        updateAnomalyLikelihoods([dataPoint],
-                                 self._distribution)
-      )
       likelihood = 1.0 - likelihoods[0]
 
+      # If anomaly score > 0.99 then we greedily update the statistics.
```
scottpurdy: This says why it is ok but not why we do it in the first place. I believe that it is to mitigate the impact of not updating the distribution at every iteration. It basically says let us only periodically update it but if we have a very high anomaly likelihood then we need it to be accurate so override the periodic update and do an update now.

nupic/algorithms/anomaly_likelihood.py
```diff
@@ -84,18 +88,20 @@ def __init__(self,
...
+    re-estimated every 10 iterations by default.
 
     :param int estimationWindowSize: size of sliding window of historical
-      data points to maintain for periodic reestimation of the Gaussian. Note:
-      the default of 8640 is based on a month's worth of history at 5-minute
-      intervals.
-
-    Anomaly likelihood scores are reported at a flat 0.5 for claLearningPeriod +
-    estimationSamples iterations.
+    data points to maintain for periodic reestimation of the Gaussian. Note:
```
scottpurdy: I find it much easier to include indentation for continued param descriptions but I don't think there is an explicit convention.

nupic/algorithms/anomaly_likelihood.py
```diff
@@ -166,12 +168,7 @@ def _calcSkipRecords(numIngested, windowSize, learningPeriod):
       'settle down'.
     """
     numShiftedOut = max(0, numIngested - windowSize)
-    return max(0, learningPeriod - numShiftedOut)
-
-
-  def forceModelRefresh(self):
-    """Force refresh of the anomaly likelihood model's state"""
-    self._distribution = None
+    return min(numIngested, max(0, learningPeriod - numShiftedOut))
```
scottpurdy: good catch

nupic/algorithms/anomaly_likelihood.py
```diff
@@ -106,11 +112,7 @@ def __init__(self,
     self._probationaryPeriod = claLearningPeriod + estimationSamples
     self._claLearningPeriod = claLearningPeriod
 
-    # How often we re-estimate the Gaussian distribution. The ideal is to
-    # re-estimate every iteration but this is a performance hit. In general the
-    # system is not very sensitive to this number as long as it is small
-    # relative to the total number of records processed.
-    self._reestimationPeriod = 100
+    self._reestimationPeriod = reestimationPeriod
```
scottpurdy: do we need to update __setstate__ to handle old saved instances of this? Or is it not being saved by anyone yet? Probably better to be safe.

nupic/algorithms/anomaly_likelihood.py
vitaly-krugl: Me, too. I was following the convention of the existing arg descriptions. I'll go ahead and fix this up. In reference to yesterday's "clean code" discussion, I noticed that this file is not using the doxygen convention for args, etc.

nupic/algorithms/anomaly_likelihood.py
vitaly-krugl: I caught this once I implemented unit tests for `_calcSkipRecords`. I think that the user of the value handled it gracefully, but it was potentially error-prone and unintuitive.

tests/unit/nupic/algorithms/anomaly_likelihood_test.py
```diff
@@ -610,16 +654,185 @@ def testEstimationWindowSize(self):
...
+        l.anomalyProbability(10, 0.1, timestamp=7)
+        self.assertEqual(estimateAnomalyLikelihoodsWrap.call_count, 2)
+
+
   def testEquals(self):
     l = an.AnomalyLikelihood(claLearningPeriod=2, estimationSamples=2)
     l2 = an.AnomalyLikelihood(claLearningPeriod=2, estimationSamples=2)
     self.assertEqual(l, l2)
 
+    # use 5>2+2 probationary samples to create distribution estimate
```
scottpurdy: This is obviously testing a specific corner case but you could make it easier on the reader by stating what the case is rather than assuming they understand what `5>2+2` is referring to

tests/unit/nupic/algorithms/anomaly_likelihood_test.py
```diff
@@ -610,16 +654,185 @@ def testEstimationWindowSize(self):
     self.assertEqual(len(l._historicalScores), 3)
 
 
+  def testAnomalyLikelihoodWindowSizeImpactOnDistributionEstimateScoreArgs(
```
scottpurdy: These tests are in the class `AnomalyLikelihoodTest` so I think you can leave "AnomalyLikelihood" out of the names.

scottpurdy:
Done with review

nupic/algorithms/anomaly_likelihood.py
vitaly-krugl: @scottpurdy: a couple of comments regarding `__setstate__`:

1. The same member variable `_reestimationPeriod` existed before; if someone saved it having the value of 100, then it would have been acceptable to them, and there is no need to fix it up upon de-serialization.
2. Also, `__setstate__` doesn't exist for this class, anyway. But I think that the above item is what really matters.

tests/unit/nupic/algorithms/anomaly_likelihood_test.py
vitaly-krugl: @scottpurdy: This was an existing comment that I moved to make pylint happy. I will figure out the meaning and improve the comment.

tests/unit/nupic/algorithms/anomaly_likelihood_test.py
vitaly-krugl: @scottpurdy: it's actually more confusing than that: the `AnomalyLikelihoodTest` class contains test methods for the low-level functions as well as for the high-level `AnomalyLikelihood` class. Hence, I added the class name in my test methods to distinguish them from the the test methods that test the low-level functions. Perhaps it would be better to split the tests into two test case classes - something like `AnomalyLikelihoodClassTest` and AnomalyLikelihoodLowLevelTest`? Any suggestions?

nupic/algorithms/anomaly_likelihood.py
scottpurdy: So old saved versions would still work fine? Sounds good.

scottpurdy:
:+1:

scottpurdy:
Please update the years in the copyright headers as well

nupic/algorithms/anomaly_likelihood.py
vitaly-krugl: I think that the old saved versions should work fine.

vitaly-krugl:
Updated the copyright years

vitaly-krugl:
@scottpurdy: I addressed all your feedback via 3ecdd730f219c6a9c04374f254741af4c42edc98. I will merge this PR as soon as the CI build completes and is green.

vitaly-krugl:
Merging based on passing Linux builds/tests. Travis is presently having an issue with Mac OS X builds and the ETA is unknown. These changes are not platform-specific, and were also manually tested on Mac OS X.

PR #2550, opened 2015-09-03
Wrap sparse matrix implementations with cortical column-centric seman…

oxtopus:
…tics as a way to abstract away the underlying implementation

cc @scottpurdy

### DIFF
--- a/src/nupic/research/spatial_pooler.py
+++ b/src/nupic/research/spatial_pooler.py
@@ -28,6 +28,7 @@ from nupic.bindings.math import (SM32 as SparseMatrix,
                                  Random as NupicRandom)
 
 
+
 realDType = GetNTAReal()
 uintType = "uint32"
 
@@ -37,6 +38,57 @@ VERSION = 2
 
 class InvalidSPParamValueError(ValueError):
   """The user passed an invalid value for a SpatialPooler parameter
+  """
+  pass
+
+
+
+class _SparseMatrixCorticalColumnAdapter(object):
+  """ Many functions in SpatialPooler operate on a columnIndex but use an
+  underlying storage implementation based on a Sparse Matrix in which cortical
+  columns are represented as rows.  This can be confusing to someone trying to
+  follow the algorithm, confusing terminology between matrix math and cortical
+  columns.  This class is provided to abstract away some of the details of the
+  underlying implementation, providing a cleaner API that isn't specific to
+  sparse matrices.
+  """
+
+  def __getitem__(self, columnIndex):
+    """ Wraps getRow() such that instances may be indexed by columnIndex.
+    """
+    return super(_SparseMatrixCorticalColumnAdapter, self).getRow(columnIndex)
+
+
+  def replace(self, columnIndex, bitmap):
+    """ Wraps replaceSparseRow()
+    """
+    return super(_SparseMatrixCorticalColumnAdapter, self).replaceSparseRow(
+      columnIndex, bitmap
+    )
+
+
+  def update(self, columnIndex, vector):
+    """ Wraps setRowFromDense()
+    """
+    return super(_SparseMatrixCorticalColumnAdapter, self).setRowFromDense(
+      columnIndex, vector
+    )
+
+
+
+class CorticalColumns(_SparseMatrixCorticalColumnAdapter, SparseMatrix):
+  """ SparseMatrix variant of _SparseMatrixCorticalColumnAdapter.  Use in cases
+  where column connections are represented as float values, such as permanence
+  values
+  """
+  pass
+
+
+
+class BinaryCorticalColumns(_SparseMatrixCorticalColumnAdapter,
+                            SparseBinaryMatrix):
+  """ SparseBinaryMatrix variant of _SparseMatrixCorticalColumnAdapter.  Use in
+  cases where column connections are represented as bitmaps.
   """
   pass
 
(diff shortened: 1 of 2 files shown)


nupic/research/spatial_pooler.py
```diff
@@ -42,6 +43,57 @@ class InvalidSPParamValueError(ValueError):
 
 
 
+class _SparseMatrixCorticalColumnAdapter(object):
+  """ Many functions in SpatialPooler operate on a columnIndex but use an
+  underlying storage implementation based on a Sparse Matrix in which cortical
+  columns are represented as rows.  This can be confusing to someone trying to
+  follow the algorithm, confusing terminology between matrix math and cortical
+  columns.  This class is provided to abstract away some of the details of the
```
scottpurdy: Can you use `abc` to prevent this from being instantiable?

nupic/research/spatial_pooler.py
```diff
@@ -42,6 +43,57 @@ class InvalidSPParamValueError(ValueError):
...
+    """ Wraps setRowFromDense()
+    """
+    return super(_SparseMatrixCorticalColumnAdapter, self).setRowFromDense(
+      columnIndex, vector
+    )
+
+
+
+class CorticalColumns(_SparseMatrixCorticalColumnAdapter, SparseMatrix):
+  """ SparseMatrix variant of _SparseMatrixCorticalColumnAdapter.  Use in cases
```
scottpurdy: nit: Only use one space after periods. Here and below a few lines

scottpurdy:
If we are essentially aliasing the generic data structure to make it algorithm specific, should we take it even further? Give the three use cases classes with class names that match the use and method names like `updateColumnPermanences` rather than the more general `update`? I don't have a strong opinion

nupic/research/spatial_pooler.py
```diff
@@ -42,6 +43,57 @@ class InvalidSPParamValueError(ValueError):
...
+  """
+
+  def __getitem__(self, columnIndex):
+    """ Wraps getRow() such that instances may be indexed by columnIndex.
+    """
+    return super(_SparseMatrixCorticalColumnAdapter, self).getRow(columnIndex)
+
+
+  def replace(self, columnIndex, bitmap):
+    """ Wraps replaceSparseRow()
```
breznak: maybe duplicate the doc and explain what it does?

nupic/research/spatial_pooler.py
```diff
@@ -246,31 +298,30 @@ def __init__(self,
...
-    # class, to reduce memory footprint and computation time of algorithms that
-    # require iterating over the data structure.
-    self._potentialPools = SparseBinaryMatrix(numInputs)
+    # connected to inputs in its potential pool.  Here, BinaryCorticalColumns
+    # is used to provide cortical column-centric semantics for what is
+    # otherwise a sparse binary matrix implementation.  Sparse binary matrix is
+    # used as an optimization since a column will only be connected to a small
+    # fraction of input bits.
+    self._potentialPools = BinaryCorticalColumns(numInputs)
     self._potentialPools.resize(numColumns, numInputs)
```
breznak: can we already initialize this with numColumns, numInputs? as here https://github.com/numenta/nupic/pull/2550/files#diff-323276e3bd86539a560868bd2961f1d7R324 ?

breznak:
pleasure for the eyes! :) 

OT: why do we use SP with bindings to a c++ code (SparseBinaryMatrix) when we have `nupic.bindings.algorithms.SpatialPooler` ? Could we just drop all this *Matrix code?

cogmission:
I've been meaning to ask... What does "OT:" mean? :-)

breznak:
>  I've been meaning to ask... What does "OT:" mean? :-)

I've been using it for ages as off-topic, hope it is used so :)

cogmission:
@breznak Ah. ok. @rhyolight  says I'm kind of behind on "web lingo" so I wouldn't know if that's the way it is used... :-)

nupic/research/spatial_pooler.py
oxtopus: How would you do that?

Only way I can think of is to add some sort of hack in `__new__()` or `__init__()` to check `__bases__` or compare `self.__class__`.

You could make `__init__()` an abstract method, which would result in a failure if someone attempts to instantiate it directly.

Instead, maybe better to assert int `__init__()` that `super(_SparseMatrixCorticalColumnAdapter, self)` has the attrs that it expects, like `getRows`, `replaceSparseRow`, etc.?

oxtopus:
re:

> If we are essentially aliasing the generic data structure to make it algorithm specific, should we take it even further? Give the three use cases classes with class names that match the use and method names like updateColumnPermanences rather than the more general update

That was the approach I initially took, but realized that the implementations were identical.  If we were to go as far as you describe, there's no need for classes.  Just create a function called `updateColumnPermanences()`, etc.

In the end, I recognized there were two basic variants: binary vs scalar.  I wanted the user to have to choose between the two depending on use-case and have only a common underlying implementation wrt abstracting away underlying storage, which is how I arrived at the multi-inheritance approach here.

nupic/research/spatial_pooler.py
scottpurdy: Yeah it looks like `abc` only prevents it from being instantiated if some abstract method hasn't been overriden. I suppose fine as is.

oxtopus:
re:

> why do we use SP with bindings to a c++ code (SparseBinaryMatrix) when we have nupic.bindings.algorithms.SpatialPooler ?

We want a python implementation of the algorithm which is distinct from the c++ implementation, but still want to take advantage of optimizations where available.

scottpurdy:
Sounds fine to me. Let's have @chetan51 review as well to make sure he agrees with the direction.

chetan51:
We are eventually going to replace the data structure in Spatial Pooler with the Connections class right? Until then, this looks more readable to me :+1:

scottpurdy:
@oxtopus - Can you merge master so we can get this wrapped up?

PR #2577, opened 2015-09-16
Use logging.debug for emitting the message about not being able to import matplotlib

vitaly-krugl:
Use logging.debug for emitting the message about not being able to import matplotlib; we log it at debug level to avoid polluting the logs of apps and services that don't care about plotting.

The traceback was ending up in logs (e.g., htmengine's model_scheduler.log), waisting time to investigate the traceback.


vitaly-krugl:
@chetan51: please review

chetan51:
:+1:

PR #2588, opened 2015-09-21
Issue 2358 enable test hotgym regression

### LINKED ISSUE #2358: Bug: Hotgym metric results differ on Travis and on OS X
While resolving #2322 we discovered that the metric results vary from OS X to Travis. We need to track down the cause of this and fix it.

oxtopus:
Fixes #2358


scottpurdy:
:+1:

PR #2617, opened 2015-09-30
Add cluster params selection module from numenta-apps htmegine

### LINKED ISSUE #2619: Move model selection utilities from HTM Engine into NuPIC
As described by an [internal JIRA task](https://jira.numenta.com/browse/ENG-73):

Presently, unicorn's backend relies on `getScalarMetricWithTimeOfDayParams()` from htmengine, which would be better moved into NuPIC.

```python
from htmengine.algorithms.modelSelection.clusterParams import (
  getScalarMetricWithTimeOfDayParams)
```

The plan it to move the `algorithms.modelSelection*` logic and data files somewhere into nupic, where they may be shared by others:

1. move `htmengine/algorithms/modelSelection/*` to `src/nupic/frameworks/opf/common_models/`
1. Another issue is the way it loads the params from disk. https://github.com/numenta/numenta-apps/blob/master/htmengine/htmengine/algorithms/modelSelection/clusterParams.py#L80 does not work when the package is not expanded.
  - Use https://pythonhosted.org/setuptools/pkg_resources.html#resource-extraction because the package will be inside a zip file
  - or pkg_resources.resource_stream

vitaly-krugl:
Added src/nupic/frameworks/opf/common_models/cluster_params.py and supporting files from numenta-apps htmengine.

A separate numenta-apps PR will remove this code from htmengine.

Fixes #2619.

### DIFF
--- a/src/nupic/frameworks/opf/common_models/cluster_params.py
+++ b/src/nupic/frameworks/opf/common_models/cluster_params.py
@@ -1,0 +1,130 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2015, Numenta, Inc.  Unless you have purchased from
+# Numenta, Inc. a separate commercial license for this software code, the
+# following terms and conditions apply:
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero Public License for more details.
+#
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+import json
+import numpy as np
+import os
+from pkg_resources import resource_stream
+
+
+
+def getScalarMetricWithTimeOfDayAnomalyParams(metricData,
+                                              minVal=None,
+                                              maxVal=None,
+                                              minResolution=None):
+  """
+    Return a dict that can be used to create an anomaly model via OPF's
+    ModelFactory.
+
+    :param metricData: numpy array of metric data. Used to calculate minVal
+      and maxVal if either is unspecified
+
+    :param minVal: minimum value of metric. Used to set up encoders. If None
+      will be derived from metricData.
+
+    :param maxVal: maximum value of metric. Used to set up input encoders. If
+      None will be derived from metricData
+
+    :param minResolution: minimum resolution of metric. Used to set up
+      encoders.  If None, will use default value of 0.001.
+
+    :returns: a dict containing "modelConfig" and "inferenceArgs" top-level
+      properties. The value of the "modelConfig" property is for passing to
+      the OPF `ModelFactory.create()` method as the `modelConfig` parameter. The
+      "inferenceArgs" property is for passing to the resulting model's
+      `enableInference()` method as the inferenceArgs parameter. NOTE: the
+      timestamp field corresponds to input "c0"; the predicted field corresponds
+      to input "c1".
+    :rtype: dict
+
+    Example:
+      from nupic.frameworks.opf.modelfactory import ModelFactory
+      from nupic.frameworks.opf.common_models.cluster_params import (
+        getScalarMetricWithTimeOfDayAnomalyParams)
+
+      params = getScalarMetricWithTimeOfDayAnomalyParams(
+        metricData=[0],
+        minVal=0.0,
+        maxVal=100.0)
+
+      model = ModelFactory.create(modelConfig=params["modelConfig"])
+      model.enableLearning()
+      model.enableInference(params["inferenceArgs"])
+  """
+  # Default values
+  if minResolution is None:
+    minResolution = 0.001
+
+  # Compute min and/or max from the data if not specified
+  if minVal is None or maxVal is None:
+    compMinVal, compMaxVal = _rangeGen(metricData)
+    if minVal is None:
+      minVal = compMinVal
+    if maxVal is None:
+      maxVal = compMaxVal
+
+  # Handle the corner case where the incoming min and max are the same
+  if minVal == maxVal:
+    maxVal = minVal + 1
+
+  # Load model parameters and update encoder params
+  paramFileRelativePath = os.path.join(
+    "anomaly_params_random_encoder",
+    "best_single_metric_anomaly_params.json")
+
+  with resource_stream(__name__, paramFileRelativePath) as infile:
+    paramSet = json.load(infile)
+
+  _fixupRandomEncoderParams(paramSet, minVal, maxVal, minResolution)
+
+  return paramSet
+
+
+
+def _rangeGen(data, std=1):
+  """
+  Return reasonable min/max values to use given the data.
+  """
+  dataStd = np.std(data)
+  if dataStd == 0:
+    dataStd = 1
+  minval = np.min(data) -  std * dataStd
+  maxval = np.max(data) +  std * dataStd
+  return minval, maxval
+
+
+
+def _fixupRandomEncoderParams(params, minVal, maxVal, minResolution):
+  """
+  Given model params, figure out the correct parameters for the
+  RandomDistributed encoder. Modifies params in place.
+  """
+  encodersDict = (
+    params["modelConfig"]["modelParams"]["sensorParams"]["encoders"]
+  )
+
+  for encoder in encodersDict.itervalues():
+    if encoder is not None:
+      if encoder["type"] == "RandomDistributedScalarEncoder":
+        resolution = max(minResolution,
+                         (maxVal - minVal) / encoder.pop("numBuckets")
+                        )
+        encodersDict["c1"]["resolution"] = resolution
--- a/src/nupic/frameworks/opf/common_models/__init__.py
+++ b/src/nupic/frameworks/opf/common_models/__init__.py
@@ -1,0 +1,20 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2015, Numenta, Inc.  Unless you have purchased from
+# Numenta, Inc. a separate commercial license for this software code, the
+# following terms and conditions apply:
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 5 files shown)


vitaly-krugl:
@scottpurdy: can we fast-track this change? It's urgently needed for numenta-apps/unicorn.

@lscheinkman Please review the use of resource_stream. Is it what you had in mind?

vitaly-krugl:
CC @marionleborgne

src/nupic/frameworks/opf/common_models/cluster_params.py
```diff
@@ -0,0 +1,139 @@
...
+from pkg_resources import resource_stream
+
+
+def setRandomEncoderParams(params, minVal, maxVal, minResolution):
+  """
+  Given model params, figure out the correct parameters for the
+  RandomDistributed encoder. Modifies params in place.
+  """
+  encodersDict= (
+    params['modelConfig']['modelParams']['sensorParams']['encoders']
```
oxtopus: Should use `'` per [Numenta Python Style Guide](https://github.com/numenta/nupic/wiki/Python-Style-Guide#quotes)

src/nupic/frameworks/opf/common_models/cluster_params.py
```diff
@@ -0,0 +1,139 @@
...
+    )
+  for v in encodersDict.itervalues():
```
oxtopus: Should use a better, more descriptive variable than `v`.

src/nupic/frameworks/opf/common_models/cluster_params.py
```diff
@@ -0,0 +1,139 @@
...
+    if v is not None:
+      if v['type'] == 'RandomDistributedScalarEncoder':
+        resolution = max(minResolution,
+                         (maxVal - minVal) / v.pop('numBuckets')
```
oxtopus: Is there a risk of `v.pop('numBuckets')` returning `0`?  For that matter, `v.pop('numBuckets')` is an odd construct in this context, since it's destructive.

src/nupic/frameworks/opf/common_models/cluster_params.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+
+def getScalarMetricWithTimeOfDayParams(metricData,
+                                       minVal=None,
+                                       maxVal=None,
+                                       minResolution=None,
+                                       useRandomEncoder=True):
+  """
+    Return an ordered list of JSON strings that can be used to create an
```
oxtopus: Docstring doesn't match any format with which I'm familiar.

src/nupic/frameworks/opf/common_models/cluster_params.py
```diff
@@ -0,0 +1,139 @@
...
+  """
+  # Default values
+  if minResolution is None:
+    minResolution = 0.001
+
+  # Read a sorted list of parameters from the appropriate directory
+  if useRandomEncoder:
+    paramsDirectory = 'anomaly_params_random_encoder'
+  else:
+    # This will go away soon MER-1442
```
oxtopus: Is this comment still valid?

src/nupic/frameworks/opf/common_models/cluster_params.py
```diff
@@ -0,0 +1,139 @@
...
+    paramsDirectory = 'scalarMetricWithTimeOfDayAnomalyParams'
+
+  with resource_stream('nupic.frameworks.opf.common_models',
+                       os.path.join(paramsDirectory,
+                                    'paramOrder.csv')) as fileObj:
+    paramFiles = csv.reader(fileObj).next()
```
oxtopus: Better to call `next(csv.reader(fileObj))` per convention.

src/nupic/frameworks/opf/common_models/cluster_params.py
```diff
@@ -0,0 +1,139 @@
...
+
+  paramSets=[]
```
oxtopus: Spacing.

src/nupic/frameworks/opf/common_models/cluster_params.py
```diff
@@ -0,0 +1,139 @@
...
+                         os.path.join(paramsDirectory, p)) as infile:
+      paramSet=json.load(infile)
+
+    encodersDict= (
+      paramSet['modelConfig']['modelParams']['sensorParams']['encoders'])
+
+    if useRandomEncoder:
+      setRandomEncoderParams(paramSet, minVal, maxVal, minResolution)
+    else:
+      # TODO Update per MER-1442
```
oxtopus: Is this comment still valid?

src/nupic/frameworks/opf/common_models/cluster_params.py
```diff
@@ -0,0 +1,139 @@
...
+  return paramSets
+
+
+
+def rangeGen(data, std=1):
+  """
+  Return reasonable min/max values to use given the data.
+  """
+  dataStd = np.std(data)
+  if dataStd==0:
```
oxtopus: Spacing

oxtopus:
LGTM, but there are a number of style-guide violations.

src/nupic/frameworks/opf/common_models/cluster_params.py
```diff
@@ -0,0 +1,139 @@
...
+  # Read a sorted list of parameters from the appropriate directory
+  if useRandomEncoder:
+    paramsDirectory = 'anomaly_params_random_encoder'
+  else:
+    # This will go away soon MER-1442
+    paramsDirectory = 'scalarMetricWithTimeOfDayAnomalyParams'
+
+  with resource_stream('nupic.frameworks.opf.common_models',
+                       os.path.join(paramsDirectory,
+                                    'paramOrder.csv')) as fileObj:
```
rhyolight: @vitaly-krugl Won't you also need to add a new entry in the `setup.py`'s `package_data` config so this data gets bundled with releases? Something like:

```python
    package_data={
      "nupic.frameworks.opf.common_models": ["*.json", "*.csv"]
    },
```

vitaly-krugl: @rhyolight, that thought came to my mind as I was driving home last night, and I made a mental note to check on it this morning. This didn't get caught, since I tested with nupic installed via `setup.py develop`.

vitaly-krugl:
Regarding

> LGTM, but there are a number of style-guide violations.

@oxtopus, thanks for the thorough review and feedback. For the record, this is not my code. The goal of the task was to relocate https://github.com/numenta/numenta-apps/blob/1ff572a21a5c27fd290822e572ce33f42e1ee19e/htmengine/htmengine/algorithms/modelSelection/clusterParams.py to the nupic repo pronto to resolve issues faced by numenta-apps/unicorn.

src/nupic/frameworks/opf/common_models/cluster_params.py
vitaly-krugl: @oxtopus: not sure what you mean here: they are all `'` already. Could you clarify?

oxtopus: I mis-typed.  It should be `"`, as described in the linked style guide.  I've since changed the original comment.

oxtopus:
@vitaly-krugl Feel free to merge https://github.com/vitaly-krugl/nupic/pull/1 into this branch to fix the violations

src/nupic/frameworks/opf/common_models/cluster_params.py
vitaly-krugl: @oxtopus: based on the docstring comment `Modifies params in place`, this destructive action is not unexpected. My guess regarding `setRandomEncoderParams` is that it should have been a private function, and I am also guessing that a value of 0 would constitute a bug in the cluster-params generation logic.

@subutai: could you please take a look and comment regarding  making `setRandomEncoderParams` private and also about the circumstances whereby `v.pop('numBuckets')` might return 0?

src/nupic/frameworks/opf/common_models/cluster_params.py
vitaly-krugl: @oxtopus: I had the same question yesterday, so I looked up MER-1442 and found that it's still open in @subutai' name. So, I added a comment in MER-1442 to let him know that the code is being moved. Subutai acknowledged my comment and didn't indicate that this change is no longer needed.

src/nupic/frameworks/opf/common_models/cluster_params.py
vitaly-krugl: There are a few of these here. I really wish that pylint would alert about these, but it doesn't, at least not with our current configuration.

src/nupic/frameworks/opf/common_models/cluster_params.py
vitaly-krugl: I think so. See my reply in the other place where you asked about MER-1442.

vitaly-krugl:
@oxtopus, thanks for https://github.com/vitaly-krugl/nupic/pull/1. I will take a look.

subutai:
@vitaly-krugl Much of this is from a really old set of files. At that time there was a chance Grok would have to choose the best parameter from a set of possible parameters, aka mini-swarm. We have never had a need for this, and only ever use one set of parameters. In addition the old cluster params are probably not even valid anymore as we have changed the algorithm implementations. As such, I think we can greatly simplify this PR. Instead of 10 JSON files and logic to switch between scalar/random encoders, we can just have one JSON file and no such switching logic.

src/nupic/frameworks/opf/common_models/cluster_params.py
subutai: > could you please take a look and comment regarding making setRandomEncoderParams private and also about the circumstances whereby v.pop('numBuckets') might return 0?

The chance of numBuckets being 0 was 0 since it taken from a known set of hard coded values.   As far as the function being private I think that's fine if it is only supposed to be used from here.

src/nupic/frameworks/opf/common_models/anomaly_params_random_encoder/5AnomalyEnergyWithTimeCentroids_1_params.json
```diff
@@ -0,0 +1,97 @@
+{
```
subutai: How about renaming this to `best_single_metric_anomaly_params.json`?

src/nupic/frameworks/opf/common_models/cluster_params.py
vitaly-krugl: @rhyolight: `package_data` may be superfluous in nupic's setup. My initial error was that there was no glob for my csv file. Subsequently, I refactored the PR, removing the .csv file altogether, while leaving setup.py as-is. My code was still loading a .json file from "nupic.frameworks.opf.common_models/anomaly_params_random_encoder". This json file is covered by the rule `recursive-include src/nupic *.capnp *.json *.tpl *.txt *.xml` in MANIFEST.in, but not covered by any rule in `package_data`. Despite this omission from `package_data`, all tests passed, and I verified that the tar.gz file generatred by a local sdist build includes that .json file.

vitaly-krugl:
@scottpurdy and @oxtopus: Need urgent re-review of the simplified code.

oxtopus:
:+1:

vitaly-krugl:
Thank you @oxtopus!

Issue #824, opened 2014-04-16
Refactor boosting/inhibition to stand-alone objects


breznak:
edit:
we discussed this with @chetan51 back then, I will try to investigate this issue again - if it is possible to extract Boosting/Inhibition functionality from SP/TP to stand-alone classes. 
pros:
* simpler SP/TP/(TM?) code
* code shared among implementations
* code shared among different classes SP/TP,...

cons:
* there are (easy) optimizations for some cases -- with right data-structure, we can optimize (using eg GPU lib) for all cases

----------------------
I think we should (again) create a class `Boosting` (or BoostingStrategy ?) that would take care of the logic of: 
when should it start boosting? 
which columns? 
and how "strongly" to boost? 

Then we could easily have implementations: GreedyBoosting (now), SmoothBoosting, PartialBoosting (1 at a time)...

rhyolight:
Discussing [on this HTM Forum thread](https://discourse.numenta.org/t/should-boosting-and-or-inhibition-code-be-pulled-out-into-objects/1000/2).

rhyolight:
This idea was rejected. [See the forum thread for discussion. ](https://discourse.numenta.org/t/should-boosting-and-or-inhibition-code-be-pulled-out-into-objects/1000)

Issue #823, opened 2014-04-16
Boosting creates disturbance in anomaly detection


breznak:
As found out by Ritchie, on a large CLA, and relatively simple problem (so columns are unused) (unexpected) spikes appear with anomaly detection after some time. 

I've found out it's related to several columns being (unexpectedly) activated because of boosting, this causes anomaly spike. 

Proposed **solutions** are: 

1/ remember the boosted columns and for some time window (eg 10 steps) ignore them from the anomaly computations (mark them as inactive for the formula above). This will let the columns settle, but still can create an anomaly at the end of the window. 

(I think it's safe to ignore them from the computation as those cols were unused before anyway)

1.B/ before the boost, mark the selected set of columns as predicted. 

2/ small change at a time: 

say we want to boost 20 columns. boost them the normal way, but stretch it over 20 iterations of the CLA - boosting 1 at a time. This will avoid anomaly. 

(? could it make boosting ineffective in other cases? because it'd take longer time)

3/ be smooth 
boost the output of the columns selected for boosting with a nice curve (0.1 0.3 0.75 1.0 1.0 0.75 0.3 0.1) (column's output*(1+curve_coef)) over time.

(i seem to prefer 2/3 for being less hacky)

Testing could be done: 

existing scenarios should not get (much) worse. And Ritchie provided a triggering example: large CLA, relatively simple problem, over time boosting kicks in and causes spikes (anomalies :P ) in anomaly detection.  

TODO:
- [ ] cleanup/separate boosting logic #824 
- [ ] submit boosting disturbance test https://github.com/numenta/nupic/issues/2634
- [ ] boosting fixed
  - [ ] discuss the solutions above before choosing one for implementation

CC @subutai @chetan51 @h2suzuki

breznak:
This is an example what adversary effect the boosting disturbances have: 

(iteration, anomaly score) on constant(!) fn, using SP only.
```
42 0.0
43 0.0
44 0.0
45 0.0
46 0.0
47 0.0
48 0.0
49 /////////////// BOOST ///////////////////
0.0
50 0.0240963855422
51 0.120481927711
52 0.120481927711
53 0.120481927711
54 0.120481927711
55 0.120481927711
56 0.120481927711
57 0.130952380952
58 0.141176470588
59 0.141176470588
60 0.120481927711
61 0.141176470588
62 0.179775280899
63 0.188888888889
64 0.120481927711
65 0.139534883721
66 0.0
```

[...]

PR #2620, opened 2015-09-30
[RFC] SpatialPooler: extract boosting to a separate file, "interface"

breznak:
Addresses boosting in the SpatialPooler, and tries to put all the boosting related logic into one class. The `Boosting` instance is then used as a member of SP.
It's only change and improvement is that it hides all the `update*DutyCycles*` logic from SP. 

TODO: 
- [ ] looking for help with `CAPNP` serialization for Boosting (and SP) https://github.com/numenta/nupic/pull/2620#issuecomment-149392750

Fixes: #824


breznak:
@chetan51 we've discussed a similar change in the past, I think this is a compromise between having the logic conveniently in SP, and separating variables/methods into blocks by their functionality.What do you think about this?

chetan51:
This is a good initiative. It's not clear to me what's the best way to organize this kind of change (this might be the best way, but I'm not yet sure). @scottpurdy do you have any thoughts on it?

scottpurdy:
I don't think it is worth splitting out simply to shorten the file size of spatial_pooler.py. As far as architecturally, boosting doesn't seem very useful for things other than spatial pooling so it makes sense to define it there.

With or without this change, the process for someone to try their own boosting method would be the same: subclass SpatialPooler and override `_updateBoostFactors`.

scottpurdy:
@chetan51 - Seems like you see some value to something like this change. Perhaps you can share the specific benefits you think we can get?

chetan51:
@scottpurdy It makes the Spatial Pooler code easier to read and makes the code easier to debug when you don't use boosting.

scottpurdy:
@chetan51 - Spatial pooling is pretty reliant on boosting to form good representations so I'd consider no-boosting a pretty niche case.

scottpurdy:
Could the boosting class be a utility class rather than a base class? That seems like a cleaner encapsulation of the behavior.

chetan51:
@scottpurdy Is it? Oftentimes I disable boosting because it messes with the results, since it kicks in suddenly.

I agree, it could be a utility class instead.

scottpurdy:
@chetan51 - It sounds like we should ignore boost factors when calculating raw anomaly scores.

breznak:
>  don't think it is worth splitting out simply to shorten the file size of spatial_pooler.py. ... to try their own boosting method would be the same: subclass SpatialPooler and override _updateBoostFactors.

@scottpurdy I also think it's improved readability of the code, boosting stretches to a lot of functions/variables/LOC over the code and I remember being pretty confused reading the SP code for the first time, those methods should have had a "# Don't need to understand, boosting dutycycle counting related code." tags. True that boosting is only used in SP, but it is a well defined functionality that fetches a good number of LOC, so it seems to me that it could be in a separate file/class. 

For the boosting overriding, yes, but imho we should be more generic. I can come up with a boosting method that does not use the `duty*Cycles` variables at all, so why carry the code? SP should have `boostColumns()` and `updateBoosting()` methods, that's all. 

> Could the boosting class be a utility class rather than a base class? That seems like a cleaner encapsulation of the behavior.

What is a utility class? At best I'd consider boosting in a separate file and SP having a `self.boosting = Boosting()` member that provides the two methods mentioned above: `boostColumns()` and `updateBoosting()`. But there were some objections to this in the past, so I figured this "inheritance" thing as the least (code) invasive. 

> It sounds like we should ignore boost factors when calculating raw anomaly scores.

you are skipping here :) That's my intended follow-up after I've cleaned the ground here and why I've started this PR. Some possible solutions are: ignoring boosted as you've suggested, smooth boosting by smaller factors, or separating the number of boosted columns into smaller amounts at a time, see https://github.com/numenta/nupic/issues/823

breznak:
Ping...

I have some other changes based off this, would like to resolve this PR. If there are no objections, I'd prefer boosting separated to a file like here. I don't have an opinion if this should be done as "includes" through inheritance (current state of this PR), or as a "part of" relationship where boosting is an instance of the Boosting class and self.boosting is a member of SP.

@scottpurdy @chetan51 @rhyolight

cogmission:
I think this is a great idea. It "outboards" the requisite state necessary
to maintain boosting parameters.

On Thu, Oct 8, 2015 at 6:10 AM, breznak <notifications@github.com> wrote:

> Ping...
>
> I have some other changes based off this, would like to resolve this PR.
> If there are no objections, I'd prefer boosting separated to a file like
> here. I don't have an opinion if this should be done as "includes" through
> inheritance (current state of this PR), or as a "part of" relationship
> where boosting is an instance of the Boosting class and self.boosting is a
> member of SP.
>
> @scottpurdy <https://github.com/scottpurdy> @chetan51
> <https://github.com/chetan51> @rhyolight <https://github.com/rhyolight>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/numenta/nupic/pull/2620#issuecomment-146506364>.
>



-- 
*With kind regards,*

David Ray
Java Solutions Architect

*Cortical.io <http://cortical.io/>*
Sponsor of:  HTM.java <https://github.com/numenta/htm.java>

d.ray@cortical.io
http://cortical.io

breznak:
This PR consists of the relocation and a few small fixes, please review on per commit basis (it's only 4 commits)

scottpurdy:
@chetan51 - Can you take this one?

chetan51:
> At best I'd consider boosting in a separate file and SP having a `self.boosting = Boosting()` member that provides the two methods mentioned above: `boostColumns()` and `updateBoosting()`.

For now, I think this is the best way to separate out the boosting logic. Now that I think about it, it doesn't make sense for boosting logic to be in a base class that SP inherits from.

However, aren't there clients of SP that call functions like `getBoostFactors`? If Boosting was an instance member of SP, would we need to implement those functions in SP and call out to the Boosting instance member?

breznak:
> For now, I think this is the best way to separate out the boosting logic.
Now that I think about it, it doesn't make sense for boosting logic to be
in a base class that SP inherits from.

When SP inherits from Boosting, some things are hackily easier: the
getters/setters for associated boosting variables can be in the boosting
file, while they stay in fact SP's methods. Then there's
`getNearestNeighbors()` which is shared for boosting & inhibition. Simalar
is for variables like `_iterationNum` or serialization.

> However, aren't there clients of SP that call functions like
getBoostFactors?

I think not, so far I've found it's only SP related.

> If Boosting was an instance member of SP, would we need to implement
those functions in SP and call out to the Boosting instance member?

I;m not sure what you mean here, eg. the getters would have to change to
`getSome(): return self.boost._some`;


All in all, I'm changing my mind here a bit and I'd vote for keeping the
inheritance approach for now as the least invasive. And we can open issue
for the "is a part of" way.


Thank you for review @chetan51 and let me know what you think?

chetan51:
@breznak Though the the inheritance way is less invasive due to (as you mentioned) "hacky" sharing of functions like boosting getters and setters, I don't think that is the right way to move forward. It just doesn't make sense to have SP inherit from a Boosting class. I recommend having SP contain an instance of a Boosting class, and delegate the boosting getters and setters to that instance. Later on we can have anyone that calls SP.boostingGetter to instead call SP.boosting.getter.

rhyolight:
On Thu, Oct 8, 2015 at 3:44 PM, Chetan Surpur <notifications@github.com>
wrote:

> I recommend having SP contain an instance of a Boosting class, and
> delegate the boosting getters and setters to that instance.


+1 to composition.

---------
Matt Taylor
OS Community Flag-Bearer
Numenta

cogmission:
+1 to Matt's +1 - Why "muddy" this or any future SP's inheritance pedigree?

breznak:
@cogmission @rhyolight @chetan51 I've updated boosting to be a composition member of SP as you've suggested. Please give the code a review. 

@scottpurdy Can you help me please to bring the capnp serialization from SP over to Boosting? I could probably do that but it will take me a lot of time as I'm unable to do the advanced stuff in CAPNP.

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,270 @@
...
+  In this scenario, the column has a good synaptic overlap over the input,
+  but it is very similar to the input field of the neighboring column(s). 
+  It results in a SDR that is not well (D)istributed, and we are wasting 
+  resources on the columns activity, as it is duplicit to others'. This leads
+  to the column never being active (winning in inhobotion). 
+  In this implementation the fix is to temporarily artificially boost 
+  the column's "importance" by a boosting factor, thus giving it a chance to
+  win in some scenario and creating it's niche. 
+  This process is computed from the '*ActivityDutyCycles' and implemented in the
+  'getBoostingFactors()' method.
```
breznak: @chetan51 @subutai can you please give a review to the documentation above? It is how I understood the current boosting process. And if it is correct, it is what I feel is the wrong approach about boosting - just forcing bigger boost constants, instead of killing off the useless synapses and regrowing new elsewhere. I'd keep this for my reimplementation of boosting.

scottpurdy:
I'm happy to help out with serialization if the organizational changes are approved.

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,270 @@
...
+  def __init__(self,
+               numColumns,
+               maxBoost,
+               minPctOverlapDutyCycle,
+               minPctActiveDutyCycle,
+               dutyCyclePeriod):
+    """
+    construct a boosting object for the SpatialPooler
+
+    @param numColumns -- these all parameters are taken from the SP, please
```
chetan51: Replace with `(See `SpatialPooler` for descriptions of these parameters.)`

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,270 @@
...
+    self._minActiveDutyCycles = numpy.zeros(numColumns, dtype=realDType)
+    self._activeDutyCycles = numpy.zeros(numColumns, dtype=realDType)
+    self._overlapDutyCycles = numpy.zeros(numColumns, dtype=realDType)
+    self._minOverlapDutyCycles = numpy.zeros(numColumns, dtype=realDType)
+    self._minPctOverlapDutyCycles = minPctOverlapDutyCycle
+    self._minPctActiveDutyCycles = minPctActiveDutyCycle
+    self._dutyCyclePeriod = dutyCyclePeriod
+
+
+# Setters/getters for SP parameters from boosting:
```
chetan51: Remove unnecessary comment

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,270 @@
...
+
+
+# Boosting logic methods:
```
chetan51: Remove unnecessary comment

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,270 @@
...
+
+  def getBoostingSuppressedColumns(self):
```
chetan51: Rename function to `getBoostFactors` for consistency.

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,270 @@
...
+      perm = numpy.array([sp.getInputDimensions()], dtype=realDType)
+      sp.getPermanence(columnIndex, perm)
+      pot = numpy.array([sp.getInputDimensions()], dtype=realDType)
+      sp.getPotential(columnIndex, pot)
+      maskPotential = numpy.where(pot > 0)[0]
+      perm[maskPotential] += self._synPermBelowStimulusInc
+      sp._updatePermanencesForColumn(perm, columnIndex, raisePerm=False)
+
+
+# Time keeping methods, counters:
```
chetan51: Remove unnecessary comment

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,270 @@
...
+    Updates the minimum duty cycles in a global fashion. Sets the minimum duty
+    cycles for the overlap and activation of all columns to be a percent of the
+    maximum in the region, specified by minPctOverlapDutyCycle and
+    minPctActiveDutyCycle respectively. Functionality it is equivalent to
+    _updateMinDutyCyclesLocal, but this function exploits the globality of the
+    computation to perform it in a straightforward, and more efficient manner.
+    """
+    self._minOverlapDutyCycles.fill(
+        self._minPctOverlapDutyCycles * self._overlapDutyCycles.max()
+      )
```
chetan51: Align closing parenthesis

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,270 @@
...
+    self._minActiveDutyCycles.fill(
+        self._minPctActiveDutyCycles * self._activeDutyCycles.max()
+      )
```
chetan51: Align closing parenthesis

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,270 @@
...
+
+    Parameters:
+    ----------------------------
+    @param dutyCycles: An array containing one or more duty cycle values that need
+                    to be updated
+    @param newInput: A new numerical value used to update the duty cycle
+    @param period:  The period of the duty cycle
+    """
+    assert(period >= 1)
+    return (dutyCycles * (period -1.0) + newInput) / period #FIXME use MovingAverage (update it to accept an array)
```
chetan51: Please remove `FIXME` comment and move it to a GitHub issue

src/nupic/research/spatial_pooler.py
```diff
@@ -239,13 +239,21 @@ def __init__(self,
...
+    columnDimensions = numpy.array(columnDimensions, ndmin=1)
+    numColumns = columnDimensions.prod()
+
+    self.boosting = Boosting(
+               numColumns=numColumns,
+               maxBoost=maxBoost,
+               minPctOverlapDutyCycle=minPctOverlapDutyCycle,
+               minPctActiveDutyCycle=minPctActiveDutyCycle,
+               dutyCyclePeriod=dutyCyclePeriod
+               )
```
chetan51: Align parameters and closing parenthesis

src/nupic/research/spatial_pooler.py
```diff
@@ -584,89 +560,111 @@ def setSynPermConnected(self, synPermConnected):
...
-    """Sets the minimum tolerated activity duty cycle, given as percent of
-    neighbors' activity duty cycle"""
-    self._minPctOverlapDutyCycles = minPctOverlapDutyCycles
-
-
-  def getMinPctActiveDutyCycles(self):
-    """Returns the minimum tolerated activity duty cycle, given as percent of
-    neighbors' activity duty cycle"""
-    return self._minPctActiveDutyCycles
+# Boosting related get/set() methods
```
chetan51: Remove unnecessary comment

src/nupic/research/spatial_pooler.py
```diff
@@ -584,89 +560,111 @@ def setSynPermConnected(self, synPermConnected):
...
+  def setMaxBoost(self, maxBoost):
+    """Sets the maximum boost value"""
+    self.boosting._maxBoost = maxBoost
 
 
   def getBoostFactors(self, boostFactors):
     """Returns the boost factors for all columns. 'boostFactors' size must
     match the number of columns"""
-    boostFactors[:] = self._boostFactors[:]
+    boostFactors[:] = self.boosting._boostFactors[:]
```
chetan51: Use `getBoostFactors()` here and elsewhere.

src/nupic/research/spatial_pooler.py
```diff
@@ -584,89 +560,111 @@ def setSynPermConnected(self, synPermConnected):
...
 
 
   def setBoostFactors(self, boostFactors):
     """Sets the boost factors for all columns. 'boostFactors' size must match
     the number of columns"""
-    self._boostFactors[:] = boostFactors[:]
+    self.boosting._boostFactors[:] = boostFactors[:]
```
chetan51: Create and use `Boosting.setBoostFactors()` here and elsewhere.

src/nupic/research/spatial_pooler.py
```diff
@@ -584,89 +560,111 @@ def setSynPermConnected(self, synPermConnected):
...
 
 
-  def getOverlapDutyCycles(self, overlapDutyCycles):
-    """Returns the overlap duty cycles for all columns. 'overlapDutyCycles'
-    size must match the number of columns"""
-    overlapDutyCycles[:] = self._overlapDutyCycles[:]
+  def getMinActiveDutyCycles(self, minActiveDutyCycles):
+    """Returns the minimum activity duty cycles for all columns.
+    '_minActiveDutyCycles' size must match the number of columns"""
+    minActiveDutyCycles[:] = self.boosting._minActiveDutyCycles[:]
```
chetan51: Don't access private variables of `Boosting` directly, use public setters / getters.

src/nupic/research/spatial_pooler.py
```diff
@@ -584,89 +560,111 @@ def setSynPermConnected(self, synPermConnected):
...
+  def getDutyCyclePeriod(self):
+    """Returns the duty cycle period"""
+    return self.boosting._dutyCyclePeriod
+
+
+  def setDutyCyclePeriod(self, dutyCyclePeriod):
+    """Sets the duty cycle period"""
+    self.boosting._dutyCyclePeriod = dutyCyclePeriod
+
+# end of Boosting related get/set() methods
```
chetan51: Remove unnecessary comment

src/nupic/research/spatial_pooler.py
```diff
@@ -766,12 +765,10 @@ def compute(self, inputVector, learn, activeArray):
 
     if learn:
       self._adaptSynapses(inputVector, activeColumns)
-      self._updateDutyCycles(overlaps, activeColumns)
-      self._bumpUpWeakColumns()
-      self._updateBoostFactors()
+      # apply Boosting
```
chetan51: Remove unnecessary comment

src/nupic/research/spatial_pooler.py
```diff
@@ -766,12 +765,10 @@ def compute(self, inputVector, learn, activeArray):
...
+      overlaps = self.boosting.update(overlaps, activeColumns, self)
```
chetan51: `overlaps` is assigned but not used

chetan51:
Are there tests for boosting, so that we can be sure that these changes don't break anything?

chetan51:
Done with review. Reviewed everything except docstrings.

src/nupic/research/boosting.py
breznak: I'd prefer to keep this one, agree to remove (most) of the others...?

src/nupic/research/spatial_pooler.py
breznak: imho useful (?)

breznak:
Thanks for review @chetan51 , will fix it tmr. 

> Are there tests for boosting, so that we can be sure that these changes don't break anything?

Yes, as a part of SP-tests where boosting used to be. I'd keep 'em there as they are closely tied to SP. There are special boosting tests and they pass OK. I still need to fix some SP-unit tests that are broken due to the separation.

src/nupic/research/boosting.py
breznak: https://github.com/numenta/nupic/issues/2673

breznak:
@chetan51 @scottpurdy I've adopted the changes, please continue with your review.

src/nupic/research/spatial_pooler.py
```diff
@@ -584,89 +559,107 @@ def setSynPermConnected(self, synPermConnected):
...
+  def setBoostFactors(self, boostFactors):
+    """Sets the boost factors for all columns. 'boostFactors' size must match
+    the number of columns"""
+    self.boosting.setBoostFactors(boostFactors)
 
 
-  def getBoostFactors(self, boostFactors):
-    """Returns the boost factors for all columns. 'boostFactors' size must
-    match the number of columns"""
-    boostFactors[:] = self._boostFactors[:]
```
breznak: @scottpurdy I've changed the getters to `pass-by-return-value`, instead of "pass by param", as I find it clearer and I think I remember from somewhere that we want to prefer that approach. Is it ok? Also, it would require changes in c++/swig, so I'm not sure if for this PR, or for a separate one?

src/nupic/research/boosting.py
chetan51: The problem is that we don't have any convention for these kinds of section headers (AFAIK). @rhyolight any thoughts on this kind of convention?

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,362 @@
...
+    """Returns the duty cycle period"""
+    return self._dutyCyclePeriod
+
+
+  def setDutyCyclePeriod(self, dutyCyclePeriod):
+    """Sets the duty cycle period"""
+    self._dutyCyclePeriod = dutyCyclePeriod
+
+
+# Boosting core logic:
```
chetan51: Remove unnecessary comment

chetan51:
Done with review (except for docstrings). I would like to review docstrings after the Travis tests are passing.

src/nupic/research/boosting.py
rhyolight: There is only so much you can define as a convention. I say leave it in for now. If someone thinks it is better to remove later, it will get removed.

breznak:
Thanks @chetan51 , I need a resolution on https://github.com/numenta/nupic/pull/2620#discussion_r41975245 
whether to jump into c++, or leave it out for now?

scottpurdy:
Make sure all code files have the copyright header at the top.

src/nupic/research/spatial_pooler.py
chetan51: @scottpurdy bump

breznak:
@chetan51 all tests passing! can you please give this a final look and review the docs? 
@scottpurdy I need to revert https://github.com/breznak/nupic/commit/16683bbf2b713c66f1ee63561f789494c0db6984 , can you help please with the capnp file for boosting, and use it from SP's serialize?

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,387 @@
...
+  def update(self, overlaps, activeColumns, sp):
+    """
+    @param sp - Instance of the Spatial Pooler we want to modify
+    """
+    doUpdateRound = sp._isUpdateRound()
+    doGlobal = sp._globalInhibition or sp._inhibitionRadius > sp._numInputs
+    iteration = sp._iterationNum
+
+    self._updateDutyCycles(overlaps, activeColumns, iteration)
+    # Boosting #1 - boost columns with weak inputs
```
chetan51: Remove unnecessary comment

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,387 @@
...
+    Updates the minimum duty cycles in a global fashion. Sets the minimum duty
+    cycles for the overlap and activation of all columns to be a percent of the
+    maximum in the region, specified by minPctOverlapDutyCycle and
+    minPctActiveDutyCycle respectively. Functionality it is equivalent to
+    _updateMinDutyCyclesLocal, but this function exploits the globality of the
+    computation to perform it in a straightforward, and more efficient manner.
+    """
+    self._minOverlapDutyCycles.fill(
+        self._minPctOverlapDutyCycles * self._overlapDutyCycles.max()
+                                   )
```
chetan51: Align parenthesis

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,387 @@
...
+    self._minActiveDutyCycles.fill(
+        self._minPctActiveDutyCycles * self._activeDutyCycles.max()
+                                  )
```
chetan51: Align parenthesis

src/nupic/research/spatial_pooler.py
```diff
@@ -239,13 +239,20 @@ def __init__(self,
       Determines if inputs at the beginning and end of an input dimension should
       be considered neighbors when mapping columns to inputs.
     """
+    columnDimensions = numpy.array(columnDimensions, ndmin=1)
+    numColumns = columnDimensions.prod()
+
+    self.boosting = Boosting(
+               numColumns=numColumns,
```
chetan51: Fix indentation

src/nupic/research/spatial_pooler.py
chetan51: Bump

src/nupic/research/spatial_pooler.py
```diff
@@ -753,11 +750,12 @@ def compute(self, inputVector, learn, activeArray):
     self._updateBookeepingVars(learn)
     inputVector = numpy.array(inputVector, dtype=realDType)
     inputVector.reshape(-1)
+
     overlaps = self._calculateOverlap(inputVector)
 
-    # Apply boosting when learning is on
+    # Boosting #2 - overlaps on peer-suppressed columns
```
chetan51: Remove unnecessary comment

src/nupic/research/spatial_pooler.py
```diff
@@ -766,12 +764,10 @@ def compute(self, inputVector, learn, activeArray):
 
     if learn:
       self._adaptSynapses(inputVector, activeColumns)
-      self._updateDutyCycles(overlaps, activeColumns)
-      self._bumpUpWeakColumns()
-      self._updateBoostFactors()
+      # Boosting #1 - boost columns with weak input overlap + do timekeeping
```
chetan51: Remove unnecessary comment

tests/unit/nupic/research/spatial_pooler_unit_test.py
```diff
@@ -1705,8 +1706,7 @@ def testGetNeighborsND(self):
     self.assertEqual(layout1D[list(negative)].any(), False)
 
 
-  @unittest.skipUnless(
-      capnp, "pycapnp is not installed, skipping serialization test.")
+  @unittest.skip("pycapnp is not installed, skipping serialization test.") #FIXME
```
chetan51: Why is this change necessary? We should be testing serialization if it is enabled.

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,387 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2013-2014, Numenta, Inc.  Unless you have an agreement
```
chetan51: Update copyright year

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,387 @@
...
+from nupic.bindings.math import (GetNTAReal,
+                                 Random as NupicRandom)
+
+realDType = GetNTAReal()
+
+
+
+class Boosting(object):
+  """
+  Implements boosting for columns of the spatial pooler.
```
chetan51: I suggest using the exact same language as in the white paper for this docstring. Replacements to the language below.

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,387 @@
...
+  Boosting is an optional process that helps the spatial pooler (SP) to adapt
+  faster to changes in the input stream. Boosting acts on 2 levels: "duplicit
+  peer columns activity" and "mismatched input overlap".
```
chetan51: "There are two separate boosting mechanisms in place to help a column learn connections."

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,387 @@
...
+
+  1/ Boosting of mispatched input:
+  Triggered when the column is "growing" (connected to) the inputs that are 
+  not used (anymore). This can happen in two cases, when a part of the input
+  vector is never used (wrongly, too generic designed encoder, etc.), or when
+  a part of the input that had been used before is not useful anymore for a long
+  period of time. 
+  Biologically the unmatched synapses would die out of hunger, in the current 
+  implementation we increase their improtance in hope that (a few) will start 
+  to act on a subset of the input and become a new feature.
```
chetan51: "If a column does not win often enough (as measured by activeDutyCycle), its overall boost value is increased."

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,387 @@
...
+
+  2/ Boosting for duplicit peer activity:
+  In this scenario, the column has a good synaptic overlap over the input,
+  but it is very similar to the input field of the neighboring column(s). 
+  It results in a SDR that is not well (D)istributed, and we are wasting 
+  resources on the columns activity, as it is duplicit to others'. This leads
+  to the column never being active (winning in inhobotion). 
+  In this implementation the fix is to temporarily artificially boost 
+  the column's "importance" by a boosting factor, thus giving it a chance to
+  win in some scenario and creating it's niche.
```
chetan51: "If a column does not win often enough (as measured by `ActivityDutyCycles`), its overall boost value is increased."

src/nupic/research/boosting.py
chetan51: "If a column's connected synapses do not overlap well with any inputs often enough (as measured by `OverlapDutyCycles`), its permanence values are boosted."

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,387 @@
...
+    self._minOverlapDutyCycles = numpy.zeros(numColumns, dtype=realDType)
+    self._minPctOverlapDutyCycles = minPctOverlapDutyCycle
+    self._minPctActiveDutyCycles = minPctActiveDutyCycle
+    self._dutyCyclePeriod = dutyCyclePeriod
+
+
+
+  def getBoostFactors(self):
+    """
+    get boosting for the #2 method - duplicit peer columns
```
chetan51: You've already described boost factors elsewhere. Thus, replace this with "Get boost factors for al columns". Also, use a single line docstring.

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,387 @@
...
+    """
+    return self._boostFactors[:]
+
+
+  def setBoostFactors(self, newBoost):
+    """
+    set new boosting paramethers for #2 Boosting method
```
chetan51: Replace with "Set boost factors for all columns".

chetan51:
Done with review.

src/nupic/research/spatial_pooler.py
breznak: sorry, reverted that one during some change. fixed back

src/nupic/research/boosting.py
breznak: how should this be aligned? to the start of the upper self?

src/nupic/research/spatial_pooler.py
breznak: how?

tests/unit/nupic/research/spatial_pooler_unit_test.py
breznak: it's a TEMP change to see if the tests are passing. I've mentioned that in a comment. Waiting if Scott can help with CAPNP for the Boosting class

src/nupic/research/boosting.py
breznak: :+1: thanks for help with the essay HW. I've just changed "Activity" to "Active"(DutyCycles) to match the variable in code.

breznak:
@chetan51 thanks, fixed! Updated the PR desc with the remaining todo item.

src/nupic/research/spatial_pooler.py
chetan51: Should be 2 spaces after `self.boosting`.

breznak:
fixed, thanks.

breznak:
@chetan51 ping?

tests/unit/nupic/research/spatial_pooler_unit_test.py
chetan51: @scottpurdy ping

chetan51:
https://github.com/numenta/nupic/pull/2620#discussion_r43195566

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,358 @@
...
+    Parameters:
+    ----------------------------
+    @param dutyCycles: An array containing one or more duty cycle values that need
+                    to be updated
+    @param newInput: A new numerical value used to update the duty cycle
+    @param period:  The period of the duty cycle
+    """
+    assert(period >= 1)
+    return (dutyCycles * (period -1.0) + newInput) / period #FIXME #2673
+
```
scottpurdy: remove blank line at end of file

scottpurdy:
I can help with the capnp serialization but there are still backwards compatibility issues:

1. Old saved models will fail when loaded after this change.
2. Has anything been done to validate that the SP still gets the same results after this change? I suspect that it will not. If at all possible, we should make sure the results don't change at all. If we really can't do that we need to be very explicit about why so that we know we aren't breaking the algorithm.

src/nupic/research/boosting.py
```diff
@@ -0,0 +1,358 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2013-2015, Numenta, Inc.  Unless you have an agreement
```
scottpurdy: This file is new so just 2015

breznak:
@scottpurdy please see below...

> Has anything been done to validate that the SP still gets the same results after this change? I suspect that it will not. If at all possible, we should make sure the results don't change at all. If we really can't do that we need to be very explicit about why so that we know we aren't breaking the algorithm.

Can you elaborate what you're suspecting to break? The PR does not add any new functionality, but refactors the code. Therefore no extra checks (should) have been done, the existing SP functionality checks are passing.

breznak:
> Old saved models will fail when loaded after this change. 

@scottpurdy I've just quickly rechecked the serialization and theoretically nothing should break, right? As the SP's API stays intact. In practice it breaks because the current implementation wrongly/directly accesses internal variables. Switching to the public getters/setters would shield this problem. Would that work also for the compatibility with older serialized models? (I think so, if not, could we implement an  another "compat switch" for the newer version of SP?)

scottpurdy:
After updating, the SP code will be different and will expect that the state matches the new format (with a separate boosting object as an instance variable of SP) so in the deserialization how, the loaded data must be updated if it is in the old format. You haven't done that.

And I see some functions getting called in slightly different order with the change so was suspicious that it might affect the results. Please add details on how you have validated that the results are the same (i.e. what the the "existing SP functionality checks" and do they check properties of SP or exact results?).

breznak:
> After updating, the SP code will be different and will expect that the state matches the new format (with a separate boosting object as an instance variable of SP) so in the deserialization how, the loaded data must be updated if it is in the old format. You haven't done that.

Hi, I haven't done anything in the de/serialization, I was hoping you'll look to that :). If you can't I'll try to do it somehow...

> And I see some functions getting called in slightly different order with the change so was suspicious that it might affect the results. Please add details on how you have validated that the results are the same (i.e. what the the "existing SP functionality checks" and do they check properties of SP or exact results?).

If so, it's a good catch, I'll try to reduce any changes here. As for the validation - I haven't added any specific test cases, but (I hope I'm safe to) assume there exist checks for correctness of a special pooler implementation (You know better than I do, but namely the `spatial_pooler` unit tests, integration tests, etc. - which check both some specific properties of SP, as well as overall performance/quality of resulting SDRs.)

@scottpurdy

scottpurdy:
I thought you were asking specifically for help with the new serialization (Cap'n Proto in read/write methods). We also need to implement the save/load custom serialization format that is the current method. You can look at other algorithm implementations for examples and that should also show how the version is used to update the data format. I'm happy to help with capnp implementation since it isn't yet in use but will wait until the rest is done (and this PR can be merged without the capnp implementations).

rhyolight:
Feel free to re-open if you start work again.

Issue #2630, opened 2015-10-06
remove duplicit Numpy dependency, already in nupic.bindings


breznak:
`numpy` is a core dependency of nupic, but it's already included in `nupic.bindings` and therefor should not be defined in nupic again. 

The problems with this were reported by @rhyolight when installation went twice through installing numpy, as different versions were required for core(bindings) and nupic. 

The removal speeds up time for installation and could avoid some tricky problems when different version is used for compilation and then for runtime.

rhyolight:
> The problems with this were reported by @rhyolight when installation went twice through installing numpy, as different versions were required for core(bindings) and nupic.

See: https://github.com/numenta/nupic/issues/2607

This was actually because nupic.bindings and nupic required different versions of numpy (one was `==1.9.2` and the other was `>=1.9.2`, so because the `>=` requirement was fulfilled first and the latest release was 1.9.3, it got installed. Then during the nupic build, 1.9.2 was installed. 

After fixing both requirements to have the same explicit dependency, this no longer happens. But thank you for following up on the video with this report!

rhyolight:
Now that I think about it... we might be able to just remove the numpy requirement from nupic completely, if it comes along with the nupic.bindings binary file, or gets installed as a dependency of nupic.bindings during a local compilation. 

Thoughts @scottpurdy @oxtopus ? Maybe someone should create a PR that removes numpy from nupic and see what happens in the builds.

rhyolight:
> Maybe someone should create a PR that removes numpy from nupic and see what happens in the builds.

Maybe I should pay attention! https://github.com/numenta/nupic/issues/2631

oxtopus:
There are multiple schools of thought, and I don't think there's consensus on the right one.

Personally, I don't think we should duplicate requirements in requirements.txt, if one is explicitly satisfied in an earlier-in-the-chain dependency.  However, others like seeing explicit dependencies in the requirements at every level.

One extreme view might be that if you require functionality of a dependency of a dependency, then you access it as part of that dependencies namespace, or wrap it in an API.  For example, if nupic requires numpy, which is part of the dependencies defined in nupic.bindings, then instead of importing `numpy` directly, then code inside of `nupic` imports `numpy` from `nupic.bindings`.  i.e. `import nupic.bindings.numpy as numpy` or `from nupic.bindings import numpy`.  At least that way, it's a little more explicit where the dependencies come from.

scottpurdy:
[...]

PR #2631, opened 2015-10-06
remove numpy as already in bindings

breznak:
Fixes: #2630


oxtopus:
I support this change, but I don't believe there's consensus among commenters in #2630 to accept this approach

Issue #2632, opened 2015-10-06
State-full Anomaly class - avoid 2 column state fields at T, T-1


breznak:
currently `Anomaly.compute()` uses 2 main params: `activeColumns` and `predictedColumns`;
there may be some confusion about those variables, but basically `predictedColumns` = output of the TP at time T-1, and `activeColumns` = input to the TP at time T. These 2 values are compared and an anomaly score is produced. 

Enhancement would be to carry one of the states in the class, and require only an actual input. This would simplify things in CLAModel and other user-scripts with anomaly. 

Question is, how this should be called? a `next()` method, or another `compute()` with different params, or even preplace the `anomaly.compute()` as a default?

breznak:
CC @scottpurdy I think this really makes sense for the Anomaly class..

cogmission:
:+1: That would be consistent with what the TemporalMemory does at least...
Though you know I'm against state in the algorithms in general. But I agree
with the interface change... If anybody cares...

On Tue, Oct 6, 2015 at 7:47 AM, breznak <notifications@github.com> wrote:

> CC @scottpurdy <https://github.com/scottpurdy> I think this really makes
> sense for the Anomaly class..
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/numenta/nupic/issues/2632#issuecomment-145846470>.
>



-- 
*With kind regards,*

David Ray
Java Solutions Architect

*Cortical.io <http://cortical.io/>*
Sponsor of:  HTM.java <https://github.com/numenta/htm.java>

d.ray@cortical.io
http://cortical.io

breznak:
> Though you know I'm against state in the algorithms in general.

Thanks for the heads up @cogmission ! Actually this one could be done state-less, if the book keeping is done only in the method, like: 

```
def next(actual):
  global old
  score = compute(actual, old)
  old = actual
  return score
```

But I'm not sure if it's any nicer, and would have problems with serialization.

rhyolight:
Global state is still state. I would much rather this be instance state than global state.

cogmission:
The idea is not to get rid of state (that's impossible) - but to remove it
from the Algorithms


On Tue, Oct 6, 2015 at 11:46 AM, Matthew Taylor <notifications@github.com>
wrote:

> Global state is still state. I would much rather this be instance state
> than global state.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/numenta/nupic/issues/2632#issuecomment-145924196>.
[...]

PR #2635, opened 2015-10-06
Boosting disturbance test

breznak:
~~Fixes: 2634~~
Fixes: #2687 
~~"Blocked": this test would be nicer after https://github.com/numenta/nupic/issues/2632~~ - Invalid

### DIFF
--- a/tests/unit/nupic/research/spatial_pooler_boost_test.py
+++ b/tests/unit/nupic/research/spatial_pooler_boost_test.py
@@ -219,7 +219,10 @@ class SpatialPoolerBoostTest(unittest.TestCase):
     self.assertGreater(_computeOverlap(self.lastSDR[0], 
                        self.lastSDR[1]),
                        SIMILAR_SDR_OVERLAP,
-                       "First two SDR's don't overlap enough: \n%r \nvs. \n%r" % (self.lastSDR[0], self.lastSDR[1]))
+                       "First two SDR's don't overlap enough: \n%r \nvs. \n%r" % (
+                          self.lastSDR[0].nonzero()[0], 
+                          self.lastSDR[1].nonzero()[0])
+                      )
     
     # Verify the last three SDR's have low overlap with everyone else.
     for i in [2, 3, 4]:
--- a/tests/unit/nupic/research/spatial_pooler_boost_test.py
+++ b/tests/unit/nupic/research/spatial_pooler_boost_test.py
@@ -36,6 +36,9 @@ uintType = "uint32"
 
 # set a single seed for running both implementations
 SEED = int((time.time()%10000)*10)
+# when we need similar SDRs, they must have overlap of this # columns
+SIMILAR_SDR_OVERLAP = 9 #FIXME is this a magic const? Could it be some numColumns*(2% =SDR active bits)*(50% =overlap of the 2 similars)?
+DIFFERENT_SDR_OVERLAP = 18 #FIXME isn't this too benevolent, that different>>similar? should be more like 9 & 2?
 
 def _computeOverlap(x, y):
   """
@@ -213,15 +216,19 @@ class SpatialPoolerBoostTest(unittest.TestCase):
     self.assertTrue(_areAllSDRsUnique(self.lastSDR), "All SDR's are not unique")
 
     # Verify that the first two SDR's have some overlap.
-    self.assertGreater(_computeOverlap(self.lastSDR[0], self.lastSDR[1]), 9,
-                       "First two SDR's don't overlap much: \n%r \nvs. \n%r")
+    self.assertGreater(_computeOverlap(self.lastSDR[0], 
+                       self.lastSDR[1]),
+                       SIMILAR_SDR_OVERLAP,
+                       "First two SDR's don't overlap enough: \n%r \nvs. \n%r" % (self.lastSDR[0], self.lastSDR[1]))
     
     # Verify the last three SDR's have low overlap with everyone else.
     for i in [2, 3, 4]:
       for j in range(5):
         if (i!=j):
-          self.assertLess(_computeOverlap(self.lastSDR[i], self.lastSDR[j]),
-                          18, "One of the last three SDRs has high overlap")
+          self.assertLess(_computeOverlap(self.lastSDR[i], 
+                          self.lastSDR[j]),
+                          DIFFERENT_SDR_OVERLAP, 
+                          "One of the last three SDRs has too high overlap")
 
 
   def boostTestPhase1(self):
@@ -253,7 +260,6 @@ class SpatialPoolerBoostTest(unittest.TestCase):
     self.assertGreaterEqual(dutyCycles[self.winningIteration > 0].min(),
                             0.2,
                             "Active columns have duty cycle that is too low.")
-
     self.verifySDRProperties()
     
 
@@ -292,7 +298,6 @@ class SpatialPoolerBoostTest(unittest.TestCase):
                 "Average on-columns duty cycle is too low.")
     self.assertLessEqual(avg, 0.30,
                 "Average on-columns duty cycle is too high.")
-
     self.verifySDRProperties()
 
 
--- a/tests/unit/nupic/research/spatial_pooler_boost_test.py
+++ b/tests/unit/nupic/research/spatial_pooler_boost_test.py
@@ -106,21 +106,58 @@ class SpatialPoolerBoostTest(unittest.TestCase):
   should not change.
   """
 
-  def setUp(self):
+  def setUp(self, defaultParams=None, imp="py", data=None):
     """
     Set various constants. Create the input patterns and the spatial pooler
     """
-    self.inputSize = 90
-    self.columnDimensions = 600
+    # Setup the SP creation parameters we will use
+    if defaultParams is None:
+      self.params = {
+        'inputDimensions':            [90],
+        'columnDimensions':           [600],
+        'potentialRadius': 	      90,
+        'potentialPct':               0.9,
+        'globalInhibition':           True,
+        'numActiveColumnsPerInhArea': 60,
+        'minPctActiveDutyCycle':      0.1,
+        'synPermActiveInc':           0.0,
+        'synPermInactiveDec':         0.0,
+        'dutyCyclePeriod':            10,
+        'maxBoost':                   10.0,
+        'seed':                       SEED,
+      }
+    else:
+      self.params = defaultParams
+
+    # Create SP
+    self.sp = CreateSP(imp, self.params)
+    self.spImplementation = imp
+
+    # set helper variables
+    self.inputSize = numpy.prod(self.sp.getInputDimensions())
+    self.columnDimensions = numpy.prod(self.sp.getColumnDimensions())
 
     # Create a set of input vectors, x
     # B,C,D don't overlap at all with other patterns
-    self.x = numpy.zeros((5, self.inputSize), dtype=uintType)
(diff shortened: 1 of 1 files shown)


tests/unit/nupic/research/spatial_pooler_boost_test.py
```diff
@@ -193,19 +207,22 @@ def verifySDRProperties(self):
     self.assertTrue(_areAllSDRsUnique(self.lastSDR), "All SDR's are not unique")
 
     # Verify that the first two SDR's have some overlap.
-    self.assertGreater(_computeOverlap(self.lastSDR[0], self.lastSDR[1]), 9,
-                       "First two SDR's don't overlap much")
+    self.assertGreater(_computeOverlap(self.lastSDR[0], 
+                       self.lastSDR[1]),
+                       SIMILAR_SDR_OVERLAP,
```
breznak: @subutai are these numbers correct? SIMILAR=9, DIFFERENT=18
See https://github.com/numenta/nupic/pull/2635/files#diff-77566f80f8ee4a67268c105e75817638R221
https://github.com/numenta/nupic/pull/2635/files#diff-77566f80f8ee4a67268c105e75817638R40

tests/unit/nupic/research/spatial_pooler_boost_test.py
```diff
@@ -339,13 +359,82 @@ def boostTestLoop(self, imp):
...
-    self.boostTestLoop("cpp")
-
+    self.setUp(imp="cpp")
+    self.boostTestLoop()
+
+
+  def testBoostingDefaultSPParams(self):
+    """Boosting test with SP's default params."""
+    orig = copy.deepcopy(self.params)
+    params = {}
```
breznak: I'd like to verify the tests with SP's default params. Just I'm not sure if all of them make sense as some numbers (0.0 syn. inc/dec) are pretty extreme. 

I'm getting

```
======================================================================
FAIL: testBoostingDefaultSPParams (__main__.SpatialPoolerBoostTest)
Boosting test with SP's default params.
----------------------------------------------------------------------
  File "tests/unit/nupic/research/spatial_pooler_boost_test.py", line 377, in testBoostingDefaultSPParams
    self.boostTestLoop()
  File "tests/unit/nupic/research/spatial_pooler_boost_test.py", line 354, in boostTestLoop
    self.boostTestPhase1()
  File "tests/unit/nupic/research/spatial_pooler_boost_test.py", line 261, in boostTestPhase1
    self.verifySDRProperties()
  File "tests/unit/nupic/research/spatial_pooler_boost_test.py", line 220, in verifySDRProperties
    "First two SDR's don't overlap enough: \n%r \nvs. \n%r" % (self.lastSDR[0], self.lastSDR[1]))
AssertionError: 1 not greater than 9 : First two SDR's don't overlap enough: 
array([0, 0, 0, ..., 0, 0, 0], dtype=uint32) 
vs. 
array([0, 0, 0, ..., 0, 0, 0], dtype=uint32)
```

tests/unit/nupic/research/spatial_pooler_boost_test.py
```diff
@@ -339,13 +358,86 @@ def boostTestLoop(self, imp):
...
+    # This typically happens on simple, periodic data with a bigger SP (num. columns)
+    def runOnce(inp, enc, sp, tp, an):
+      enD = enc.encode(inp)
+      spD = numpy.zeros(sp.getColumnDimensions(), dtype='int32')
+      sp.compute(enD, True, spD)
+      tpD = numpy.zeros(tp.numberOfCols)
+      pastPred = tp.columnConfidences().nonzero()[0] # = predictions at T-1
+      active = spD # active input at T
+      tpD = tp.compute(spD, True, True)
+      anS = an.compute(active, pastPred)
```
breznak: @scottpurdy are these params correct? In context of the mistake I made in https://github.com/numenta/nupic/issues/2632 
Should we add `def getPredictedColumns() = tp.columnConfidences().nonzero()[0]` just for completeness?

breznak:
The tests are failing on 2 occasions, 
the first is for review, not 100% sure if bug in test or nupic, 
the latter uncovers a bug in Nupic. 

```
FAIL: testBoostingDefaultSPParams (__main__.SpatialPoolerBoostTest)
Boosting test with SP's default params.
----------------------------------------------------------------------
  File "tests/unit/nupic/research/spatial_pooler_boost_test.py", line 381, in testBoostingDefaultSPParams
    self.boostTestLoop()
  File "tests/unit/nupic/research/spatial_pooler_boost_test.py", line 359, in boostTestLoop
    self.boostTestPhase1()
  File "tests/unit/nupic/research/spatial_pooler_boost_test.py", line 266, in boostTestPhase1
    self.verifySDRProperties()
  File "tests/unit/nupic/research/spatial_pooler_boost_test.py", line 224, in verifySDRProperties
    self.lastSDR[1].nonzero()[0])
AssertionError: 1 not greater than 9 : First two SDR's don't overlap enough: 
array([ 886, 1458, 1609, 1727, 1905, 2078, 2137, 2435]) 
vs. 
array([ 222,  514,  555,  600, 1563, 1905, 2561, 2627, 2992, 3779])

======================================================================
FAIL: testBoostingNoDisturbances (__main__.SpatialPoolerBoostTest)
Boosting should not create any (significant) anomalies/disturbances.
----------------------------------------------------------------------
  File "tests/unit/nupic/research/spatial_pooler_boost_test.py", line 438, in testBoostingNoDisturbances
    self.assertTrue(res < 0.1, "There was an artificial disturbance in consecutive results.")
AssertionError: False is not True : There was an artificial disturbance in consecutive results.
```

scottpurdy:
These tests are enforcing behavior that isn't guaranteed with current boosting so we can't include them. These tests would only make sense for a form of boosting with the explicit intent of having the tested behavior.

breznak:
How do you mean it Scott? The tests prove there's a problem caused by
boosting. Would it be ok to include once the fixed implementation is in
place then?

scottpurdy:
No, the tests prove that the current boosting doesn't have some specific property. But this isn't a bug, it is the expected behavior of the current boosting mechanism. If we want to change that then we can experiment with other boosting algorithms. But adding a test for behavior that isn't expected for the current boosting doesn't make any sense.

I have some ideas for next steps for your experimentation with boosting and anomaly scores. Which issue best covers the high level objective? I'll put my comments there.

breznak:
> No, the tests prove that the current boosting doesn't have some specific property. But this isn't a bug, it is the expected behavior of the current boosting mechanism. If we want to change that then we can experiment with other boosting algorithms. But adding a test for behavior that isn't expected for the current boosting doesn't make any sense. 

Imho to test replicates a very old bug that makes false anomalies/disturbances in prediction on "simple" tasks. I believe that is a "bug" in HTM implementation, current boosting is just a cause of it. 

> I have some ideas for next steps for your experimentation with boosting and anomaly scores. Which issue best covers the high level objective? I'll put my comments there. 

@scottpurdy Please see https://github.com/numenta/nupic/issues/823 for history and sub-tasks I've identified.

rhyolight:
@breznak @scottpurdy Are you guys still disagreeing on this issue? I think there is a miscommunication here. Mark thinks there is a bug and Scott doesn't, as far as I can tell.

scottpurdy:
The behavior is correct (it is as it was designed) so I won't accept this test. We would consider an alternative implementation but it will require going through the steps that @subutai outlined in the nupic-committers list. We can continue the discussion in #823.

tests/unit/nupic/research/spatial_pooler_boost_test.py
```diff
@@ -20,18 +21,23 @@
...
+from nupic.bindings.algorithms import SpatialPooler as CppSP
 
 uintType = "uint32"
 
 # set a single seed for running both implementations
-SEED = int((time.time()%10000)*10)
+SEED = 42
+# when we need similar SDRs, they must have overlap of this # columns
+SIMILAR_SDR_OVERLAP = 9 #FIXME is this a magic const? Could it be some numColumns*(2% =SDR active bits)*(50% =overlap of the 2 similars)?
+DIFFERENT_SDR_OVERLAP = 18 #FIXME isn't this too benevolent, that different>>similar? should be more like 9 & 2?
```
breznak: please review these?

tests/unit/nupic/research/spatial_pooler_boost_test.py
breznak: The question is: for "default params test", should the inputs (with 50% overlap) still pass the test? I believe they should

breznak:
Please review the PR, I have some questions, propose code cleanup there. The "offending" test was removed, still 2 new tests - one passes ok, 1 causes problems and I would like to know if (why?) the test is pointless, or if it addresses a real problem. 

Here is the output of 2 SDRs of 2 patterns with 50% overlap:
```
E     AssertionError: 1 not greater than 9 : First two SDR's don't overlap enough:
E     array([ 886, 1458, 1609, 1727, 1905, 2078, 2137, 2435])
E     vs.
E     array([ 222,  514,  555,  600, 1563, 1905, 2561, 2627, 2992, 3779])
```

rhyolight:
@scottpurdy @breznak Is there any way we can get into a chat room or a hangout to discuss this?

scottpurdy:
@chetan51 - Can you review this PR?

A few high level comments:

- `setUp` is automatically run by the test framework before each test function. You shouldn't call it in your own code
- It seems cleaner that the SP implementation be passed directly to the utility function `boostTestLoop` rather than storing it as an instance attribute that the `boostTestLoop` function depends on. If there is a need for testing different data then I think it is better to make that a `boostTestLoop` argument instead of setting `self.x`.
- This is a unit test so it should be checking for exact behavior. Quality metrics and relative overlaps belong in algorithmic experiments and benchmarks/quality tests but not in unit tests.

tests/unit/nupic/research/spatial_pooler_boost_test.py
```diff
@@ -339,13 +359,49 @@ def boostTestLoop(self, imp):
...
+    # py
+    self.setUp(params, "py") # reinitialize with new params
+    self.boostTestLoop()
+    # cpp
+    self.setUp(params, "cpp")
+    self.boostTestLoop()
+
+
+  def testBoostingSDRPatterns(self):
+    """Boosting test with random patterns, instead of continuous data."""
```
chetan51: It's not clear what this is testing from this docstring. The patterns are not really random, since the first two are identical and the third one shares 50% of the bits. Can you elaborate here on exactly what this new test is testing?

chetan51:
Regarding @scottpurdy's high level comments:

> setUp is automatically run by the test framework before each test function. You shouldn't call it in your own code

I agree, please move the logic you have in `setUp` to another function so you can call it explicitly.

> It seems cleaner that the SP implementation be passed directly to the utility function boostTestLoop rather than storing it as an instance attribute that the boostTestLoop function depends on. If there is a need for testing different data then I think it is better to make that a boostTestLoop argument instead of setting self.x.

I agree with this too, please try to have each function (for example `boostTestLoop` and `boostTestPhase1`) take the relevant data they need, rather than relying on `self` (I do recognize that the existing code isn't great either, so whatever you can do to clean it up like this would be appreciated).

> This is a unit test so it should be checking for exact behavior. Quality metrics and relative overlaps belong in algorithmic experiments and benchmarks/quality tests but not in unit tests.

We should move this test to the integration tests directory, but that can be a separate PR.

chetan51:
I will review again once the purpose of the new tests have been elaborated and @scottpurdy's suggestions have been implemented.

Issue #2640, opened 2015-10-07
Saving a model created within a multiprocessing.Process and saved on the main process will fail


JonnoFTW:
I'm running a numerous models, each created inside it's own process from a [`multiprocessing.Pool`](https://docs.python.org/2/library/multiprocessing.html#multiprocessing.pool.multiprocessing.Pool). Afterwards, if I want to save this model on the main process, I get the following error: 
```
File "index.py", line 254, in save_model
    model.save(out_dir)
  File "/scratch/temp/venvs/py2.7.10/lib/python2.7/site-packages/nupic/frameworks/opf/model.py", line 226, in save
    self._serializeExtraData(extraDataDir=self._getModelExtraDataDir(saveModelDir))
  File "/scratch/temp/venvs/py2.7.10/lib/python2.7/site-packages/nupic/frameworks/opf/clamodel.py", line 1302, in _serializeExtraData
    self._netInfo.net.save(outputDir)
AttributeError: 'NoneType' object has no attribute 'save'
```
Where the value of `out_dir` is `/scratch/Dropbox/PhD/htm_models_adelaide/engine/model_store/3103`

My multiprocess code is here (I've commented out the saving for now but you get the idea): https://github.com/JonnoFTW/htm-models-adelaide/blob/master/engine/index.py#L306
My model creation code is here: https://github.com/JonnoFTW/htm-models-adelaide/blob/master/engine/index.py#L74

Additionally, trying to `save` or `load` a model from within a [`multiprocessing.Process`](https://docs.python.org/2/library/multiprocessing.html#multiprocessing.Process) will fail

rhyolight:
@scottpurdy @oxtopus @jcasner I believe this is an important bug to address, especially because @JonnoFTW is working on an HTM Challenge submission. 

Can we divert engineering resources to investigate this?

oxtopus:
I don't think it's a bug, per se.  It's an issue w/ serialization and shared memory in python multiprocesing pools.  Consider this example:

```python
from multiprocessing import Pool
import os

class FooObject(object): pass

def foo(*args):
  fooObj = FooObject()
  print "PID {} created: {}".format(os.getpid(), id(fooObj))
  return fooObj

def main():
  pool = Pool(3)
  results = pool.map(foo, range(3))
  for result in results:
    print "PID {} Result: {}".format(os.getpid(), id(result))

if __name__ == "__main__":
  main()
```

The output for which is:

```
PID 1368 created: 4450700048
[...]

PR #2650, opened 2015-10-08
Issue a descriptive TypeError when user passes non-<type 'str'> in save()

### LINKED ISSUE #2649: model.save() explicitly requires <type 'str'>
See #2640 in which a user was inadvertently passing `<type 'unicode'>`, resulting in an error in the underlying c++ bindings.  NuPIC should check for type, and raise a more descriptive exception beforehand.

oxtopus:
Fixes #2649

See https://github.com/numenta/nupic/issues/2640#issuecomment-146431286 for context

This also includes some stylistic cleanup.  Relevant change is in https://github.com/numenta/nupic/commit/1be1bd961ce6bb6f678d55505a70a572ce32946a

### DIFF
--- a/src/nupic/engine/__init__.py
+++ b/src/nupic/engine/__init__.py
@@ -718,6 +718,8 @@ class Network(engine.Network):
     """
     @doc:place_holder(Network.save)
     """
+    if len(args) > 0 and not isinstance(args[0], str):
+      raise TypeError("Save path must be of type {}.".format(str))
     engine.Network.save(self, *args, **kwargs)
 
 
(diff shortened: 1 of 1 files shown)


src/nupic/engine/__init__.py
```diff
@@ -718,12 +717,13 @@ def save(self, *args, **kwargs):
     """
     @doc:place_holder(Network.save)
     """
+    if len(args) > 0 and not isinstance(args[0], str):
+      raise TypeError("Save path must be of type {}.".format(str))
```
rhyolight: This seems to be the only functional change in this PR, yes?

oxtopus: Yes, rest is cleanup.

rhyolight:
:+1:

PR #2670, opened 2015-10-13
Fixes #2667 Retry decorator bug and unit tests

SaganBolliger:
Fixes #2667 

Brings over retry decorator bug fix from numenta/numenta-apps#399
Brings over unit tests for retry decorator from numenta/numenta-apps#382

@scottpurdy @vitaly-krugl

PR #2677, opened 2015-10-14
2672 Replace retry decorator getLoggerCallback with logger parameter

### LINKED ISSUE #2672: Replace retry decorator getLoggerCallback with logger parameter
https://github.com/numenta/nupic/blob/f06bed0931aa3879fcd18da91e2526df8d314476/src/nupic/support/decorators.py#L148

This code allows the user to pass a callback function that returns a logger. This was due to an issue with loggers being created before we initialize logging but this issue has been fixed so we should change this parameter to take an actual logger instance.

The numenta-apps retry decorator has the desired behavior here:
https://github.com/numenta/numenta-apps/blob/994de8f8d99e7df9f585c7f762b8a3b46a2fe8b8/nta.utils/nta/utils/error_handling.py#L124

SaganBolliger:
Fixes #2672

@vitaly-krugl Please review?
@scottpurdy

### DIFF
--- a/src/nupic/database/ClientJobsDAO.py
+++ b/src/nupic/database/ClientJobsDAO.py
@@ -38,7 +38,7 @@ from nupic.support import pymysqlhelpers
 
 
 _MODULE_NAME = "nupic.database.ClientJobsDAO"
-
+_LOGGER = logging.getLogger(__name__)
 
 
 class InvalidConnectionException(Exception):
@@ -49,16 +49,6 @@ class InvalidConnectionException(Exception):
   pass
 
 
-
-def _getLogger():
-  """ NOTE: this cannot be a global variable because the logging subsystem
-  needs to be initialized by the host app, and that usually happens after
-  imports
-  """
-  return logging.getLogger(
-    ".".join(['com.numenta', _MODULE_NAME, ClientJobsDAO.__name__]))
-
-
 # Create a decorator for retrying idempotent SQL operations upon transient MySQL
 #  failures.
 # WARNING: do NOT indiscriminately decorate non-idempotent operations with this
@@ -66,7 +56,7 @@ def _getLogger():
 #  insertions, etc.
 # NOTE: having this as a global permits us to switch parameters wholesale (e.g.,
 #  timeout)
-g_retrySQL = pymysqlhelpers.retrySQL(getLoggerCallback=_getLogger)
+g_retrySQL = pymysqlhelpers.retrySQL(logger=_LOGGER)
 
 
 
@@ -542,7 +532,7 @@ class ClientJobsDAO(object):
 
 
   @staticmethod
-  @logExceptions(_getLogger)
+  @logExceptions(_LOGGER)
   def get():
     """ Get the instance of the ClientJobsDAO created for this process (or
     perhaps at some point in the future, for this thread).
@@ -563,7 +553,7 @@ class ClientJobsDAO(object):
     return ClientJobsDAO._instance
 
 
-  @logExceptions(_getLogger)
+  @logExceptions(_LOGGER)
   def __init__(self):
     """ Instantiate a ClientJobsDAO instance.
 
@@ -571,7 +561,7 @@ class ClientJobsDAO(object):
     ----------------------------------------------------------------
     """
 
-    self._logger = _getLogger()
+    self._logger = _LOGGER
 
     # Usage error to instantiate more than 1 instance per process
     assert (ClientJobsDAO._instance is None)
@@ -623,7 +613,7 @@ class ClientJobsDAO(object):
     return ''.join(pubWords)
 
 
-  @logExceptions(_getLogger)
+  @logExceptions(_LOGGER)
   @g_retrySQL
   def connect(self, deleteOldVersions=False, recreate=False):
     """ Locate the current version of the jobs DB or create a new one, and
@@ -653,7 +643,7 @@ class ClientJobsDAO(object):
     return
 
 
-  @logExceptions(_getLogger)
+  @logExceptions(_LOGGER)
   def _initTables(self, cursor, deleteOldVersions, recreate):
     """ Initialize tables, if needed
 
@@ -1283,7 +1273,7 @@ class ClientJobsDAO(object):
     return self._connectionID
 
 
-  @logExceptions(_getLogger)
+  @logExceptions(_LOGGER)
   def jobSuspend(self, jobID):
     """ Requests a job to be suspended
 
@@ -1310,7 +1300,7 @@ class ClientJobsDAO(object):
     return
 
 
-  @logExceptions(_getLogger)
+  @logExceptions(_LOGGER)
   def jobResume(self, jobID, alreadyRunning=False):
     """ Resumes processing of an existing job that is presently in the
     STATUS_COMPLETED state.
@@ -1357,7 +1347,7 @@ class ClientJobsDAO(object):
     return
 
 
-  @logExceptions(_getLogger)
+  @logExceptions(_LOGGER)
   def jobInsert(self, client, cmdLine, clientInfo='', clientKey='', params='',
                 alreadyRunning=False, minimumWorkers=0, maximumWorkers=0,
                 jobType='', priority=DEFAULT_JOB_PRIORITY):
@@ -1435,7 +1425,7 @@ class ClientJobsDAO(object):
     return jobID
 
 
-  @logExceptions(_getLogger)
+  @logExceptions(_LOGGER)
   def jobInsertUnique(self, client, cmdLine, jobHash, clientInfo='',
                       clientKey='', params='', minimumWorkers=0,
                       maximumWorkers=0, jobType='',
@@ -1580,7 +1570,7 @@ class ClientJobsDAO(object):
     return
 
 
-  @logExceptions(_getLogger)
+  @logExceptions(_LOGGER)
   def jobStartNext(self):
     """ For use only by Nupic Scheduler (also known as ClientJobManager) Look
     through the jobs table and see if any new job requests have been
@@ -1608,7 +1598,7 @@ class ClientJobsDAO(object):
     return jobID
 
 
-  @logExceptions(_getLogger)
+  @logExceptions(_LOGGER)
   @g_retrySQL
   def jobReactivateRunningJobs(self):
     """ Look through the jobs table and reactivate all that are already in the
@@ -1628,7 +1618,7 @@ class ClientJobsDAO(object):
     return
 
 
-  @logExceptions(_getLogger)
+  @logExceptions(_LOGGER)
   def jobGetDemand(self,):
     """ Look through the jobs table and get the demand - minimum and maximum
     number of workers requested, if new workers are to be allocated, if there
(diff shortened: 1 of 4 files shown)


src/nupic/database/ClientJobsDAO.py
```diff
@@ -55,8 +55,7 @@ def _getLogger():
   needs to be initialized by the host app, and that usually happens after
   imports
   """
-  return logging.getLogger(
-    ".".join(['com.numenta', _MODULE_NAME, ClientJobsDAO.__name__]))
+  return logging.getLogger(__name__)
```
SaganBolliger: Change suggested by @scottpurdy to simplify logging

scottpurdy: Actually I think it would be better to make a module-level constant so it doesn't keep getting re-created.

src/nupic/support/decorators.py
```diff
@@ -195,8 +194,8 @@ def myFunctionFoo():
   assert isinstance(retryExceptions, tuple), (
     "retryExceptions must be tuple, but got %r") % (type(retryExceptions),)
   
-  if getLoggerCallback is None:
-    getLoggerCallback = logging.getLogger
+  if logger is None:
+    logger = logging.getLogger(__name__)
```
scottpurdy: again, it may make sense to create this once at the module level rather than re-creating it each function call.

src/nupic/support/pymysqlhelpers.py
```diff
@@ -90,7 +90,7 @@
 
 
 
-def retrySQL(timeoutSec=60*5, getLoggerCallback=logging.getLogger):
+def retrySQL(timeoutSec=60*5, getLoggerCallback=None):
```
scottpurdy: Can we update this to take a logger instead of a callback for consistency?

src/nupic/database/ClientJobsDAO.py
vitaly-krugl: @scottpurdy, @SaganBolliger: as far as I know, `logging.getLogger(...)` caches them in a dictionary by logging context name, so it won't create a new logger if one with the same context name already exists.

src/nupic/support/pymysqlhelpers.py
vitaly-krugl: +1

src/nupic/support/pymysqlhelpers.py
```diff
@@ -113,28 +113,33 @@ def retrySQL(timeoutSec=60*5, getLoggerCallback=logging.getLogger):
     def jobInfo(self, jobID):
         ...
   """
-  
+
+  if getLoggerCallback:
```
vitaly-krugl: If the default value is None, then this check should explicitly check `if getLoggerCallback is not None`, instead of relying on coercion to bool. This is because in python objects (including callables) may override the bool coercion to attain some application-specific esoteric meaning (e.g., evaluates to false if it's never been called). So, if `None` is the documented default value designator, then check explicitly for `None`.

vitaly-krugl:
I would also expect this change to impact unit tests of the affected modules, but I don't see any test changes here. So, either the tests are missing (and need to be added), or the tests are not testing the passing vs. not passing of the logger arg (but they should), or the tests will be broken and need to be fixed.

vitaly-krugl:
I am done with my review, and there is feedback.

src/nupic/database/ClientJobsDAO.py
SaganBolliger: @scottpurdy The reason I held off doing this is that the logExceptions decorator also expects a callback, requiring us to redefine and change all references to the decorator as well. Though, looking into it, it appears that other than the test for that decorator, this file is the only time it is used. Should I go ahead and change it too? I suppose the other approach would be to use a lambda when calling it in this file.

vitaly-krugl: We should make them consistent.

SaganBolliger:
@vitaly-krugl There are currently no tests for the retry decorator in nupic; however #2670 is waiting for review and would add unit tests from nta.utils.

vitaly-krugl:
@SaganBolliger: I'll review @2670 shortly

src/nupic/support/decorators.py
SaganBolliger: From https://docs.python.org/2/library/logging.html: "Multiple calls to getLogger() with the same name will always return a reference to the same Logger object."

SaganBolliger:
@scottpurdy @vitaly-krugl Feedback implemented, and commits squashed. Should be ready for merge.

scottpurdy:
@vitaly-krugl - Merge when you are happy with the changes.

SaganBolliger:
WARNING: numenta/numenta-apps#433 must be merged first or else this PR will break numenta-apps.

src/nupic/support/decorators.py
```diff
@@ -49,16 +47,14 @@ def myFunctionFoo():
         raise RuntimeError("something bad happened")
         ...
   """
+  logger = (logger if logger is not None else logging.getLogger())
```
vitaly-krugl: Consistency: you used `logging.getLogger(__name__)` in the other modification in this module

src/nupic/support/decorators.py
```diff
@@ -159,54 +155,52 @@ def retry(timeoutSec, initialRetryDelaySec, maxRetryDelaySec,
...
 
   Usage Example:
     NOTE: logging must be initialized *before* any loggers are created, else
       there will be no output; see nupic.support.initLogging()
-    
-    _retry = retry(timeoutSec=300, initialRetryDelaySec=0.2, maxRetryDelaySec=10,
-                   retryExceptions=[socket.error])
+
+    _retry = retry(timeoutSec=300, initialRetryDelaySec=0.2,
+                    maxRetryDelaySec=10, retryExceptions=[socket.error])
```
vitaly-krugl: Indentation here doesn't look right

src/nupic/support/decorators.py
```diff
@@ -214,50 +208,47 @@ def retryWrap(*args, **kwargs):
           result = func(*args, **kwargs)
         except retryExceptions, e:
           if not retryFilter(e, args, kwargs):
-            logger = getLoggerCallback()
+            # logger = getLoggerCallback()
```
vitaly-krugl: Remove dead code

vitaly-krugl:
@SaganBolliger: I completed the code review. Please see my feedback.

SaganBolliger:
@vitaly-krugl I've made the changes in your review. This passes all tests on my system except for one regression test that fails and one that errors; however, these same tests fail/error on master, so I think it's good to go.

vitaly-krugl:
@SaganBolliger: please share the complete failure/error/traceback information for the failing/erroring tests.

vitaly-krugl:
@SaganBolliger: your changes look good to me, but I am concerned about the failing tests. Please see my prior comment.

SaganBolliger:
@vitaly-krugl I've included the test output that fails both on this branch and on master on my computer. Travis doesn't run the regression tests, which these are.

```
================================================================================================== ERRORS ===================================================================================================
_________________________________________________________________________ ERROR at teardown of OPFBenchmarkRunner.testOPFBenchmarks _________________________________________________________________________

self = <run_opf_benchmarks_test.OPFBenchmarkRunner testMethod=testOPFBenchmarks>

    def tearDown(self):
      if self.isSaveResults:
>       self.saveResults()

../../tests/regression/run_opf_benchmarks_test.py:1379: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <run_opf_benchmarks_test.OPFBenchmarkRunner testMethod=testOPFBenchmarks>

    def saveResults(self):
>     outpath = os.path.join(self.outdir, "BenchmarkResults.csv")

../../tests/regression/run_opf_benchmarks_test.py:509: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

a = None, p = ('BenchmarkResults.csv',), path = None, b = 'BenchmarkResults.csv'

    def join(a, *p):
        """Join two or more pathname components, inserting '/' as needed.
        If any component is an absolute path, all previous path components
        will be discarded.  An empty last part will result in a path that
        ends with a separator."""
        path = a
        for b in p:
            if b.startswith('/'):
                path = b
>           elif path == '' or path.endswith('/'):
E           AttributeError: 'NoneType' object has no attribute 'endswith'

/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/posixpath.py:70: AttributeError
================================================================================================= FAILURES ==================================================================================================
___________________________________________________________________________________ OPFBenchmarkRunner.testOPFBenchmarks ____________________________________________________________________________________

self = <run_opf_benchmarks_test.OPFBenchmarkRunner testMethod=testOPFBenchmarks>

    def testOPFBenchmarks(self):
      """Run the entire set of OPF benchmark experiments
        """
      # Check for benchmark misspellings
>     for bm in self.listOfBenchmarks:
E     AttributeError: 'OPFBenchmarkRunner' object has no attribute 'listOfBenchmarks'

../../tests/regression/run_opf_benchmarks_test.py:1292: AttributeError
======================================================================= 1 failed, 933 passed, 47 skipped, 1 error in 4985.10 seconds ========================================================================
```

vitaly-krugl:
@rhyolight: is there anything in nupic CI that runs the regression tests? I feel somewhat insecure if we have failing tests in nupic that don't get detected by automation. See @SaganBolliger's reference to failing regression tests above.

CC @scottpurdy

rhyolight:
@vitaly-krugl Yes, regression tests are run after every push to nupic/master. See https://travis-ci.org/numenta/nupic.regression/builds. However, they are not run against pull requests (regression tests are by nature long-running tests). 

Many companies run nightly regression tests, which highlight failures once a day. Ours at least will run after each push to master, and when they fail an email is sent to nupic-hackers.

rhyolight:
BTW, the regression tests will only run after the master build passes, so as soon as https://travis-ci.org/numenta/nupic/builds/87754543 passes, a new regression test will run (and fail, if what @SaganBolliger says is correct).

vitaly-krugl:
@rhyolight, what's the URL for viewing recent runs of the regression test?

rhyolight:
@vitaly-krugl Same one I posted above: https://travis-ci.org/numenta/nupic.regression/builds

SaganBolliger:
@vitaly-krugl @rhyolight nupic.regression does not appear to run the regression tests under nupic/tests/regression. In particular searching for the failed test file, run_opf_benchmarks_test, returns no results in the build log posted above.

vitaly-krugl:
That's a good point, @SaganBolliger. I don't see testOPFBenchmarks in the nupic.regression build logs.

vitaly-krugl:
@SaganBolliger: on a related note, I noticed that `run_opf_benchmarks_test.py` accepts command-line args that specify the locations of various directories of interest. If you haven't already, try running it with `--help` or just look up the arg parsing code in the test's script.

rhyolight:
I believe the `run_opf_benchmarks_test.py` should be run in regression tests. Can you guys create a new issue for it?

Issue #2678, opened 2015-10-14
Implement C++ capnp struct to pycapnp conversion


chetan51:
Complementary issue for https://github.com/numenta/nupic.core/issues/545.

PR #2679, opened 2015-10-15
Add methods for read and write of PyRegion

chetan51:
Fixes https://github.com/numenta/nupic/issues/2678.

### DIFF
--- a/src/nupic/regions/PyRegion.py
+++ b/src/nupic/regions/PyRegion.py
@@ -273,6 +273,25 @@ class PyRegion(object):
     array: the value to assign to the requested parameter (a numpy array)
     """
 
+
+  def write(self, proto):
+    """This method is called during network serialization with a proto object
+    so that this region can write its data to it.
+
+    proto: PyRegionProto capnproto object
+    """
+    pass
+
+
+  def read(self, proto):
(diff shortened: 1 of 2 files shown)


chetan51:
@scottpurdy

src/nupic/regions/PyRegion.py
```diff
@@ -273,6 +273,25 @@ def setParameterArray(self, name, index, array):
...
     """
 
+
+  def write(self, proto):
+    """This method is called during network serialization with a proto object
+    so that this region can write its data to it.
+
+    proto: PyRegionProto capnproto object
+    """
+    pass
```
scottpurdy: Should this raise NotImplemented?

scottpurdy:
Should we just delete the TestNode in `nupic` since we have one in the bindings?

src/nupic/regions/PyRegion.py
chetan51: I don't think so, because there can be subclasses of this that don't need serialization.

scottpurdy:
:+1:

chetan51:
The `TestNode` in `nupic` is used in `network_test.py`.

PR #2695, opened 2015-10-26
Invalid csv files

### LINKED ISSUE #2694: Invalid lines ending in three CSV files
Three CSV files have invalid line endings. The Git automatic newline handling option is set correctly in NuPIC. But the CSV files end up with bad line endings that concatenations all the lines in the file. Even though Github website can interpret and show the files properly. 
Research shows that those files may have been saved from Excel for MacOS that may only have used a CR, rather than usual LF, or CR with LF (carriage return, line feed).
Ref; http://stackoverflow.com/questions/1552749/difference-between-cr-lf-lf-and-cr-line-break-types

rcrowder:
Fixes #2694 This change helps pass tests that use these CSV files.


rcrowder:
@scottpurdy Ready for review.

rhyolight:
:+1:

scottpurdy:
The files are missing newline characters at the end, which I believe CSV files should have. Otherwise looks good.

rcrowder:
@scottpurdy Ready.

PR #2697, opened 2015-10-26
Adding cmd support

### LINKED ISSUE #2696: Add Windows cmd line support alongside bash

rcrowder:
Fixes #2696 Bash only found in two places. This adds Windows command line support alongside them.


rcrowder:
@scottpurdy Ready for review.

rhyolight:
:+1:

rcrowder:
For reference; https://technet.microsoft.com/en-gb/library/bb490880.aspx

scottpurdy:
Why does it execute the command via bash instead of directly executing the command? I think we should be able to remove the bash/cmd parts.

rcrowder:
No idea. Haven't got to the swarming tests yet. So unsure of what permutation runner is used for. I;m guessing so that the process has a consistent environment?

scottpurdy:
@rcrowder - I think `subprocess.call` and similar have a "shell" boolean option that should do the same thing. Can you try removing the bash/cmd and pass the command directly to the subprocess function and see if that still works? I'm happy to debug if issues come up with that.

rcrowder:
Sure, I'll have a look now.

rcrowder:
@scottpurdy Sorry for the delay. My Windows NuPIC is in a state of flux. And ```mysql-server``` failed install in my Ubuntu 14.04 VM, then updates to Linux and recent Oracle releases to VirtuaBox, broken the VM (no mouse cursor!). Eventually regaining mouse integration after Oracle's changes to how it works!

Changing ```args``` to ```cmdLine``` and passing ```Shell=True``` works fine with swarming tests. But as mentioned in links below has security risks (vunerable to shell injection). Not committed changes yet, preferring to hear from you on whether those risks are important or not. Probably not, but wanted to check first.

Using ```Shell=False``` and ```cmdLine``` fails quickly.

https://docs.python.org/2/library/subprocess.html#using-the-subprocess-module
https://docs.python.org/2/library/subprocess.html#frequently-used-arguments

scottpurdy:
@rcrowder - AFAIK we aren't passing user-specific arguments in those commands. Even if we are, they are coming from the person running the code so they are only hurting themselves if they inject something malicious. In other words, don't worry about it here.

scottpurdy:
@rcrowder - Actually, now that I look at the code I don't think we even need to set `shell=True`. Can you remove that and see if it works?

rcrowder:
@scottpurdy I thought it'd be ok, and not pose a risk. 
The default with ```subprocess.call``` is ```Shell=False```, that asserts when set to false. And also has ```, stdin=None, stdout=stdout, stderr=stderr)``` following the Shell parameter (see below for assert). Only option seems to be passing cmdLine and Shell=True.

Using ```subprocess.Popen(cmdLine, bufsize=1, env=os.environ)``` produces...
  File "/home/nupic/nupic/src/nupic/swarming/permutations_runner.py", line 634, in _launchWorkers
    p = subprocess.Popen(cmdLine, bufsize=1, env=os.environ)
  File "/usr/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1327, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory

scottpurdy:
Let me take a look, will follow up in a sec

scottpurdy:
Ok I understand the issue better. The problem is included environment variables before the executable in the argument list. So if `shell=True` works then lets do that. If not then just leave the bash/cmd logic.

rcrowder:
@scottpurdy Ready now.

PR #2707, opened 2015-10-28
Updating test_db.py output comments

### LINKED ISSUE #2676: MySQL setup and Swarming
I'm trying to follow the Wiki to run swarming tests (https://github.com/numenta/nupic/wiki/Running-Swarms). This is on a clone of master that has been setup, but I missed install of MySQL so I've just installed that. Not sure if I have to run setup.py again?

MySQL didn't accept a blank root password during installation, so I added a simple one. Running ```$NUPIC/examples/swarm/test_db.py``` fails to connect, and says to look at ```$NUPIC/nupic/support/nupic-default.xml```. That file has been moved into ```$NUPIC/src/nupic/support```. 

The xml says not to edit the file but make a copy called ```nupic-site.xml``` and edit it. I'm not sure where to put this new XML file, or whether I should delete everything I don't need (just need to change root passwd) and just leave the bits I want? 
If I delete things should the XML format be valid? (start tags, configuration and the property to change, with associated ```/``` closing tags) Or does this site XML get inserted into the default one?

rcrowder:
Fixes #2676 @rhyolight Ready for review.

### DIFF
--- a/examples/swarm/test_db.py
+++ b/examples/swarm/test_db.py
@@ -30,10 +30,16 @@ passwd = Configuration.get('nupic.cluster.database.passwd')
 
 print 
 print "This script will validate that your MySQL is setup correctly for NuPIC."
-print "MySQL is required for NuPIC swarming. The settings are defined in "
-print "NuPIC's configuration files in $NUPIC/nupic/support/nupic-default.xml "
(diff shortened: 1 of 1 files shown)


rhyolight:
thanks! :+1:

PR #2713, opened 2015-10-29
fix rawOverlap calc, add pctOverlapOfInput calc

### LINKED ISSUE #2712: KNNClassifier distance calculations
The [`rawOverlap` distance metric](https://github.com/numenta/nupic/blob/master/src/nupic/algorithms/KNNClassifier.py#L791-797) in the KNNClassifier does not calculate distance as specified [in the docstring](https://github.com/numenta/nupic/blob/master/src/nupic/algorithms/KNNClassifier.py#L102-104). Rather, it calculates a percent overlap of the input.

BoltzmannBrain:
Fixes #2712 by correcting the `rawOverlap` metric, and adding a `pctOverlapOfInput` metric.

### DIFF
--- a/src/nupic/algorithms/KNNClassifier.py
+++ b/src/nupic/algorithms/KNNClassifier.py
@@ -99,6 +99,8 @@ class KNNClassifier(object):
         "norm": When distanceNorm is 2, this is the euclidean distance,
                 When distanceNorm is 1, this is the manhattan distance
                 In general: sum(abs(x-proto) ^ distanceNorm) ^ (1/distanceNorm)
+                The distances are normalized such that farthest prototype from
+                a given input is 1.0.
         "rawOverlap": Only appropriate when inputs are binary. This computes:
                 (width of the input) - (# bits of overlap between input
                 and prototype).
@@ -787,7 +789,7 @@ class KNNClassifier(object):
         self._protoSizes = self._Memory.rowSums()
       overlapsWithProtos = self._Memory.rightVecSumAtNZ(inputPattern)
       inputPatternSum = inputPattern.sum()
-    
+
       if self.distanceMethod == "rawOverlap":
         dist = inputPattern.sum() - overlapsWithProtos
       elif self.distanceMethod == "pctOverlapOfInput":
@@ -939,7 +941,7 @@ class KNNClassifier(object):
 
     self._vt = self._vt[:self.numSVDDims]
 
-    # Added when svd is not able to decompose vectors - uses raw spare vectors  
+    # Added when svd is not able to decompose vectors - uses raw spare vectors
     if len(self._vt) == 0:
       return
 
--- a/src/nupic/algorithms/KNNClassifier.py
+++ b/src/nupic/algorithms/KNNClassifier.py
@@ -789,7 +789,7 @@ class KNNClassifier(object):
       inputPatternSum = inputPattern.sum()
     
       if self.distanceMethod == "rawOverlap":
-        dist = (inputPattern.sum() - self._Memory.rightVecSumAtNZ(inputPattern))
+        dist = inputPattern.sum() - overlapsWithProtos
       elif self.distanceMethod == "pctOverlapOfInput":
         dist = inputPatternSum - overlapsWithProtos
         if inputPatternSum > 0:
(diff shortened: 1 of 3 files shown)


subutai:
If you are correcting the overlap calculation, please include at least one test case that highlights the use case that is now fixed.

BoltzmannBrain:
@subutai wasn't ready for review yet :), but now is.

BoltzmannBrain:
Also, the `rawOverlap` still isn't necessarily a raw overlap. As the docstring specifies, it's `(input width) - (# overlapping bits)`. Wouldn't a raw overlap just be `(# of overlapping bits)`?

subutai:
> Also, the rawOverlap still isn't necessarily a raw overlap. As the docstring specifies, it's (input width) - (# overlapping bits). Wouldn't a raw overlap just be (# of overlapping bits)?

Yes, it should be, but with KNN it picks the closest vectors (i.e. smallest distance score). Using (input width) - (# overlapping bits) makes it so that the lower scores have higher overlap.

src/nupic/algorithms/KNNClassifier.py
```diff
@@ -95,19 +95,22 @@ class labels
...
                 When distanceNorm is 1, this is the manhattan distance
                 In general: sum(abs(x-proto) ^ distanceNorm) ^ (1/distanceNorm)
         "rawOverlap": Only appropriate when inputs are binary. This computes:
                 (width of the input) - (# bits of overlap between input
                 and prototype).
-        "pctOverlapOfLarger": Only appropriate for binary inputs. This computes
+        "pctOverlapOfInput": Only appropriate for binary inputs. This computes
                 1.0 - (# bits overlap between input and prototype) /
-                        max(# bits in input, # bits in prototype)
+                        (# ON bits in input)
```
subutai: Don't you mean ((width of input) - (# bits of overlap)) / (width of input)?

tests/unit/nupic/algorithms/knn_classifier_test.py
```diff
@@ -30,6 +30,67 @@
...
+    protoB = np.array([20, 28, 30], dtype=np.int32)
+    
+    classifier.learn(protoA, 0, isSparse=dimensionality)
+    classifier.learn(protoB, 0, isSparse=dimensionality)
+    
+    input = np.zeros(dimensionality)
+    input[:4] = 1.0
+    
+    # Test l2 norm metric
+    _, _, dist, _ = classifier.infer(input)
```
subutai: Nice test cases. Please also include edge cases, i.e. when distance should be 0 or the largest possible.

subutai:
Left some feedback. Otherwise :+1:

src/nupic/algorithms/KNNClassifier.py
BoltzmannBrain: No because you're first dividing by the width of input, giving a value between 0.0 (no overlapping bits) and 1.0 (all bits of the input overlap). We want to flip this, where the minimum distance is maximal overlap, so we subtract from 1.0.

BoltzmannBrain:
@subutai I addressed your comments. Please merge if it's good to go.

BoltzmannBrain:
@scottpurdy Subutai may have left for his vacation, so please take a look.

BoltzmannBrain:
or @rhyolight

tests/unit/nupic/algorithms/knn_classifier_test.py
```diff
@@ -30,6 +30,95 @@
...
+    # input0 is used to test that the distance from a point to itself is 0
+    input0 = np.zeros(dimensionality)
+    input0[protoA] = 1.0
+
+    # Test l2 norm metric
+    _, _, dist, _ = classifier.infer(input)
+    l2Distances = [0.65465367,  1.0]
+    for actual, predicted in zip(l2Distances, dist):
+      self.assertAlmostEqual(actual, predicted, places=5,
+        msg="l2 distance norm is not calculated as expected.")
```
scottpurdy: If you want to break like this then start the first argument on a new line. See [PEP-8](https://www.python.org/dev/peps/pep-0008/#indentation).

scottpurdy:
:+1:

BoltzmannBrain:
Thanks @scottpurdy, please merge if ready.

PR #2723, opened 2015-11-02
Completely build nupic bindings from source in Dockerfile

### LINKED ISSUE #2721: Docker build needs to install via compile from source instructions
Currently it uses a mix of binary for nupic.bindings and source from nupic.

oxtopus:
Fixes #2721

Previous approach of mixing `pip install` with `python setup.py install` was resulting in a broken installation with respect to namespace packages.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,19 @@ MAINTAINER Allan Costa <allaninocencio@yahoo.com.br>
 
 # Install dependencies
 RUN apt-get update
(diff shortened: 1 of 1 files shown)


lscheinkman:
:+1: It worked for me.
Tested using 
```
docker build -t numenta/nupic:2723 .
docker run -ti numenta/nupic:2723  python -c 'from nupic.encoders.scalar import ScalarEncoder; print ScalarEncoder'
```

rhyolight:
Nice :+1: Thanks!

Issue #2271, opened 2015-06-29
KNNClassifier improvements


rmccall84:
Improvements to KNNClassifier

- [ ] infer() should have options for sparse and dense inputs. Right now, for example, if distance metric is 'rawOverlap' infer must be called with a dense input even if classifier was trained on sparse input repr. If sparse input is used, the classification is non-deterministic
- [ ] Add getPatternCount() accessor method, possibly other similar methods for a nice API
- [ ] Understand `partitionId` and implement, remove, or document appropriately
- [ ] Understand and then remove or document `rowID`
- [ ] Move `leaveOneOutTest` method from KNNClassifier module to an appropriate test module or delete
- [ ] Remove unused arguments from infer() method
- [ ] Move `leaveOneOutTest` to another module
- [ ] SVD / PCA features need better docs
- [ ] Reorder parameters in constructor putting dependent parameters after their dependencies

rmccall84:
@mihail911 @BoltzmannBrain FYI, recreating the tasks here since old issue was closed

BoltzmannBrain:
Thank you @rmccall84.
When pushing PRs for these tasks, please link with "addresses #2271"; this way the issue won't close :smile:

breznak:
> Modify KNN 'infer' so that client has option of outputting top n most frequent categories rather than just the most frequent

It would be nice if the `n` predictions could be sorted by probability. We had an issue on this, it's not possible to sort a dict by its values, so this would need some hacking. 
Ie. `print infs` -> `{'a': 0.8, 'd': 0.5, 'b': 0.1, 'c': 0.1}`

BoltzmannBrain:
@breznak in progress in #2284, specifically [here](https://github.com/mihail911/nupic/blob/augmentKNNClassifier/nupic/algorithms/KNNClassifier.py#L632) :smile:

breznak:
@BoltzmannBrain very nice! :+1: :+1: I'm just running some data and wondering why it looks much better :wink:

SaganBolliger:
Another issue: removeCategory() seems to be broken. Raises `AttributeError: 'KNNClassifier' object has no attribute '_categoryRecencyList'`

SaganBolliger:
There also seems to be an issue with infer having a side effect when using partitionId. In particular, this unit test fails:

```
    params = {"distanceMethod": "rawOverlap"}
    classifier = KNNClassifier(**params)

    dimensionality = 40
[...]

PR #2725, opened 2015-11-02
RES-112 Documented partitionId in KNNClassifier and added unit test.

### LINKED ISSUE #2730: Understand partitionId and implement, remove, or document appropriately
This is a sub-issue addressing the point "Understand partitionId and implement, remove, or document appropriately" from #2271. Specifically, several methods in the [KNNClassifier](src/nupic/algorithms/KNNClassifier.py) have a parameter partitionId, which is undocumented. This parameter needs to either be documented or removed.

SaganBolliger:
In partial fulfillment of https://jira.numenta.com/browse/RES-112

Fixes #2730 (a sub-issue of #2271). 

@BoltzmannBrain Ready for review

### DIFF
--- a/src/nupic/algorithms/KNNClassifier.py
+++ b/src/nupic/algorithms/KNNClassifier.py
@@ -361,7 +361,11 @@ class KNNClassifier(object):
     @param inputCategory (int) The category to be associated to the training
         pattern
 
-    @param partitionId (int) UNKNOWN
+    @param partitionId (int) partitionID allows you to partition the data set
+        by associating unique IDs with sets of vectors. One use case is to
+        ignore a specific set of vectors during inference for k-fold cross
+        validation (see description of infer() for further details).
+        This is an optional parameter.
 
     @param isSparse (int) If 0, the input pattern is a dense representation. If
         isSparse > 0, the input pattern is a list of non-zero indices and
@@ -379,9 +383,6 @@ class KNNClassifier(object):
 
     if rowID is None:
       rowID = self._iterationIdx
-
-    assert partitionId is None, \
-      "No documentation is available for partitionId, not sure how it works."
 
     # Dense vectors
     if not self.useSparseMemory:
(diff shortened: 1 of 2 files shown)


src/nupic/algorithms/KNNClassifier.py
```diff
@@ -356,7 +356,9 @@ def learn(self, inputPattern, inputCategory, partitionId=None, isSparse=0,
     @param inputCategory (int) The category to be associated to the training
         pattern
 
-    @param partitionId (int) UNKNOWN
+    @param partitionId (int) The k-fold cross validation partition for this
```
BoltzmannBrain: Why specific to k-folds? It seems this is just an example use for `partitionId`.

SaganBolliger: I'm pretty sure that is the intended use including the using standard k-folds terminology of "partition." One may be able to find a different way of using the functionality, but I expect that such alternate uses would be much less common.

BoltzmannBrain: I agree, but my point is we shouldn't assume an implementation :wink: 
So maybe we can instead say `During inference all training vectors with partitionId equal to that of the test vector are ignored, e.g. when switching between train and test partitions in k-folds cross validation.`

SaganBolliger:
@BoltzmannBrain Ok, I've changed the wording not to assume a use case.

BoltzmannBrain:
LGTM
@subutai or @scottpurdy please take a look and merge.

subutai:
I think we should describe the meaning of PartitionID. What does it represent?  Currently it just mentions one specific use case. 

Also I think we need to link this to an issue.

SaganBolliger:
@subutai Doesn't "During inference all training vectors with partitionId equal to that of the test vector are ignored" describe the meaning of the parameter? This is all the parameter does. The k-fold cv is the most likely and intended use case, although someone could conceivably use it for another purpose. 

I've created an issue as a sub-issue of the originally linked issue that can be safely referenced via "fixes" so that all the validators pass.

SaganBolliger:
@scottpurdy @subutai I've expanded the description and provided an example for use in k-fold cross validation. Hopefully this clears up any ambiguity.

BoltzmannBrain:
ping @subutai @scottpurdy

scottpurdy:
Fine with me once @subutai is happy with the description

src/nupic/algorithms/KNNClassifier.py
```diff
@@ -361,7 +361,9 @@ def learn(self, inputPattern, inputCategory, partitionId=None, isSparse=0,
     @param inputCategory (int) The category to be associated to the training
         pattern
 
-    @param partitionId (int) UNKNOWN
+    @param partitionId (int) During inference all training vectors with
+        partitionId equal to that of the test vector are ignored.
+        See description of parameter at infer() for further details.
```
subutai: We should have a helpful higher level description of what partition ID is because it actually has multiple uses. Something like "partitionID allows you to partition the data set and associate unique IDs with sets of vectors. One use case is to ignore a specific set of vectors during inference for k-fold cross validation (see description of infer() for further details). This is an optional parameter."

subutai:
:+1: other than the comment I left.

SaganBolliger:
@scottpurdy I've change the docstring for learn() to Subutai's suggestion. Should be good to merge.

PR #2734, opened 2015-11-05
Windows logging and line endings

### LINKED ISSUE #2735: Integration tests failing on Windows
Down to five failing integration tests on the Windows build.

rcrowder:
Fixes #2735 

@scottpurdy Here's a PR for the last couple of weeks chipping away at Windows issues. Including necessary logging (probably temp.) support, and handling ASCII/text file handling. Plus a way for me to try these changes on Mac OS (via Travis). Unit tests are still failing (more with pycapnp TestNode ones), but Integration tests have drop to five failing now. Feels like one more sprint and GH commit could get the Integration tests all passing. The logging change naturally is an important one too.


rcrowder:
Output of stdout and stderr on Windows can be done using the following, and ties into the Gist stated below;
> py.test tests\integration 2>&1 > integration_tests.log

https://gist.github.com/rcrowder/ab7209ff99b62774f572

rcrowder:
Last error, as seen in the Gist, looks like it's related to a potential major problem? Further info here; http://stackoverflow.com/questions/556269/importerror-no-module-named-copy-reg-pickle

rcrowder:
All swarming tests now passing with the last commit - https://gist.github.com/rcrowder/f547fbad3ed45cf87bb0

src/nupic/data/file_record_stream.py
```diff
@@ -62,6 +62,7 @@
 """
 
 import os
+import sys
```
scottpurdy: this isn't used

src/nupic/support/__init__.py
```diff
@@ -431,13 +431,18 @@ def makeKey(name):
     replacements[makeKey('SYSLOG_HANDLER_ADDRESS')] = '"/var/run/syslog"'
   elif platform.startswith('linux'):
     replacements[makeKey('SYSLOG_HANDLER_ADDRESS')] = '"/dev/log"'
+  elif platform.startswith('win'):
+    replacements[makeKey('SYSLOG_HANDLER_ADDRESS')] = '"log"'
```
scottpurdy: The changes in this file appear to swtich from absolute to relative paths. Does this mean the log file location will depend on the working directory the code is run from?

src/nupic/data/file_record_stream.py
rcrowder: Noted. Relic from old code.

src/nupic/support/__init__.py
rcrowder: Yes, at the moment that happens. On my list to look further at this, the current log files are empty so further work needed to find out why (and where to put them).

scottpurdy: @vitaly-krugl - thoughts on this? In general, this logging code is pretty hacky compared to what we have since implemented for use in numenta-apps.

src/nupic/swarming/permutations_runner.py
```diff
@@ -367,7 +367,10 @@ def _setUpExports(exports):
     return ret
   exportDict = json.loads(exports)
   for key in exportDict.keys():
-    ret+= "export %s=%s;" % (str(key), str(exportDict[key]))
+    if (sys.platform.startswith('win')):
```
scottpurdy: Yuck, we really need to rewrite this code to properly set up the environment rather than passing on command line. This is fine for this PR though.

tests/integration/nupic/opf/hotgym_regression_test.py
```diff
@@ -42,7 +42,7 @@ class HotgymRegressionTest(unittest.TestCase):
 
   def testHotgymRegression(self):
     experimentDir = os.path.join(
-      os.path.dirname(__file__).partition("tests/integration/nupic/opf")[0],
+      os.path.dirname(__file__).partition(os.path.normpath("tests/integration/nupic/opf"))[0],
```
scottpurdy: wrap to stay under 80 characters

scottpurdy:
Done reviewing. My only concern is the relative logging path for Windows but we should just get this merged and keep moving on. Let me know when you're ready for it to be merged. You can fix the rest of the tests in a follow up PR to prevent this one from getting really big.

src/nupic/support/__init__.py
rcrowder: Sorry, a late in the day reply. My comment referred to the nupic.log (file handler one). A lot of temporary files, such as those produced via py.test runs, get placed into a Users hidden AppData temp directory (with semi-random directory names to avoid clashes). We'll have a look at the sys log handling, in another PR.

src/nupic/swarming/permutations_runner.py
rcrowder: kk, I'll raise an issue. That can then be assigned to someone to address. And once this PR is merged, others can see the Windows changes here to help determine the best foot forward. Another one is the adding of "python.exe" to the command line that has caught a few people out (Ryan McCall recently spring to mind).

src/nupic/support/__init__.py
rcrowder: @vitaly-krugl For reference on a Windows sys logger see https://docs.python.org/2/library/logging.handlers.html#nteventloghandler I'd be interested to here from your experience on potential issues with sys log handlers.

rcrowder:
@scottpurdy This one is ready to merge now. With other issues already raised, or to be raised, to cover the failing tests and log handlers.

src/nupic/support/__init__.py
vitaly-krugl: @scottpurdy: there are a few things going on here that are not quite right. 

1. It's better to remove the SYSLOG_HANDLER altogether from this script and from nupic-logging.conf, because we actually don't make use of it due to critical limitations. At some point, we experimented with SysLogHandler on Linux and Mac OS X. We found that it didn't work  with long log lines that we often need for debugging, such as capturing a deep traceback. We experimented and found that the two platforms differed in how they handled "overflow": one of them would drop the log message, while the other would truncate at the limit. So, we decided against using them. We must have forgotten to remove the SYSLOG_HANDLER references from this script and from nupic-logging.conf. You can see that `nupic.support.initLogging()` doesn't actually add SysLogHandler to the `handlers` list, so it's not going to be used.
2. SysLogHandler is not supported on windows, but the above line recommends to get rid of this stuff, anyway. If we really need to support SysLogHandler and NTEventLogHandler-like functionality, then we would need an additional effort to make it work right. If there is no such requirement, let's just drop this SysLogHandler/NEventLogHandler functionality altogether.
 ```
  if platform.startswith('darwin'):
    replacements[makeKey('SYSLOG_HANDLER_ADDRESS')] = '"/var/run/syslog"'
  elif platform.startswith('linux'):
    replacements[makeKey('SYSLOG_HANDLER_ADDRESS')] = '"/dev/log"'
  elif platform.startswith('win'):
    replacements[makeKey('SYSLOG_HANDLER_ADDRESS')] = '"log"'
  else:
    raise RuntimeError("This platform is neither darwin, win32, nor linux: %s" % (
      sys.platform,))
```
3. The first block of code below was intended to initialize file logging in the suppressed state; hence, `/dev/null`. However, the `'win'` initialization doesn't match that intent and needs to be fixed. It's only in the `if configLogDir is not None:` block that we should be enabling file logging.
 ```
  # Nupic logs go to file
  replacements[makeKey('PERSISTENT_LOG_HANDLER')] = 'fileHandler'
  if platform.startswith('win'):
    replacements[makeKey('FILE_HANDLER_LOG_FILENAME')] = '"nupic.log"'
  else:
    replacements[makeKey('FILE_HANDLER_LOG_FILENAME')] = '"/dev/null"'

  # Set up log file path for the default file handler and configure handlers
  handlers = list()
  
  if configLogDir is not None:
    logFilePath = _genLoggingFilePath()
    makeDirectoryFromAbsolutePath(os.path.dirname(logFilePath))
    replacements[makeKey('FILE_HANDLER_LOG_FILENAME')] = repr(logFilePath)

    handlers.append(replacements[makeKey('PERSISTENT_LOG_HANDLER')])
```

rcrowder: @vitaly-krugl Thanks for the detailed reply. I'll make sure aspects are in an other Windows PR. And see what Scott thinks re. the file handler.

scottpurdy: I'll defer to @vitaly-krugl on this stuff. Sounds like it might make sense to remove SysLogHandler.

rcrowder: kk, I'll try out using ```NUL``` on Windows, as the known alternative for ```/dev/null```

vitaly-krugl: @rcrowder: "nul" might do the trick on Windows per https://msdn.microsoft.com/en-us/library/aa939249.aspx.

rcrowder: Yes. Although command line stderr redirection is tricky, needing to use ```2>&1``` to redirect it to stdout (usually alongside ```>``` file redirect, or ```>>``` append)

Issue #2729, opened 2015-11-03
Allow using FastTemporalMemory through OPF


chetan51:
By setting `temporalImp='fast_tm_py'` in model params.

PR #2737, opened 2015-11-05
Allow using FastTemporalMemory through OPF

chetan51:
Fixes https://github.com/numenta/nupic/issues/2729.

### DIFF
--- a/src/nupic/regions/TPRegion.py
+++ b/src/nupic/regions/TPRegion.py
@@ -42,11 +42,13 @@ def _getTPClass(temporalImp):
     return TP10X2.TP10X2
   elif temporalImp == 'tm_py':
     return TP_shim.TPShim
+  elif temporalImp == 'fast_tm_py':
+    return TP_shim.FastTPShim
   elif temporalImp == 'monitored_tm_py':
     return TP_shim.MonitoredTPShim
   else:
     raise RuntimeError("Invalid temporalImp '%s'. Legal values are: 'py', "
-              "'cpp', and 'tm_py'" % (temporalImp))
+              "'cpp', 'tm_py', 'fast_tm_py', 'monitored_tm_py'" % (temporalImp))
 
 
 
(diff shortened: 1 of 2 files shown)


chetan51:
@scottpurdy

src/nupic/regions/TPRegion.py
```diff
@@ -42,11 +42,13 @@ def _getTPClass(temporalImp):
     return TP10X2.TP10X2
   elif temporalImp == 'tm_py':
     return TP_shim.TPShim
+  elif temporalImp == 'fast_tm_py':
```
scottpurdy: how about `tm_cpp` for consistency?

src/nupic/regions/TPRegion.py
```diff
@@ -418,7 +420,7 @@ def initialize(self, dims, splitterMaps):
     if self._tfdr is None:
       tpClass = _getTPClass(self.temporalImp)
 
-      if self.temporalImp in ['py', 'cpp', 'r', 'tm_py', 'monitored_tm_py']:
+      if self.temporalImp in ['py', 'cpp', 'r', 'tm_py', 'fast_tm_py', 'monitored_tm_py']:
```
scottpurdy: wrap line

scottpurdy:
Done reviewing. Eventually I think "TP_shim.py" should be renamed to "tm_shim.py" but this isn't the time to worry about that.

src/nupic/regions/TPRegion.py
chetan51: But then it would be inconsistent with the `TemporalMemory` and `FastTemporalMemory` naming.

chetan51:
@scottpurdy Updated.

src/nupic/regions/TPRegion.py
scottpurdy: In some cases we have had "MyClass" (Python) and "FastMyClass" (C++), like the CLA Classifier. In other cases, we have just had "MyClass" in both and referred to them as either "py" or "cpp" versions. My preference for the implementation identifier would be "py" vs "cpp" since that is the only difference. If you think "fast" is better that is ok but it shouldn't have "py" in the name if it is the C++ implementation.

chetan51: Ah, I didn't realize it would have `py` in the name even though it is using `C++`. I've changed `fast_tm_py` to `tm_cpp`.

chetan51:
@scottpurdy Can I merge after successful build?

scottpurdy:
:+1:

PR #2747, opened 2015-11-09
Fixes unnecessary time complexity in _inhibitColumnsGlobal

### LINKED ISSUE #2746: Unnecessary time complexity in _inhibitColumnsGlobal
In the Python implementation of Spatial Pooler the method to find the global winners does so by [sorting the entire list of activations](https://github.com/numenta/nupic/blob/master/src/nupic/research/spatial_pooler.py#L1427-L1429), and then taking the top k values. However, this is an inefficient approach. With n as the number of columns, sorting in general is O( n log n ), however it is possible to find the top k winners in only O( n ) time complexity. Using numpy, this can be accomplished using the [argpartition](http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.argpartition.html) function. Note that the [nupic.core implementation](https://github.com/numenta/nupic.core/blob/master/src/nupic/algorithms/SpatialPooler.cpp#L1163-L1167) uses a O( n ) approach that avoids sorting all the activations.

@scottpurdy @subutai

SaganBolliger:
Fixes #2746 

Passes all nupic unit and integration tests.

@scottpurdy @subutai

### DIFF
--- a/src/nupic/research/spatial_pooler.py
+++ b/src/nupic/research/spatial_pooler.py
@@ -1422,12 +1422,20 @@ class SpatialPooler(object):
     @return list with indices of the winning columns
     """
     #calculate num active per inhibition area
-
     numActive = int(density * self._numColumns)
-    winners = sorted(range(overlaps.size),
-                     key=lambda k: overlaps[k],
-                     reverse=True)[0:numActive]
-    return winners
+
+    # argpartition returns an array where all values to the left of index
+    # numActive are less than or equal to all values to its right. Negating
+    # overlaps produces a descending-order partition.
+    winnerIndices = numpy.argpartition(-overlaps, numActive)[:numActive]
+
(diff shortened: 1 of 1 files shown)


scottpurdy:
For reference, relevant unit tests are here:
https://github.com/numenta/nupic/blob/6363f21ab1fbf647dcc1c2b5d64a183389fcbac8/tests/unit/nupic/research/spatial_pooler_unit_test.py#L1275-L1295

Have you comparing the difference in running time between the implementations? How much faster is it now?

SaganBolliger:
Yes, I benchmarked it. With 2048 columns and 0.02 density and averaged over 50 calls, the current implementation takes 1.2e-3 s per call, whereas the new implementation takes 1.8e-5 s per call. This is a 65.8x speedup.

@scottpurdy

PR #2749, opened 2015-11-10
Make Cap'n Proto serialization optional in NuPIC.

### LINKED ISSUE #2750: Temporarily make capnp optional for Windows
pycapnp does not currently work on Windows so we should prevent it from being installed for now.

scottpurdy:
@rcrowder - Does this work for Windows for now?
@oxtopus - Please review.

This is intended to be a temporary solution until we get pycapnp working on Windows.

fixes #2750

### DIFF
--- a/ci/travis/install.sh
+++ b/ci/travis/install.sh
@@ -34,7 +34,7 @@ pip --version
 
 # Build NuPIC
 cd ${TRAVIS_BUILD_DIR}
-pip install --user .
+pip install --user .[capnp]
 
 # Show nupic installation folder by trying to import nupic, if works, it prints
 # the absolute path of nupic.__file__, which the installation folder itself.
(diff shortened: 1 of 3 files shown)


rcrowder:
@scottpurdy I tried this only in nupic.core and it sorts out the issue there when installing the resulting wheel. I believe it'll sort out the same issue here too.

rcrowder:
@scottpurdy I'll duplicate this across to nupic.bindings

scottpurdy:
@oxtopus ping

ci/travis/install.sh
```diff
@@ -34,7 +34,7 @@ pip --version
 
 # Build NuPIC
 cd ${TRAVIS_BUILD_DIR}
-pip install --user .
+pip install --user .[capnp]
```
oxtopus: nice!

oxtopus:
:+1: lgtm

Issue #2754, opened 2015-11-12
Make python SP and C++ SP compatible


chetan51:
Bindings for C++ SP methods have been added in https://github.com/numenta/nupic.core/pull/689. Add unit tests for those new methods and make sure these methods are identical to the python versions.

PR #2755, opened 2015-11-12
Make python SP and C++ SP compatible

chetan51:
Fixes https://github.com/numenta/nupic/issues/2754.


chetan51:
@subutai Ready for review after https://github.com/numenta/nupic.core/pull/689 is merged.

chetan51:
Follow-up: https://github.com/numenta/nupic/issues/2774

chetan51:
@scottpurdy Ready for review.

src/nupic/research/spatial_pooler.py
```diff
@@ -1424,16 +1422,10 @@ def _inhibitColumnsGlobal(self, overlaps, density):
...
-    winnerIndices = numpy.argpartition(-overlaps, numActive)[:numActive]
-
-    # Compatibility with nupic.core requires that the winners are sorted;
-    # however, sorting only the winners is far less expensive than sorting all
-    # columns.
-    winnerValues = overlaps[winnerIndices]
-    sortedWinnerIndices = winnerIndices[numpy.argsort(-winnerValues)]
+    # Calculate winners and stable sort for compatibility with C++
+    indices = numpy.array(xrange(self._numColumns), dtype=uintType)
+    winnerIndices = indices[numpy.argsort(overlaps, kind='mergesort')]
```
scottpurdy: Is `kind='mergesort'` responsible for the stable sorting? If so, put a comment before this line stating that that is why it is specified.

scottpurdy:
done reviewing, looks good. We have to get nupic.core in sync before we can merge this though.

src/nupic/research/spatial_pooler.py
chetan51: Good catch, I forgot to include that in the comment. Updated.

src/nupic/research/spatial_pooler.py
```diff
@@ -1424,16 +1422,11 @@ def _inhibitColumnsGlobal(self, overlaps, density):
...
-    # Compatibility with nupic.core requires that the winners are sorted;
-    # however, sorting only the winners is far less expensive than sorting all
-    # columns.
-    winnerValues = overlaps[winnerIndices]
-    sortedWinnerIndices = winnerIndices[numpy.argsort(-winnerValues)]
+    # Calculate winners using stable sort algorithm (mergesort)
+    # for compatibility with C++
+    indices = numpy.array(xrange(self._numColumns), dtype=uintType)
+    winnerIndices = indices[numpy.argsort(overlaps, kind='mergesort')]
+    sortedWinnerIndices = winnerIndices[-numActive:][::-1]
```
SaganBolliger: @chetan51 This was an optimization that I implemented. We discussed how this optimization has to be reversed due to the lack of a stable sort version of argpartition. However, it looks like this reverted version does some unnecessary work. What would be wrong with this?:

```
sortedWinnerIndices = numpy.argsort(-overlaps, kind='mergesort')[:numActive]
```

chetan51: @SaganBolliger Thanks for the suggestion. I tested it, and it produces different results from the version I used. This makes it incompatible with the C++ spatial pooler.

scottpurdy:
There are some issues with Linux still but OS X seems to be in sync between nupic and nupic.core but fails on compatabiltiy test:

https://travis-ci.org/numenta/nupic/jobs/95645721

from this PR:

https://github.com/numenta/nupic/pull/2798

So you could start from that PR and get it working but won't be able to merge until we resolve the issues with linux and get them back in sync.

scottpurdy:
It looks like the compatibility test is the only thing standing in the way now. See #2801

chetan51:
@scottpurdy Passing! :fireworks:

rcrowder:
Nice work!

src/nupic/research/spatial_pooler.py
```diff
@@ -325,8 +325,9 @@ def __init__(self,
 
     # Initialize a tiny random tie breaker. This is used to determine winning
     # columns where the overlaps are identical.
-    self._tieBreaker = 0.01*numpy.array([self._random.getReal64() for i in
-                                        xrange(self._numColumns)])
+    self._tieBreaker = numpy.array([0.01 * self._random.getReal64() for i in
+                                      xrange(self._numColumns)],
```
scottpurdy: bad line wrap

src/nupic/research/spatial_pooler.py
```diff
@@ -1424,16 +1422,10 @@ def _inhibitColumnsGlobal(self, overlaps, density):
...
-    winnerIndices = numpy.argpartition(-overlaps, numActive)[:numActive]
-
-    # Compatibility with nupic.core requires that the winners are sorted;
-    # however, sorting only the winners is far less expensive than sorting all
-    # columns.
-    winnerValues = overlaps[winnerIndices]
-    sortedWinnerIndices = winnerIndices[numpy.argsort(-winnerValues)]
+    # Calculate winners using stable sort algorithm (mergesort)
+    # for compatibility with C++
+    winnerIndices = numpy.argsort(overlaps, kind='mergesort')
```
scottpurdy: double quotes

tests/unit/nupic/research/spatial_pooler_compatability_test.py
```diff
@@ -389,6 +389,27 @@ def testSerializationRun(self):
     self.runSerialize("cpp", params)
 
 
+  def testInhibitColumnsGlobal(self):
+    params = {
+      "inputDimensions": [512],
+      "columnDimensions": [512],
+      "globalInhibition": True,
+      "numActiveColumnsPerInhArea": 40,
+      "seed": 19
```
scottpurdy: nit: usually good to put trailing comma even for last element. Keeps diff clean when modifying sequences.

cogmission:
@rhyolight @mirkoklukas According to [this diff](https://github.com/numenta/nupic/pull/2755/files#r47160116) it doesn't appear to have been removed entirely, just changed?

subutai:
@mirkoklukas The tie breaker induced a fixed random ordering over the columns - in case of identical overlap scores, it would be used to decide which columns won. However this ordering was fixed throughout the lifetime of the SP.  We determined that there was already enough randomness due to random initialization, that a simple stable sort was sufficient (i.e. we just prefer lower column indices). Tests showed no difference in performance or the various properties we wanted.

In both cases, the tie breaking is fixed during the lifetime of the SP so that (at a given point in learning) a given input always returns the same output. 

The member variable self._tieBreaker is still there due to compatibility reasons, but it is not used. #2774 is there to remove it.

mirkoklukas:
@subutai Thanks a lot for clarifying that! Didn't realize that the tie
breaker was fixed the whole time.

2016-02-22 18:19 GMT+01:00 Subutai Ahmad <notifications@github.com>:

subutai:
@mirkoklukas No problem. It was a nice observation - glad you are looking into the implementation details!

PR #2760, opened 2015-11-18
Use resource handler deeper into expGen tests

### LINKED ISSUE #2735: Integration tests failing on Windows

rcrowder:
Fixes #2735 This sorts out the last failing integration tests on Windows. The four expGenerator tests that load and run the generated description.py and permutations.py for each test. Relying on the use of the resource handler when the HOTGYM_INPUT is not an absolute path. And no requirement to modify anything other than the look up of this data file.


rcrowder:
@scottpurdy Ready for review

PR #2779, opened 2015-11-30
Skipping unit tests that rely on pycapnp (win32 only)

### LINKED ISSUE #2778: Skip unit tests that rely on pycapnp package, when impl is missing
Certain platforms do not have an implementation of pycapnp. This leads to failures in unit testing that rely on the capnp module import. Mark these tests with skipif that can temporarily overcome this issue.
https://pytest.org/latest/skipping.html

rcrowder:
Fixes #2778 Relies on https://github.com/numenta/nupic.core/pull/705

Local output of `py.test tests\unit 2>&1 > .\unit_tests.log` can be found here; https://gist.github.com/rcrowder/b4702019cbda3e323968

### DIFF
--- a/tests/unit/nupic/engine/network_test.py
+++ b/tests/unit/nupic/engine/network_test.py
@@ -20,6 +20,8 @@
 # http://numenta.org/licenses/
 # ----------------------------------------------------------------------
 
+import sys
+import pytest
 from mock import Mock
 from mock import patch
 import unittest2 as unittest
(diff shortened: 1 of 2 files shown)


oxtopus:
I like this approach, but we should be using the native `unittest.skip*` functions rather than the ones provided by `pytest`.

See https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures for additional details.

rcrowder:
Ah, thanks for that link and advice. I'll try that alternative way out now.

oxtopus:
Thanks!

Issue #2780, opened 2015-11-30
Implement accessor routines for network


chetan51:
Implement accessor routines to get regions and their underlying objects by type (sensor, SP, TM, TP, KNN, etc.).

PR #2781, opened 2015-11-30
Implement accessor routines for network

### LINKED ISSUE #2784: For a particular region, get inputs and outputs of last compute
For debugging purposes.

chetan51:
Fixes https://github.com/numenta/nupic/issues/2780.
Fixes #2784.

### DIFF
--- a/examples/network/network_api_demo.py
+++ b/examples/network/network_api_demo.py
@@ -203,15 +203,15 @@ if __name__ == "__main__":
   network = createNetwork(dataSource)
   network.initialize()
 
-  spRegion = network.getRegion(SPRegion)
-  sp = spRegion.getSelf().getAlgorithm()
+  spRegion = network.getRegionsByType(SPRegion)[0]
+  sp = spRegion.getSelf().getAlgorithmInstance()
   print "spatial pooler region inputs: {0}".format(spRegion.getInputNames())
   print "spatial pooler region outputs: {0}".format(spRegion.getOutputNames())
   print "# spatial pooler columns: {0}".format(sp.getNumColumns())
   print
 
-  tmRegion = network.getRegion(TPRegion)
-  tm = tmRegion.getSelf().getAlgorithm()
+  tmRegion = network.getRegionsByType(TPRegion)[0]
+  tm = tmRegion.getSelf().getAlgorithmInstance()
   print "temporal memory region inputs: {0}".format(tmRegion.getInputNames())
   print "temporal memory region outputs: {0}".format(tmRegion.getOutputNames())
   print "# temporal memory columns: {0}".format(tm.numberOfCols)
--- a/examples/network/network_api_demo.py
+++ b/examples/network/network_api_demo.py
@@ -203,10 +203,19 @@ if __name__ == "__main__":
   network = createNetwork(dataSource)
   network.initialize()
 
-  sp = network.getRegion(SPRegion).getAlgorithm()
+  spRegion = network.getRegion(SPRegion)
+  sp = spRegion.getSelf().getAlgorithm()
+  print "spatial pooler region inputs: {0}".format(spRegion.getInputNames())
+  print "spatial pooler region outputs: {0}".format(spRegion.getOutputNames())
   print "# spatial pooler columns: {0}".format(sp.getNumColumns())
-  tm = network.getRegion(TPRegion).getAlgorithm()
+  print
+
+  tmRegion = network.getRegion(TPRegion)
+  tm = tmRegion.getSelf().getAlgorithm()
+  print "temporal memory region inputs: {0}".format(tmRegion.getInputNames())
+  print "temporal memory region outputs: {0}".format(tmRegion.getOutputNames())
   print "# temporal memory columns: {0}".format(tm.numberOfCols)
+  print
 
   outputPath = os.path.join(os.path.dirname(__file__), _OUTPUT_PATH)
   with open(outputPath, "w") as outputFile:
--- a/examples/network/network_api_demo.py
+++ b/examples/network/network_api_demo.py
@@ -31,6 +31,8 @@ from nupic.algorithms.anomaly import computeRawAnomalyScore
 from nupic.data.file_record_stream import FileRecordStream
 from nupic.engine import Network
 from nupic.encoders import MultiEncoder, ScalarEncoder, DateEncoder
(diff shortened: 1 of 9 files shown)


rhyolight:
By analyzing the blame information on this pull request, we identified @scottpurdy, @oxtopus and @vsinha to be potential reviewers

src/nupic/engine/__init__.py
```diff
@@ -721,10 +721,19 @@ def save(self, *args, **kwargs):
       raise TypeError("Save path must be of type {}.".format(str))
     engine.Network.save(self, *args, **kwargs)
 
-  def inspect(self):
-    """Launch a GUI inpector to inspect the network"""
-    from nupic.analysis import inspect
```
chetan51: This no longer exists, so I removed it.

chetan51:
@scottpurdy Please review.

examples/network/network_api_demo.py
```diff
@@ -199,6 +201,13 @@ def runNetwork(network, writer):
   dataSource = FileRecordStream(streamID=_INPUT_FILE_PATH)
 
   network = createNetwork(dataSource)
+  network.initialize()
+
+  sp = network.getRegion(SPRegion).getAlgorithm()
+  print "# spatial pooler columns: {0}".format(sp.getNumColumns())
```
scottpurdy: Is this useful for the demo or just for your debugging?

src/nupic/engine/__init__.py
```diff
@@ -721,10 +721,19 @@ def save(self, *args, **kwargs):
...
-    inspect(self)
+  def getRegion(self, regionClass):
```
scottpurdy: There can be multiple regions of the same type in a network so this should return a sequence (empty if none match). Rename to `getRegionsByType`?

src/nupic/regions/PyRegion.py
```diff
@@ -177,6 +177,15 @@ def getOutputElementCount(self, name):
     name: the name of the output
     """
 
+  def getAlgorithm(self):
```
scottpurdy: I don't think this makes sense to add. The return value for a region of unknown type is an unknown object. It would only make sense if we required all regions to have an "algorithm" object that subclasses the some common base class.

tests/unit/nupic/engine/network_test.py
```diff
@@ -408,6 +410,15 @@ def testTwoRegionPyNodeNetwork(self):
     self.assertEqual(r2dims[1], 2)
 
 
+  def testGetRegion(self):
+    n = engine.Network()
+
+    region1 = n.addRegion("region1", "py.TestNode", "")
+
+    self.assertEqual(n.getRegion(TestNode), region1.getSelf())
```
scottpurdy: This should be `self.assertIs` since we want to make sure that it is the same region instance, not one that has the same values. Also, do we want it to match `region1.getSelf()` or `region1`?

scottpurdy:
Done reviewing, not ready for merge.

src/nupic/regions/PyRegion.py
chetan51: This is for debugging purposes anyway, and when debugging people oftentimes try accessing `._sfdr` and `_tfdr`, so the need is there. How else can we add this functionality cleanly?

chetan51:
@scottpurdy Incorporated some of your feedback, responded to others. Also pulled in https://github.com/numenta/nupic/pull/2785 since those changes we intimately tied to these, so please review the additional changes here.

src/nupic/regions/PyRegion.py
subutai: I think this is one of those Pythonish-hacky things that is perhaps not clean but would be extremely useful in practice. This request seems to come up all the time.

src/nupic/regions/PyRegion.py
```diff
@@ -177,6 +177,15 @@ def getOutputElementCount(self, name):
...
+    """
+    Returns the underlying algorithm that is performing the computation.
+    This method should be overridden by the region subclass.
+
+    Returns None if there is no underlying algorithm for the region.
+    """
+    return None
```
subutai: I think it should throw an exception here rather than silently returning None.

src/nupic/regions/PyRegion.py
```diff
@@ -177,6 +177,15 @@ def getOutputElementCount(self, name):
     name: the name of the output
     """
 
+  def getAlgorithm(self):
+    """
+    Returns the underlying algorithm that is performing the computation.
+    This method should be overridden by the region subclass.
```
subutai: Per Scott's note we should be a bit more clear in the comments that the return type cannot be determined in advance and depends on the Region.

src/nupic/regions/PyRegion.py
subutai: Should we rename it to be a bit more precise, such as getAlgorithmInstance or getAlgorithmObject ?

src/nupic/regions/SPRegion.py
```diff
@@ -756,6 +756,10 @@ def getSpec(cls):
     return spec
 
 
+  def getAlgorithm(self):
```
subutai: I think we should have one for the CLA Classifier, and KNNClassifier as well.  And how about RecordSensor?

chetan51:
@subutai Incorporated your feedback.

src/nupic/regions/PyRegion.py
```diff
@@ -177,6 +177,18 @@ def getOutputElementCount(self, name):
     name: the name of the output
     """
 
+  @not_implemented
+  def getAlgorithmInstance(self):
```
scottpurdy: I don't think this should be added to PyRegion. Adding it here requires that every region have an "algorithm" which may not always make sense and the return value type will be different in every case. But fine to use the same name in the regions where we choose to implement it as a convention if that makes the API more discoverable.

scottpurdy:
done reviewing

src/nupic/regions/PyRegion.py
subutai: I'm fine with this suggestion. The rest of the code samples will still work as-is.  If we do this we should add a doc string to the implementation for each individual Region.

subutai:
:+1: other than handling Scott's comment.

chetan51:
Done. Will merge after tests pass.

PR #2789, opened 2015-12-03
Fix up RDSE serialization test.

### LINKED ISSUE #2788: RDSE serialization test could be incorrect
It doesn't call encode the same number of times on original and deserialized encoder.

scottpurdy:
fixes #2788 

@oxtopus please review


oxtopus:
:+1:

Issue #2830, opened 2015-12-15
Fix serialization of MultiEncoder


chetan51:
When using AdaptiveScalarEncoder, it crashes.

PR #2833, opened 2015-12-15
Fix serialization of MovingAverage

chetan51:
Fixes https://github.com/numenta/nupic/issues/2830.

### DIFF
--- a/src/nupic/movingaverage.capnp
+++ b/src/nupic/movingaverage.capnp
@@ -4,5 +4,5 @@
 struct MovingAverageProto {
   windowSize @0 :UInt32;
   slidingWindow @1 :List(Float32);
-  total @2 :UInt32;
+  total @2 :Float32;
 }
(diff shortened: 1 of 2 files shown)


src/nupic/movingaverage.capnp
```diff
@@ -4,5 +4,5 @@
 struct MovingAverageProto {
   windowSize @0 :UInt32;
   slidingWindow @1 :List(Float32);
-  total @2 :UInt32;
+  total @2 :Float32;
```
chetan51: Do we need to make this a different field? Or do we not care about backwards-compatibility at this point?

chetan51:
@scottpurdy

src/nupic/movingaverage.capnp
scottpurdy: Fine to reuse now but not much longer.

PR #2839, opened 2015-12-15
Add capnp serialization to SPRegion

chetan51:
Fixes https://github.com/numenta/nupic/issues/2838.

### DIFF
--- a/src/nupic/regions/SPRegion.capnp
+++ b/src/nupic/regions/SPRegion.capnp
@@ -1,0 +1,14 @@
+@0xf4a1336f535a66ba;
+
+using import "/nupic/proto/SpatialPoolerProto.capnp".SpatialPoolerProto;
+
+# Next ID: 7
+struct SPRegionProto {
+  spatialPooler @0 :SpatialPoolerProto;
+  columnCount @1 :UInt32;
+  inputWidth @2 :UInt32;
+  learningMode @3 :UInt32;
+  inferenceMode @4 :UInt32;
+  anomalyMode @5 :UInt32;
+  topDownMode @6 :UInt32;
+}
--- a/src/nupic/regions/SPRegion.capnp
+++ b/src/nupic/regions/SPRegion.capnp
@@ -2,13 +2,14 @@
 
 using import "/nupic/proto/SpatialPoolerProto.capnp".SpatialPoolerProto;
 
-# Next ID: 7
+# Next ID: 8
 struct SPRegionProto {
-  spatialPooler @0 :SpatialPoolerProto;
-  columnCount @1 :UInt32;
-  inputWidth @2 :UInt32;
-  learningMode @3 :UInt32;
-  inferenceMode @4 :UInt32;
-  anomalyMode @5 :UInt32;
-  topDownMode @6 :UInt32;
+  spatialImp @0 :Text;
+  spatialPooler @1 :SpatialPoolerProto;
+  columnCount @2 :UInt32;
+  inputWidth @3 :UInt32;
+  learningMode @4 :UInt32;
+  inferenceMode @5 :UInt32;
+  anomalyMode @6 :UInt32;
+  topDownMode @7 :UInt32;
 }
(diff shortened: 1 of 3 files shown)


chetan51:
@scottpurdy

src/nupic/regions/SPRegion.capnp
```diff
@@ -0,0 +1,14 @@
+@0xf4a1336f535a66ba;
+
+using import "/nupic/proto/SpatialPoolerProto.capnp".SpatialPoolerProto;
+
+# Next ID: 7
+struct SPRegionProto {
+  spatialPooler @0 :SpatialPoolerProto;
+  columnCount @1 :UInt32;
+  inputWidth @2 :UInt32;
+  learningMode @3 :UInt32;
```
scottpurdy: Each of the modes will only ever have a few possible values, right? Would `UInt8` work (0-255)?

src/nupic/regions/SPRegion.py
```diff
@@ -844,6 +850,39 @@ def setParameter(self, parameterName, index, parameterValue):
...
+    regionImpl.columnCount = self.columnCount
+    regionImpl.inputWidth = self.inputWidth
+    regionImpl.learningMode = 1 if self.learningMode else 0
+    regionImpl.inferenceMode = 1 if self.inferenceMode else 0
+    regionImpl.anomalyMode = 1 if self.anomalyMode else 0
+    regionImpl.topDownMode = 1 if self.topDownMode else 0
+
+
+  @classmethod
+  def read(cls, proto):
```
scottpurdy: This needs to do `self._sfdr.read(...)` at some point, right?

scottpurdy:
It would be nice to have the PyRegion Python base class to extract the "regionImpl" so every implementation doesn't have to. Find to do that refactoring separately.

src/nupic/regions/SPRegion.capnp
chetan51: Yes, but the datatype in the region currently is `UInt32`, so I thought I should be consistent.

scottpurdy: Yeah that makes sense. Begs the questions why it is so big in the region though...

chetan51:
@scottpurdy Updated, please review latest changes.

src/nupic/regions/SPRegion.py
```diff
@@ -844,6 +851,54 @@ def setParameter(self, parameterName, index, parameterValue):
...
+    instance.inferenceMode = regionImpl.inferenceMode
+    instance.anomalyMode = regionImpl.anomalyMode
+    instance.topDownMode = regionImpl.topDownMode
+
+    spatialImp = regionImpl.spatialImp
+
+    if spatialImp == 'py':
+      instance._sfdr = PYSpatialPooler.read(regionImpl.spatialPooler)
+    elif spatialImp == 'cpp':
+      instance._sfdr = CPPSpatialPooler()
```
scottpurdy: We should make a constructor for the C++ implementation that takes the proto instance rather than using the empty constructor and then `read`.

src/nupic/research/spatial_pooler.py
```diff
@@ -1663,7 +1663,7 @@ def write(self, proto):
     proto.potentialRadius = self._potentialRadius
     proto.potentialPct = self._potentialPct
     proto.inhibitionRadius = self._inhibitionRadius
-    proto.globalInhibition = self._globalInhibition
+    proto.globalInhibition = self._globalInhibition == True
```
scottpurdy: What values can `self._globalInhibition` have other than `True` and `False`?

chetan51: It seems to also take on the value of `1`, which is why I added this conversion (otherwise capnproto crashes).

scottpurdy: Perhaps this would be cleaner:

```
proto.globalInhibition = bool(self._globalInhibition)
```

chetan51:
@scottpurdy Can I merge this?

src/nupic/regions/SPRegion.py
chetan51: https://github.com/numenta/nupic.core/issues/793

scottpurdy:
:+1:

Issue #2840, opened 2015-12-15
hello_sp.py using C++ SP isn't producing correct results


chetan51:
In the "identical input" test, using C++ SP, different SDRs are produced from the SP.

PR #2841, opened 2015-12-15
Fix hello_sp.py identical input test, and fix random seeds

chetan51:
Fixes https://github.com/numenta/nupic/issues/2840.

### DIFF
--- a/examples/sp/hello_sp.py
+++ b/examples/sp/hello_sp.py
@@ -22,8 +22,12 @@
 """A simple program that demonstrates the working of the spatial pooler"""
 
 import numpy as np
-from random import randrange, random
+import random
 from nupic.bindings.algorithms import SpatialPooler as SP
+
+
+
+uintType = "uint32"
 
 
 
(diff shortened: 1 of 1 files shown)


subutai:
:+1: Thanks!

scottpurdy:
Can we make the default the fixed seed `0`? This was discussed in numenta/nupic.core#649

cogmission:
Can I ask how setting a different seed makes the demo not fail? Just trying to understand if I need to change something in HTM.java's SP example?

chetan51:
@scottpurdy @cogmission Actually, the issue wasn't the seed, but rather the datatype of the input numpy arrays. I set the seed just to make this example more deterministic, but it isn't necessary to fix this issue.

subutai:
> Can we make the default in the SP the fixed seed 0?

@scottpurdy Yes, we should do that (but in a separate PR). In CPP it is currently 1 for the SP, and in python it is currently -1.

PR #2885, opened 2015-12-21
RES-142 Add prototypeScores output to KNN region

SaganBolliger:
https://jira.numenta.com/browse/RES-142

WIP


SaganBolliger:
JIRA was updated as no longer needed.

Issue #2886, opened 2015-12-22
Refactor regions to use read class methods in SWIG bindings


chetan51:
Counterpart to https://github.com/numenta/nupic.core/issues/793.

PR #2887, opened 2015-12-22
Refactor SPRegion and CLAClassifierRegion to use read class methods in SWIG bindings

chetan51:
Fixes https://github.com/numenta/nupic/issues/2886.

Depends on https://github.com/numenta/nupic.core/pull/795.


chetan51:
@scottpurdy

Issue #2889, opened 2015-12-22
Add capnproto serialization to CLAModel


chetan51:

chetan51:
Currently using these hacky changes to test serialization of Network through CLAModel: https://github.com/numenta/nupic/compare/master...chetan51:network_serialization?expand=1

PR #2908, opened 2016-01-05
Add capnproto serialization to CLAModel

chetan51:
Fixes https://github.com/numenta/nupic/issues/2889.

### DIFF
--- a/src/nupic/frameworks/opf/CLAModelProto.capnp
+++ b/src/nupic/frameworks/opf/CLAModelProto.capnp
@@ -1,0 +1,20 @@
+@0xabf2792aa23483be;
+
+using import "/nupic/proto/NetworkProto.capnp".NetworkProto;
+
+# Next ID: 2
+struct CLAModelProto {
+  inferenceType @0 :InferenceType;
+  numRunCalls @1 :UInt32;
+  network @2 :NetworkProto;
+
+  enum InferenceType {
+    temporalNextStep @0;
+    temporalClassification @1;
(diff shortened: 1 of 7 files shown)


chetan51:
@scottpurdy

scottpurdy:
Please add basic tests to make sure it works and let me know, I'll review then.

chetan51:
@scottpurdy So the way I've been testing is by modifying OPF to call `writeToCheckpoint` instead of `save` and `readFromCheckpoint` instead of `load`, and running the `test_NonTemporalMultiStep` in `opf_checkpoint_test.py`. Is there any way I can check this in without changing the default serialization method for users of OPF?

scottpurdy:
You can test serialization at the CLAModel level or at the OPF level. It sounds like testing at the OPF level doesn't give you and option for which serialization functions to use so you might have to do it at the CLAModel level for now. Either way, if there is an existing test that works then make an additional test that does the same thing except calls the new serialization rather than old.

breznak:
I remember a ML discussion on too large serialized files, can you verify with the new method? Or leave it as a seperate issue?

scottpurdy:
@breznak - There was a question about the serialized state (old method) creating large files. I think the conclusion was that it was expected. The new method will use less space.

chetan51:
@scottpurdy Please see latest commit for the added tests.

scottpurdy:
:+1: after fixing AppVeyor by disabling new test for Windows.

Issue #2913, opened 2016-01-06
Add capnproto serialization to TMRegion


chetan51:
Depends on https://github.com/numenta/nupic.core/issues/816.

vitaly-krugl:
Closing as duplicate of #2183

PR #2928, opened 2016-01-08
Serialization migration utility

### LINKED ISSUE #2917: Create serialization migration utility
That converts old serialized model into new capnproto serialized model.

scottpurdy:
fixes #2917 

@chetan51 please review

### DIFF
--- a/src/nupic/frameworks/opf/checkpoint_migration.py
+++ b/src/nupic/frameworks/opf/checkpoint_migration.py
@@ -31,7 +31,10 @@ from nupic.frameworks.opf.modelfactory import ModelFactory
 
 
 def migrateModel(checkpointDir):
-  """Migrates a single model."""
+  """Migrates a single model.
+
+  :param checkpointDir: String containing the path to the checkpoint directory.
+  """
   model = ModelFactory.loadFromCheckpoint(checkpointDir)
 
   tempDir = tempfile.mkdtemp()
@@ -48,6 +51,11 @@ def migrateModel(checkpointDir):
 
 
 def migrateMultipleModels(checkpointDirs):
+  """Migrate multiple model checkpoints.
+
+  :param checkpointDirs: A sequence of strings that contain the checkpoint
+      directories.
+  """
   for checkpointDir in checkpointDirs:
     migrateModel(checkpointDir)
 
--- a/src/nupic/frameworks/opf/checkpoint_migration.py
+++ b/src/nupic/frameworks/opf/checkpoint_migration.py
@@ -26,13 +26,13 @@ import os
 import shutil
 import tempfile
 
-from nupic.frameworks.opf import modelfactory
+from nupic.frameworks.opf.modelfactory import ModelFactory
 
 
 
 def migrateModel(checkpointDir):
   """Migrates a single model."""
-  model = modelfactory.ModelFactory.loadFromCheckpoint(checkpointDir)
+  model = ModelFactory.loadFromCheckpoint(checkpointDir)
 
   tempDir = tempfile.mkdtemp()
   try:
--- a/src/nupic/frameworks/opf/checkpoint_migration.py
+++ b/src/nupic/frameworks/opf/checkpoint_migration.py
@@ -1,0 +1,65 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2016, Numenta, Inc.  Unless you have an agreement
+# with Numenta, Inc., for a separate license for this software code, the
+# following terms and conditions apply:
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero Public License for more details.
+#
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+"""Provides utility for migrating saved models to the latest version."""
+
+import argparse
+import os
+import shutil
+import tempfile
+
+from nupic.frameworks.opf import modelfactory
+
+
+
+def migrateModel(checkpointDir):
+  """Migrates a single model."""
+  model = modelfactory.ModelFactory.loadFromCheckpoint(checkpointDir)
+
+  tempDir = tempfile.mkdtemp()
+  try:
+    # Write new checkpoint to temporary location.
+    tempCheckpointDir = os.path.join(tempDir, "tmp")
+    model.writeToCheckpoint(tempCheckpointDir)
+    # Delete old checkpoint and move new one to replace it.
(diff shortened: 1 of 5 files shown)


scottpurdy:
@chetan51 - I currently get this error when I run the utility:

```
$ nupic-migrate-checkpoints examples/opf/experiments/multistep/hotgym/savedmodels/*
  File "/Users/spurdy/Library/Python/2.7/bin/nupic-migrate-checkpoints", line 9, in <module>
    load_entry_point('nupic==0.3.7.dev0', 'console_scripts', 'nupic-migrate-checkpoints')()
  File "/Users/spurdy/nta/nupic/src/nupic/frameworks/opf/checkpoint_migration.py", line 65, in migrateCheckpointsMain
    migrateMultipleModels(args.checkpointDirs)
  File "/Users/spurdy/nta/nupic/src/nupic/frameworks/opf/checkpoint_migration.py", line 52, in migrateMultipleModels
    migrateModel(checkpointDir)
  File "/Users/spurdy/nta/nupic/src/nupic/frameworks/opf/checkpoint_migration.py", line 41, in migrateModel
    model.writeToCheckpoint(tempCheckpointDir)
  File "/Users/spurdy/nta/nupic/src/nupic/frameworks/opf/model.py", line 204, in writeToCheckpoint
    self.write(proto)
  File "/Users/spurdy/nta/nupic/src/nupic/frameworks/opf/clamodel.py", line 1312, in write
    self._netInfo.net.write(proto.network)
  File "/Users/spurdy/nta/nupic.core/bindings/py/nupic/bindings/engine_internal.py", line 1203, in write
    return _engine_internal.Network_write(self, *args)
  File "/Users/spurdy/nta/nupic.core/bindings/py/nupic/bindings/regions/PyRegion.py", line 300, in write
    regionImpl = proto.regionImpl.as_struct(self.getProtoType())
  File "/Users/spurdy/nta/nupic.core/bindings/py/nupic/bindings/regions/PyRegion.py", line 291, in getProtoType
    raise NotImplementedError()
NotImplementedError
```

chetan51:
@scottpurdy What are you running the utility on? If the network has a `TPRegion` in it, you will see that error because serialization isn't yet implemented for `TPRegion` (see https://github.com/numenta/nupic/issues/2913).

src/nupic/frameworks/opf/checkpoint_migration.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+
+def migrateMultipleModels(checkpointDirs):
+  for checkpointDir in checkpointDirs:
+    migrateModel(checkpointDir)
+
+
+
+def migrateCheckpointsMain():
```
chetan51: Can this just be `__main__` so that the script is runnable directly?

src/nupic/frameworks/opf/checkpoint_migration.py
```diff
@@ -0,0 +1,65 @@
...
+# ----------------------------------------------------------------------
+
+"""Provides utility for migrating saved models to the latest version."""
+
+import argparse
+import os
+import shutil
+import tempfile
+
+from nupic.frameworks.opf import modelfactory
```
chetan51: I would use `from nupic.frameworks.opf.modelfactory import ModelFactory`.

src/nupic/frameworks/opf/checkpoint_migration.py
```diff
@@ -0,0 +1,65 @@
...
+  """Migrates a single model."""
+  model = modelfactory.ModelFactory.loadFromCheckpoint(checkpointDir)
+
+  tempDir = tempfile.mkdtemp()
+  try:
+    # Write new checkpoint to temporary location.
+    tempCheckpointDir = os.path.join(tempDir, "tmp")
+    model.writeToCheckpoint(tempCheckpointDir)
+    # Delete old checkpoint and move new one to replace it.
+    shutil.rmtree(checkpointDir)
```
chetan51: Should we move this to a `backup` directory instead?

src/nupic/frameworks/opf/modelfactory.py
```diff
@@ -80,13 +81,17 @@ def create(modelConfig, logLevel=logging.ERROR):
...
+  def loadFromCheckpoint(savedModelDir):
+    """Load saved model.
+
     @param savedModelDir (string)
-           Directory of where the experiment is to be or was saved
+        Directory of where the experiment is to be or was saved
     @returns (nupic.frameworks.opf.model.Model) The loaded model instance.
     """
-    if newSerialization:
+    if os.path.exists(os.path.join(savedModelDir, "model.data")):
```
chetan51: Instead, you can use `Model._getModelCheckpointFilePath`.

chetan51:
Done reviewing.

src/nupic/frameworks/opf/modelfactory.py
scottpurdy: I'd prefer not to use a private method.

src/nupic/frameworks/opf/checkpoint_migration.py
scottpurdy: Is there a use case for that? I had considered allowing an optional base directory for the new models, which would leave the originals alone, but wasn't sure when we would use it. You can also create a backup yourself before running the script.

src/nupic/frameworks/opf/checkpoint_migration.py
scottpurdy: Sure. Usually I always import the model rather than things inside it. This usually gives more explicit code but in this case the module and class names are redundant.

src/nupic/frameworks/opf/checkpoint_migration.py
scottpurdy: The problem with that is that you end up running the script from the source location instead of install location which can cause issues. Austin, Vitaly, and I have been avoiding using the `__main__` block convention since it can lead to confusing errors.

src/nupic/frameworks/opf/checkpoint_migration.py
chetan51: Okay, no worries then.

src/nupic/frameworks/opf/checkpoint_migration.py
chetan51: Got it.

src/nupic/frameworks/opf/modelfactory.py
chetan51: Should we make it a public method?

scottpurdy: Yeah maybe. What do you think? And would we need two functions then for the two checkpoint formats?

chetan51: Yeah I think so. And we already have two functions for the two formats.

scottpurdy: Alright, made these changes

scottpurdy:
@chetan51 - any recommendation for tests? I could create a CLAModel, call save, use the tool to convert, and validate that it worked and pre- and post-serialization models are the same.

scottpurdy:
@chetan51 ping, just want to make sure you don't have another idea for testing this

chetan51:
@scottpurdy You can also add tests in `opf_checkpoint_test.py`, modifying it to optionally use the tool when checkpointing and loading (specifying which serialization type to use with the parameter `newSerialization` in ExperimentRunner).

scottpurdy:
I don't think I can easily test this until #2913 is complete so I will work on that now.

rhyolight:
Re open this PR if there are new updates.

PR #2937, opened 2016-01-12
SpatialPoolerEncoder

codedhard:
SpatialPooler as an Encoder.
Uses SDR list as input.


breznak:
Hi @codedhard , 
thanks for the patches! 
What is the motivation for this change? Could the same be achieved by PassThroughEncoder/directly inputting a vector to SP/OPF Network?

codedhard:
Pass_Through_Encoder.py
Different input (%i) and output (%i) sizes.

With an SP encoder you can have input bigger then output.
This would be useful when doing input from the TP/TM cells.

rhyolight:
@codedhard Thanks for your contribution! We always appreciate the help, but you might be more successful with your proposed changes if you read through our [development process](https://github.com/numenta/nupic/wiki/Development-Process). When we get PRs "out of the blue" it takes longer to understand them, vet them, and merge them. 

Adding a new encoder is a significant change, so we really should discuss this on [nupic-hackers](http://lists.numenta.org/mailman/listinfo/nupic-hackers_lists.numenta.org) before considering this PR. Are you willing to bring it up in a new thread there?

codedhard:
@rhyolight Now I had posted it to nupic-hackers Mailing list. My code would need some improvements, mainly on performance.

rhyolight:
Thanks very much. We will continue this discussion there.

---------
Matt Taylor
OS Community Flag-Bearer
Numenta

scottpurdy:
I'd agree with @breznak that pass through -> SPRegion would be a better way to do this.

rhyolight:
This is interesting, but based on the discussion above, I don't think we want to have an SP encoder in the NuPIC codebase. That doesn't mean it's not useful, but we don't have a reason for it. 

@codedhard The cool thing about creating encoders is that they are an easy extension point you can build up in your own codebases. One idea I've had in the past is to collect encoders the community has created in a repository in https://github.com/htm-community. If you want to go that route and store the SP Encoder there, let me know and I can set it up.

Issue #2935, opened 2016-01-11
AttributeError: 'TPShim' object has no attribute 'topDownCompute'


rhyolight:
I got this error by switching from `'temporalImp': 'cpp'` to `'temporalImp': 'tm_py'` in the model params of a model. The code worked before the change, but after the change I got this:

```
Traceback (most recent call last):
  File "./run.py", line 226, in <module>
    runModel(model, data, field, plot, options.log)
  File "./run.py", line 195, in runModel
    "value": value
  File "/Users/mtaylor/Library/Python/2.7/lib/python/site-packages/nupic/frameworks/opf/clamodel.py", line 386, in run
    self._tpCompute()
  File "/Users/mtaylor/Library/Python/2.7/lib/python/site-packages/nupic/frameworks/opf/clamodel.py", line 503, in _tpCompute
    tp.compute()
  File "/Users/mtaylor/nta/nupic/src/nupic/engine/__init__.py", line 446, in compute
    return self._region.compute()
  File "/Users/mtaylor/nta/nupic.core/bindings/py/nupic/bindings/engine_internal.py", line 1384, in compute
    return _engine_internal.Region_compute(self)
  File "/Users/mtaylor/nta/nupic/src/nupic/regions/TPRegion.py", line 480, in compute
    self._compute(inputs, outputs)
  File "/Users/mtaylor/nta/nupic/src/nupic/regions/TPRegion.py", line 534, in _compute
    outputs['topDownOut'][:] = self._tfdr.topDownCompute().copy()
AttributeError: 'TPShim' object has no attribute 'topDownCompute'
```

I am running tip of master of both core and nupic, compiled locally on my machine:

    nupic 0.3.7.dev0 (SHA 58c96d0f39b074e5c306c6863ae57345e47f2908)
    nupic.bindings 0.2.8.dev0 (SHA 621a1b7fb3d0ff4ff3d2358a782095350945c266)

This error is very easy to replicate in the `river-runner` demo app.

```
git clone git@github.com:nupic-community/river-runner.git
cd river-runner
```

Test that it works before switching temporal implementation:

    python run.py [--plot]

If not plotting, results are written to a CSV file.

Now edit [`model_params/anomaly_params.py`](https://github.com/nupic-community/river-runner/blob/master/model_params/anomaly_params.py#L86) and switch from `'temporalImp': 'cpp'` to `'temporalImp': 'tm_py'`.

Now when running, you should get the error pasted above.

rhyolight:
[...]

PR #2940, opened 2016-01-12
Fix AttributeError: 'TPShim' object has no attribute 'topDownCompute'

chetan51:
Fixes https://github.com/numenta/nupic/issues/2935.

### DIFF
--- a/src/nupic/research/TP_shim.py
+++ b/src/nupic/research/TP_shim.py
@@ -105,6 +105,21 @@ class TPShim(TemporalMemory):
     return output
 
 
+  def topDownCompute(self, topDownIn=None):
+    """
+    (From `TP.py`)
+    Top-down compute - generate expected input given output of the TP
+
+    @param topDownIn top down input from the level above us
+
+    @returns best estimate of the TP input that would have generated bottomUpOut.
+    """
+    output = numpy.zeros(self.numberOfColumns())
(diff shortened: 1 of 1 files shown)


chetan51:
@scottpurdy

src/nupic/research/TP_shim.py
```diff
@@ -105,6 +105,21 @@ def compute(self, bottomUpInput, enableLearn, computeInfOutput=None):
     return output
 
 
+  def topDownCompute(self, topDownIn=None):
```
scottpurdy: Could you create a mixin class that defines functions like this and add it as an additional base class for each shim? Just trying to think of a way to avoid so much duplicated code.

scottpurdy:
Done reviewing

src/nupic/research/TP_shim.py
chetan51: I tried but I couldn't get the inheritance working correctly when I did. I can try again, but we should probably work to switch to the new `TemporalMemoryRegion` that Jared created instead.

scottpurdy:
:+1:

chetan51:
@rhyolight Can I merge?

rhyolight:
@chetan51 I retriggered appveyor, let's wait and ensure it is green.

PR #2950, opened 2016-01-16
Remove leaveOneOutTest() in KNNClassifier

### LINKED ISSUE #2902: Remove leaveOneOutTest and doSelfValidation in KNN classes.
Remove leaveOneOutTest() in KNNClassifier.  It is a legacy function - only used by the Region for a doSelfValidation mode.  That mode is not used by anything in NuPIC.

This issue is to remove both. We should make sure serialization backward compatibility is handled correctly.

subutai:
Remove leaveOneOutTest() in KNNClassifier. It is a legacy function - only used by the Region for a doSelfValidation mode. That mode is not used by anything in NuPIC.

Fixes #2902


scottpurdy:
:+1:

subutai:
@scottpurdy Do you think I need to add a check in __setstate for doSelfValidation to handle backward compatibility? Currently I didn't - thinking it would just initialize an unused but harmless member variable.

scottpurdy:
@subutai - Yeah probably a good idea to just delete it from the dict if it exists.

PR #2952, opened 2016-01-18
Updated hotgym regression test

### LINKED ISSUE #2951: AppVeylor NuPIC test failure after Merge pull request #2950
NuPIC test failure https://ci.appveyor.com/project/numenta-ci/nupic/branch/master (after Merge pull request #2950 from subutai/2902)   
```
tests.integration.nupic.opf.hotgym_regression_test.HotgymRegressionTest testMethod=testHotgymRegression

        # Changes that affect prediction results will cause this test to fail.
        # If the change is understood and reviewers agree that there has not been a
        # regression then this value can be updated to reflect the new result.
>      self.assertAlmostEqual(float(lastRow[14]), 5.90725098207)
E      AssertionError: 5.87970891453 != 5.90725098207 within 7 places
```

subutai:
Fixes #2951 

Don't merge - dummy commit to test possible intermittent failure

### DIFF
--- a/tests/integration/nupic/opf/hotgym_regression_test.py
+++ b/tests/integration/nupic/opf/hotgym_regression_test.py
@@ -61,8 +61,8 @@ class HotgymRegressionTest(unittest.TestCase):
                          "steps=1:window=1000:field=consumption")
         lastRow = collections.deque(reader, 1)[0]
 
-      # Changes that affect prediction results will cause this test to fail.
-      # If the change is understood and reviewers agree that there has not been a
+      # Changes that affect prediction results will cause this test to fail. If
+      # the change is understood and reviewers agree that there has not been a
       # regression then this value can be updated to reflect the new result.
       self.assertAlmostEqual(float(lastRow[14]), 5.90725098207)
 
--- a/tests/integration/nupic/opf/hotgym_regression_test.py
+++ b/tests/integration/nupic/opf/hotgym_regression_test.py
@@ -63,7 +63,7 @@ class HotgymRegressionTest(unittest.TestCase):
       # Changes that affect prediction results will cause this test to fail. If
       # the change is understood and reviewers agree that there has not been a
       # regression then this value can be updated to reflect the new result.
-      self.assertLess(abs(float(lastRow[14])-5.90725098207), 0.04)
+      self.assertAlmostEqual(float(lastRow[14]), 5.87970891453)
 
     finally:
       shutil.rmtree(resultsDir, ignore_errors=True)
--- a/tests/integration/nupic/opf/hotgym_regression_test.py
+++ b/tests/integration/nupic/opf/hotgym_regression_test.py
@@ -28,7 +28,6 @@ explicitly updated to match the new results.
 import collections
 import csv
 import os
-import pkg_resources
 import shutil
 import unittest
 
@@ -64,7 +63,7 @@ class HotgymRegressionTest(unittest.TestCase):
       # Changes that affect prediction results will cause this test to fail. If
       # the change is understood and reviewers agree that there has not been a
       # regression then this value can be updated to reflect the new result.
-      self.assertAlmostEqual(float(lastRow[14]), 5.90725098207)
+      self.assertLess(abs(float(lastRow[14])-5.90725098207), 0.04)
 
     finally:
       shutil.rmtree(resultsDir, ignore_errors=True)
--- a/.nupic_modules
+++ b/.nupic_modules
@@ -1,3 +1,3 @@
 # Default nupic.core dependencies (override in optional .nupic_config)
 NUPIC_CORE_REMOTE = 'git://github.com/numenta/nupic.core.git'
-NUPIC_CORE_COMMITISH = '621a1b7fb3d0ff4ff3d2358a782095350945c266'
+NUPIC_CORE_COMMITISH = '7de0c5992364dc847dc2b8303093ff8aea52b61e'


subutai:
@vitaly-krugl Please review, thanks.  Either this test is somewhat non-deterministic or there is a tiny platform-specific difference in the error. I'm guessing it's probably the latter since this test never fails with non-Windows tests. Either way, I think this is the right fix for now.

rcrowder:
@subutai cc @scottpurdy 
This change may not be required. I've just been trying to get the current nupic.core (https://github.com/numenta/nupic.core/commit/7f24691b0d72fcf2adfc6faadf94cd7454edbeef) to work with different commits to nupic (all the way back to 0.2.8.dev0, nupic PR #2876, commit sha 01946897fcbc7c48ebe282e7f76d07c476413493). That integration test in hot gym regression would also show the differing value seen on AppVeyor.
It was only when I used the latest nupic master with nupic.core https://github.com/numenta/nupic.core/commit/621a1b7fb3d0ff4ff3d2358a782095350945c266 did the test pass as expected. My hunch is that it is related to this change https://github.com/numenta/nupic.core/commit/bdc7d890497af242e68b5344dff1e21ca799cc46

subutai:
@rcrowder Are you saying that the change won't be needed once NuPIC is caught up with the latest version of nupic.core and NUPIC_CORE_COMMITISH has the latest SHA of nupic.core?

vitaly-krugl:
@subutai, if this is the case, then I would rather punt on this PR and wait for NuPIC to catch up.

tests/integration/nupic/opf/hotgym_regression_test.py
```diff
@@ -61,10 +60,10 @@ def testHotgymRegression(self):
                          "steps=1:window=1000:field=consumption")
         lastRow = collections.deque(reader, 1)[0]
 
-      # Changes that affect prediction results will cause this test to fail.
-      # If the change is understood and reviewers agree that there has not been a
+      # Changes that affect prediction results will cause this test to fail. If
+      # the change is understood and reviewers agree that there has not been a
       # regression then this value can be updated to reflect the new result.
-      self.assertAlmostEqual(float(lastRow[14]), 5.90725098207)
+      self.assertLess(abs(float(lastRow[14])-5.90725098207), 0.04)
```
vitaly-krugl: The automatically-generated failure message might habe been more useful for debugging in the original assertion (would show the explicit values that differed), while the message generated automatically by the new statement would masquerade the actual values.

subutai:
> @subutai, if this is the case, then I would rather punt on this PR and wait for NuPIC to catch up with core.

@vitaly-krugl Agreed. Let's wait for @rcrowder's response.

rcrowder:
I awoke this morning to see the nupic update of the NUPIC_CORE_COMMITISH sha. As can now be seen on Travis [1] the hot gym regression integration test is failing in the same way as AppVeyor [2]. 
Extensive testing last night on Windows and Linux of recent changes in nupic (taking a stance that nupic.core is correct), always showed that nupic changes were not the issue. Changing stance, to assume that the latest nupic master is correct, and checking out previous nupic.core commits lead to the discovery of the recent segment synaptic management change [3].
I've not gone further than testing commit revisions in nupic and nupic.core. Numenta need to determine whether the _current_ algorithmic changes introduced to nupic.core synapse freeing/reuse from segments is appropriate [3a, 3b]. The intention is correct and necessary, the question relates to the current form of implementation in the Segment::freeNSynapses changes.
As Vitaly implies, the change in this PR would mask any subsequent changes made in nupic or nupic.core that would lead to tests failing in the future, such as this hot gym regression one failing now. So imo this PR is not required.

1 https://travis-ci.org/numenta/nupic/builds/103242418
2 https://ci.appveyor.com/project/numenta-ci/nupic/build/0.1.0.874
3 https://github.com/numenta/nupic.core/commit/bdc7d890497af242e68b5344dff1e21ca799cc46
3a https://github.com/numenta/nupic.core/commit/7f24691b0d72fcf2adfc6faadf94cd7454edbeef#diff-9c0bbdc23eb0bad8a7ad0457b57a43d1R376
3b https://github.com/numenta/nupic.core/commit/7f24691b0d72fcf2adfc6faadf94cd7454edbeef#diff-9c0bbdc23eb0bad8a7ad0457b57a43d1R387

subutai:
@rcrowder Thanks!  It seems like there were two problems:

1) Appveyor is not using NUPIC_CORE_COMMITISH properly. If the test failure was indeed tied to a nupic.core SHA that is more recent than NUPIC_CORE_COMMITISH, then the NuPIC PR should not fail. @rhyolight - who should look into this?

2) Small algorithm changes can cause small differences in the error.  This happens very often. If after review, the algorithm change itself is ok, then we should update the test error.  It is great the error is consistent across platforms, then we don't need the guard band I put in this PR. We can just update the number. The algorithm fixes in the PRs you pointed to look correct to me. I will review internally and (if all looks ok) I think we should update the expected error in the test, and we should be good to go.

Thanks for the very helpful investigation!

tests/integration/nupic/opf/hotgym_regression_test.py
scottpurdy: You can instead add the parameter `delta=0.04` to the `assertAlmostEqual` method.

scottpurdy:
:+1:

tests/integration/nupic/opf/hotgym_regression_test.py
```diff
@@ -61,10 +60,10 @@ def testHotgymRegression(self):
                          "steps=1:window=1000:field=consumption")
         lastRow = collections.deque(reader, 1)[0]
 
-      # Changes that affect prediction results will cause this test to fail.
-      # If the change is understood and reviewers agree that there has not been a
+      # Changes that affect prediction results will cause this test to fail. If
+      # the change is understood and reviewers agree that there has not been a
       # regression then this value can be updated to reflect the new result.
-      self.assertAlmostEqual(float(lastRow[14]), 5.90725098207)
+      self.assertAlmostEqual(float(lastRow[14]), 5.87970891453)
```
vitaly-krugl: @subutai, according to @rcrowder's comment https://github.com/numenta/nupic/pull/2952#issuecomment-172668787, the test should pass without this assertion change with a more recent nupic core.

vitaly-krugl:
LGTM

CC @subutai

rcrowder:
@subutai The handling of the core commit sha in the nupic and nupic.core AppVeyor scripts is something that does need looking at (before any Bamboo transition). Changes to the AV scripts to handle this are fairly trivial. A simple case of appropriate nupic.bindings wheel file renaming, AWS deployment change, with subsequent change on the nupic side to obtain a wheel based on the sha from .nupic_modules. Myself and/or Matt could do those changes. But bear in mind I won't be available until start of Feb to make actual commit changes. At the moment with that hot gym test value now changed, AV should continue to be stable for the time being.

subutai:
@rcrowder Thanks - I created #2965 to track this issue.

PR #2954, opened 2016-01-19
Python vs C++ TemporalMemory compatibility test

### LINKED ISSUE #2948: Create compatibility test for python vs C++ TemporalMemory
Similar to network_checkpoint_test.py.

SaganBolliger:
Fixes #2948

This currently fails, but @chetan51 mentioned that this is to be expected at this point.

I also want to verify that "py" and "tm_py" are the right two implementations to be comparing.


chetan51:
Can we combine this with the network_checkpoint_test.py into a new test that encompasses both? Because there is a lot of duplicate code between the two.

Ultimately, you will be comparing `tm_py` and `tm_cpp`, but we don't yet have `tm_cpp` working so this is fine for now.

SaganBolliger:
What if we separate out the common code into a third file, and import that? That way we wouldn't have to combine two conceptually different kinds of tests, checkpointing tests and compatibility tests, into one file.

chetan51:
@SaganBolliger Sounds good to me.

SaganBolliger:
@chetan51 Refactored as discussed.

SaganBolliger:
@chetan51 Interesting that the Travis builds are timing out and not even getting to the point where the test fails. I wonder if this means we should use a smaller dataset.

chetan51:
@SaganBolliger Are you sure it's this test that's causing Travis to time out?

SaganBolliger:
@chetan51 Actually, no I'm not sure.

tests/integration/nupic/engine/network_creation_common.py
```diff
@@ -0,0 +1,228 @@
...
+if capnp:
+  from nupic.proto import NetworkProto_capnp
+
+
+_VERBOSITY = 0  # how chatty the test should be
+_SEED = 1956  # the random seed used throughout
+_INPUT_FILE_PATH = resource_filename(
+  "nupic.datafiles", "extra/hotgym/rec-center-hourly.csv"
+)
+_NUM_RECORDS = 2000
```
chetan51: We can reduce this to 1000 to make the test run faster.

tests/integration/nupic/engine/temporal_memory_compatibility_test.py
```diff
@@ -0,0 +1,63 @@
...
+    Test compatibility between C++ and Python TM implementation.
+    """
+    results1 = createAndRunNetwork(TPRegion,
+                                   "bottomUpOut",
+                                   checkpointMidway=False,
+                                   temporalImp="tm_py")
+    results2 = createAndRunNetwork(TPRegion,
+                                   "bottomUpOut",
+                                   checkpointMidway=False,
+                                   temporalImp="py")
```
chetan51: This should be `tm_py` for now. We'll change it to `tm_cpp` once that is supported.

chetan51:
@rhyolight Can you restart the AppVeyor build?

SaganBolliger:
@chetan51 Looks like this is good to merge.

PR #2962, opened 2016-01-19
Add TM accessors, change compute, and update callers

### LINKED ISSUE #2959: TM: Add accessor methods and update API to use lists of indices
For `temporal_memory.py`:

- Add accessor methods for `activeCells`, `predictiveCells`, `activeSegments`, `winnerCells`, `matchingSegments`, `matchingCells`. These should return a list of indices. Update all callers to use these methods instead of accessing the properties directly.
- Change `compute` method to take a list of indices (instead of a set), and update all callers.

SaganBolliger:
Fixes #2959 

@chetan51 @subutai

### DIFF
--- a/src/nupic/research/TP_shim.py
+++ b/src/nupic/research/TP_shim.py
@@ -105,8 +105,7 @@ class TPShim(TemporalMemory):
     self.infActiveState["t"] = activeState
 
     output = numpy.zeros(numberOfCells)
-    output[list(set(self.getPredictiveCells()) | 
-                set(self.getActiveCells()))] = 1
+    output[self.getPredictiveCells() + self.getActiveCells()] = 1
     return output
 
 
@@ -213,8 +212,7 @@ class FastTPShim(FastTemporalMemory):
     self.infActiveState["t"] = activeState
 
     output = numpy.zeros(numberOfCells)
-    output[list(set(self.getPredictiveCells()) | 
-                set(self.getActiveCells()))] = 1
+    output[self.getPredictiveCells() + self.getActiveCells()] = 1
     return output
 
 
@@ -321,8 +319,7 @@ class MonitoredTPShim(MonitoredTemporalMemory):
     self.infActiveState["t"] = activeState
 
     output = numpy.zeros(numberOfCells)
-    output[list(set(self.getPredictiveCells()) | 
-                set(self.getActiveCells()))] = 1
+    output[self.getPredictiveCells() + self.getActiveCells()] = 1
     return output
 
 
@@ -429,8 +426,7 @@ class MonitoredFastTPShim(MonitoredFastTemporalMemory):
     self.infActiveState["t"] = activeState
 
     output = numpy.zeros(numberOfCells)
-    output[list(set(self.getPredictiveCells()) | 
-                set(self.getActiveCells()))] = 1
+    output[self.getPredictiveCells() + self.getActiveCells()] = 1
     return output
 
 
--- a/src/nupic/research/TP_shim.py
+++ b/src/nupic/research/TP_shim.py
@@ -101,11 +101,12 @@ class TPShim(TemporalMemory):
     numberOfCells = self.numberOfCells()
 
     activeState = numpy.zeros(numberOfCells)
-    activeState[self.getCellIndices(self.activeCells)] = 1
+    activeState[self.getActiveCells()] = 1
     self.infActiveState["t"] = activeState
 
     output = numpy.zeros(numberOfCells)
-    output[self.getCellIndices(self.predictiveCells | self.activeCells)] = 1
+    output[list(set(self.getPredictiveCells()) | 
+                set(self.getActiveCells()))] = 1
     return output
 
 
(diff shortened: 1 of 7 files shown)


chetan51:
Will review when WIP label is removed.

SaganBolliger:
@chetan51 What should we do [here](https://github.com/numenta/nupic/pull/2962/files#diff-b7deb21f9a3337fa03896dc6a396b5d1R110) given that we have no segment accessors?

chetan51:
@SaganBolliger Just print the # of segments instead.

SaganBolliger:
@chetan51 Ready for review.

src/nupic/research/TP_shim.py
```diff
@@ -101,11 +101,12 @@ def compute(self, bottomUpInput, enableLearn, computeInfOutput=None):
...
 
     activeState = numpy.zeros(numberOfCells)
-    activeState[self.getCellIndices(self.activeCells)] = 1
+    activeState[self.getActiveCells()] = 1
     self.infActiveState["t"] = activeState
 
     output = numpy.zeros(numberOfCells)
-    output[self.getCellIndices(self.predictiveCells | self.activeCells)] = 1
+    output[list(set(self.getPredictiveCells()) | 
+                set(self.getActiveCells()))] = 1
```
chetan51: It should be possible to just do:

    output[self.getPredictiveCells() + self.getActiveCells()] = 1

src/nupic/research/TP_shim.py
```diff
@@ -208,11 +209,12 @@ def compute(self, bottomUpInput, enableLearn, computeInfOutput=None):
...
```
chetan51: See https://github.com/numenta/nupic/pull/2962/files#r50485861.

src/nupic/research/TP_shim.py
```diff
@@ -315,11 +317,12 @@ def compute(self, bottomUpInput, enableLearn, computeInfOutput=None):
...
```
chetan51: See https://github.com/numenta/nupic/pull/2962/files#r50485861.

src/nupic/research/TP_shim.py
```diff
@@ -422,11 +425,12 @@ def compute(self, bottomUpInput, enableLearn, computeInfOutput=None):
...
```
chetan51: See https://github.com/numenta/nupic/pull/2962/files#r50485861.

src/nupic/research/fast_temporal_memory.py
```diff
@@ -175,7 +175,7 @@ def columnForCell(self, cell):
     """
     self._validateCell(cell)
 
-    return int(cell.idx / self.cellsPerColumn)
+    return int(cell / self.cellsPerColumn)
```
chetan51: Why was this change necessary?

chetan51:
Done with review.

src/nupic/research/fast_temporal_memory.py
SaganBolliger: @chetan51 Well, two reasons:

1. One the way the function was implemented before was at odds with the docstring. The docstring specifies an int parameter, but the implementation expected a Cell object. Since no code that I wasn't already modifying was using the old version, this seemed like an easy way to fix an inconsistency in the API.

2. Previously, code that accessed attributes like `activeCells` in `FastTemporalMemory` directly would get a collection of `Cell` objects, whereas code that accessed the same attribute in `TemporalMemory` would get a collection of indices, so it made sense that the former's implementation of `columnForCell` would accept `Cell`s whereas the latter's would accept indices. After changing the callers to use the accessors, we've standardized on returning indices for both implementations, and so this function can always expect indices.

Though, actually, now that I think of it, since `FastTemporalMemory` inherits from `TemporalMemory` we can delete this definition since two are now identical.

SaganBolliger: I suppose though the question is whether this function is part of the API or is only meant to be used internally. It is used externally in various places, but if it shouldn't be, we can create a new accessor for that purpose and leave this one as is (updating the docstring to reflect the actual functionality).

SaganBolliger:
@chetan51 I've simplified the bitmask creation code as you suggested. I also deleted the definition of `columnForCell` in `FastTemporalMemory`, but see my [comments](https://github.com/numenta/nupic/pull/2962#discussion_r50574790) on the matter and let me know if you think this was the right thing to do.

chetan51:
@SaganBolliger That makes sense to me. A client of `fast_temporal_memory.py` would only be able to access indices (rather that `Cell` objects) by using the public accessor methods, right? Which means we can remove overrides in `fast_temporal_memory.py` for `columnForCell` (which you've already done), `cellsForColumn`, and `getCellIndex` in `fast_temporal_memory` and `temporal_memory.py`.

SaganBolliger:
@chetan51 The one complication is that `cellsForColumn` is used internally by `fast_temporal_memory.py`, where it requires `Cell` objects. So there's a bit of vagueness in terms of what's an internal helper method and what's part of the external API. I'm treating `getCellIndex` as an internal helper method that takes whatever the internal cell representation is (index or `Cell`) and returns an index, which lets me DRY on the accessor methods by avoiding overriding them in `fast_temporal_memory.py`.

chetan51:
@SaganBolliger Looks like only `burstColumns` in `fast_temporal_memory.py` uses `cellsForColumn`. Can we just have it call `temporal_memory.py`'s `cellsForColumn` and convert it into a list of `Cell`s?

SaganBolliger:
@chetan51 Good idea! I've pushed that change.

chetan51:
I think we can also remove override for `_validateCell` in `fast_temporal_memory` and remove the functions `getCellIndex` in `fast_temporal_memory` and `temporal_memory.py`.

SaganBolliger:
@chetan51 Agreed about `_validateCell`. In regards to `getCellIndex`, see how I'm using it in the accessors, [getActiveCells](https://github.com/numenta/nupic/pull/2962/files#diff-e9ba1e72c0fd6d9b4f05ef4117d3cfeeR676) for instance. I'm counting on the implementation of `getCellIndex` being different in `temporal_memory.py` vs `fast_temporal_memory.py` to avoid having to override the accessors in `fast_temporal_memory.py`. It keeps things DRY.

chetan51:
@SaganBolliger Ah, I see. Sounds good then.

chetan51:
:+1:

SaganBolliger:
@chetan51 CI checks passed. Can you merge?

PR #2979, opened 2016-01-25
Update TM performance test to use hotgym data, implement test reliability fix, and progress bar

### LINKED ISSUE #2971: Update TM performance test to use hotgym data
`temporal_memory_performance_test.py` should use hotgym data instead of fake data so that it is more representative of real results.

SaganBolliger:
Fixes #2971. Implements changes suggested in #2991 (does not close until changes are confirmed to fix the issue).

@chetan51 Ready for review.

### DIFF
--- a/tests/integration/nupic/algorithms/temporal_memory_performance_test.py
+++ b/tests/integration/nupic/algorithms/temporal_memory_performance_test.py
@@ -39,7 +39,7 @@ _INPUT_FILE_PATH = resource_filename(
   "nupic.datafiles", "extra/hotgym/rec-center-hourly.csv"
 )
 
-NUM_PATTERNS = 100
+NUM_PATTERNS = 2000
 
 
 # ==============================
--- a/tests/integration/nupic/algorithms/temporal_memory_performance_test.py
+++ b/tests/integration/nupic/algorithms/temporal_memory_performance_test.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # ----------------------------------------------------------------------
 # Numenta Platform for Intelligent Computing (NuPIC)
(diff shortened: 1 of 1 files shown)


tests/integration/nupic/algorithms/temporal_memory_performance_test.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # ----------------------------------------------------------------------
 # Numenta Platform for Intelligent Computing (NuPIC)
-# Copyright (C) 2014, Numenta, Inc.  Unless you have an agreement
```
chetan51: Should be 2014-2016

chetan51:
Done with review.

SaganBolliger:
Investigating AV failure with compatibility test. See #2980.

SaganBolliger:
Compatibility test does not pass. See issue #2986. May be related to performance difference.

SaganBolliger:
I've changed the test to run the first 2000 records instead of all of hotgym since the latter was causing timeouts in CI.

chetan51:
@rhyolight Can you restart the AppVeyor build?

rhyolight:
Build is now queued.

SaganBolliger:
@chetan51 @scottpurdy CI has now passed, but see my [comment here](https://github.com/numenta/nupic/issues/2991#issuecomment-179554112). Should we merge?

chetan51:
@scottpurdy What do you think?

SaganBolliger:
Looks like the AV build was restarted and then cancelled. Not sure why?

scottpurdy:
:+1:

PR #3000, opened 2016-02-04
Fix: TM compatibility tests fail when predictedSegmentDecrement is nonzero

chetan51:
Starts on https://github.com/numenta/nupic/issues/2999.

Issue #2999, opened 2016-02-04
TM compatibility tests fail when predictedSegmentDecrement is nonzero


chetan51:
Sister issue for https://github.com/numenta/nupic.core/issues/851.

PR #3006, opened 2016-02-09
Lower TM epsilon threshold for compatibility.

scottpurdy:
Replaces #3000 
Starts on #2999 

Probably depends on numenta/nupic.core#860


scottpurdy:
@subutai please review

subutai:
:+1:

PR #3014, opened 2016-02-12
Fix performance test

### LINKED ISSUE #2991: TM performance tests are unreliable
The temporal memory [performance test](https://github.com/numenta/nupic/blob/master/tests/integration/nupic/algorithms/temporal_memory_performance_test.py) passes unreliably in CI and locally. Sometimes the test fails to show that the C++ TM is faster than python. See for instance [this build failure](https://ci.appveyor.com/project/numenta-ci/nupic/build/0.1.0.1008). This should be further investigated. However, it is possible that these failures are not symptomatic of worse performance of the C++ implementation and rather are due to random fluctuations of the server load affecting the real time it takes to run each of the TM implementations. If this is the case, one possible solution is to train each model for only one time step in each iteration of the outer loop over time steps, rather than the current test that fully trains each model before proceeding to the next model. This proposal in pseudocode: 
```python
for record in records:
  for model in models:
    times[model] += real time of model.compute(record)
```
This would reduce the tendency for fluctuations in server load to bias the results of the test.

This issue was found while testing #2979

@chetan51 @scottpurdy

scottpurdy:
This changes the performance test to be a manually-run command line script that just prints out the results, making no assumptions about how fast the different algorithms should be. An automated test isn't feasible since it will produce different results on different platforms.

@subutai please review

fixes #2991


subutai:
@scottpurdy This looks good but since it is a script that should be manually invoked, I'd recommend putting it into scripts/profiling rather than in nupic/research.

subutai:
:+1: Other than that comment.

PR #3035, opened 2016-03-05
Fix bugs and inconsistencies in the custom region demo

### LINKED ISSUE #3031: The "custom region demo" doesn't do what it says it does
There are a couple of issues in:

/examples/network/custom_region/identity_region.py
/examples/network/custom_region_demo.py

## Issue 1

It says it passes "the encoded representation to an Identity Region". But it actually passes the unencoded input. 

This line uses the wrong source on the RecordSensor:

~~~python
  network.link("sensor", "identityRegion", "UniformLink", "",
               srcOutput="sourceOut", destInput="in")
~~~

It should use srcOutput "dataOut", or just use the default.

~~~python
  network.link("sensor", "identityRegion", "UniformLink", "")
~~~

## Issue 2

The IdentityRegion is not actually writing to the output. It's reassigning the output variable to reference a totally different array.

The fix: Change

~~~python
outputs["out"] = inputs["in"]
~~~

to

~~~python
outputs["out"][:] = inputs["in"]
~~~

## Issue 3

Now that we've solved those, we need to put the correct `dataWidth` in the IdentityRegion.

Basically, change the line:

~~~python
network.addRegion("identityRegion", "py.IdentityRegion", json.dumps(I_PARAMS))
~~~

to:

~~~python
network.addRegion("identityRegion", "py.IdentityRegion", json.dumps({
        "dataWidth": sensor.encoder.getWidth(),
    }))
~~~

mrcslws:
Fixes #3031 

It's weird that the IdentityRegion doesn't infer the output size from the input size, but a quick look tells me that this is a flaw in Python regions in general. It's possible to do that with C++ regions, but not Python regions. I can't find a way to access inputs from the `initialize` method. (The parameter names are lies -- they're not "inputs, outputs", they're "dims, splitterMaps"). You can see incorrect args in the nupic bindings... https://github.com/numenta/nupic.core/blob/master/bindings/py/nupic/bindings/regions/PyRegion.py#L131

Also `getOutputElementCount` is getting called before `initialize`, which seems like a bug.

So I fixed the inconsistencies in this demo, but there's more to investigate here.

### DIFF
--- a/examples/network/custom_region/identity_region.py
+++ b/examples/network/custom_region/identity_region.py
@@ -33,11 +33,10 @@ class IdentityRegion(PyRegion):
   def __init__(self, dataWidth):
     if dataWidth <= 0:
       raise ValueError("Parameter dataWidth must be > 0")
-    
     self._dataWidth = dataWidth
 
 
-  def initialize(self, inputs, outputs):
+  def initialize(self, dims, splitterMaps):
     pass
 
 
@@ -45,14 +44,12 @@ class IdentityRegion(PyRegion):
     """
     Run one iteration of IdentityRegion's compute
     """
-    outputs["out"] = inputs["in"]
+    outputs["out"][:] = inputs["in"]
 
 
   @classmethod
   def getSpec(cls):
-    """Return the base Spec for MySPRegion.
-
-    Doesn't include the spatial, temporal and other parameters
+    """Return the Spec for IdentityRegion.
     """
     spec = {
         "description":IdentityRegion.__doc__,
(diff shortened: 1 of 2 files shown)


mrcslws:
**Not required reading**

Reporting back, it turns out I don't know how to make a C++ region check an input size when determining its output size. It seems like it should work, but the input link is still not "initialized" at the point that it's determining the output size.

Also,

> `getOutputElementCount` is getting called before `initialize`

 is the same in C++, so disregard me calling that a bug.

Anyway this commentary is independent of this change.

Issue #2270, opened 2015-06-29
KNN Classifier Unittests


rmccall84:
There are several possible instances of KNN Classifier that one can make by specifying the following constructor arguments and, ideally, we should have adequate test coverage for the different possibilities. At a minimum I feel we should test learn() and infer(). 

- [ ] useSparseMemory True 
- [ ] useSparseMemory False

distanceMethod (string)
- [ ] "norm"
- [ ] "rawOverlap"
- [ ] "pctOverlapOfLarger"
- [ ] "pctOverlapOfProto"

rmccall84:
@mihail911 @BoltzmannBrain FYI, recreating the tasks here.

Issue #2172, opened 2015-06-03
Redo/add kNN classifier tests


BoltzmannBrain:
The [kNN classifier integration tests](https://github.com/numenta/nupic/tree/master/tests/integration/nupic/algorithms/knn_classifier_test) are difficult to follow and inconsistently formatted. They should be redone in standard testing framework. We should also add unit test coverage.

PR #3042, opened 2016-03-08
Added minSparsity option

### LINKED ISSUE #2975: Add minSparsity option to KNNClassifierRegion
Add minSparsity option to KNNClassifierRegion. This option would only add a vector if its sparsity is >= minSparsity. This is useful in situations where the input sparsity varies greatly and you want to have some control over minSparsity. Default would be to have the same behavior as now, which is to add all vectors (minSparsity=0.0).

subutai:
Fixes #2975

### DIFF
--- a/src/nupic/algorithms/KNNClassifier.py
+++ b/src/nupic/algorithms/KNNClassifier.py
@@ -91,7 +91,8 @@ class KNNClassifier(object):
                      verbosity=0,
                      maxStoredPatterns=-1,
                      replaceDuplicates=False,
-                     cellsPerCol=0):
+                     cellsPerCol=0,
+                     minSparsity=0.0):
     """Constructor for the kNN classifier.
 
     @param k (int) The number of nearest neighbors used in the classification of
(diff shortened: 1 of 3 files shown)


subutai:
@scottpurdy @BoltzmannBrain 

The implementation is a bit awkward because I followed the existing style.  The KNN implementation could use a bit of cleanup (not addressed in this PR).

scottpurdy:
:+1: nice tests

BoltzmannBrain:
:+1: 
@subutai the KNN can definitely use some cleaning up. Started to collect some of the issues last summer (#2271, #2270, #2172), which can use some reorganization when we decide to actually clean it up.

subutai:
Thanks. Just merged latest from master into this PR - will merge pull request as soon as tests pass again.

src/nupic/algorithms/KNNClassifier.py
```diff
@@ -201,6 +207,7 @@ class labels
     self.replaceDuplicates = replaceDuplicates
     self.cellsPerCol = cellsPerCol
     self.maxStoredPatterns = maxStoredPatterns
+    self.minSparsity = minSparsity
```
oxtopus: I believe you need to add something in `__getstate__()`/`__setstate__()` to handle deserialized objects for which `self.minSparsity` has not been set.  Otherwise https://github.com/subutai/nupic/blob/d6e04686f78a27e1dbd5ed20ac6c283dba94ccdd/src/nupic/algorithms/KNNClassifier.py#L545 can cause an `AttributeError` exception, as I'm currently experiencing in Imbu.

subutai: @oxtopus  I was hoping I didn't need to worry about backward compatibility here since this class is not used much outside of research, but I can do that.

oxtopus: Thanks.  That, or fix it in Imbu, which I think would involve retraining and serializing models.

Issue #3053, opened 2016-03-16
Implement SDR Classifier in NuPIC


ywcui1990:
The SDR classifier is a single layer classification network that maps SDRs to class labels. Preliminary experiments demonstrate that the classifier gives better result than the CLA classifier on both artificial and real world datasets. This task is to port the SDR classifier (currently in nupic.research) to nupic, and create unit tests for it.

breznak:
:+1: watching..

PR #3066, opened 2016-03-23
Implement SDR classifier in NuPIC

ywcui1990:
@subutai @scottpurdy 

Fixes: https://github.com/numenta/nupic/issues/3053

- Implement SDR classifier in Nupic
- Add unit tests for SDR classifier. These tests are similar to the cla_classifier_test, except for the obsolete ones and the testWriteRead (due to lack of SDRClassifierProto)

### DIFF
--- a/src/nupic/algorithms/sdr_classifier.py
+++ b/src/nupic/algorithms/sdr_classifier.py
@@ -1,6 +1,6 @@
 # ----------------------------------------------------------------------
 # Numenta Platform for Intelligent Computing (NuPIC)
-# Copyright (C) 2014, Numenta, Inc.  Unless you have an agreement
+# Copyright (C) 2016, Numenta, Inc.  Unless you have an agreement
 # with Numenta, Inc., for a separate license for this software code, the
 # following terms and conditions apply:
 #
@@ -20,7 +20,7 @@
 # ----------------------------------------------------------------------
 
 """
-Implementation for a SDR classifier.
+Implementation of a SDR classifier.
 
 The SDR classifier takes the form of a single layer classification network
 that takes SDRs as input and outputs a predicted distribution of classes.
@@ -47,8 +47,8 @@ class SDRClassifier(object):
   encoders (the "classification") describing the true (target) input.
 
   The SDR classifier maps input patterns to class labels. There are as many
-  output units as the number of class labels (buckets). The output is a
-  probabilistic distribution over all class labels
+  output units as the number of class labels or buckets (in the case of scalar
+  encoders). The output is a probabilistic distribution over all class labels.
 
   During inference, the output is calculated by first doing a weighted summation
   of all the inputs, and then perform a softmax nonlinear function to get
@@ -79,15 +79,26 @@ class SDRClassifier(object):
 
     Parameters:
     ---------------------------------------------------------------------
-    @param numInputs (int) Length of the input activation pattern
     @param steps (list) Sequence of the different steps of multi-step
         predictions to learn
     @param alpha (float) The alpha used to adapt the weight matrix during
-        learning. A lower alpha results in longer term memory.
+        learning. A larger alpha results in faster adaptation to the data.
     @param actValueAlpha (float) Used to track the actual value within each
         bucket. A lower actValueAlpha results in longer term memory
     @param verbosity (int) verbosity level, can be 0, 1, or 2
     """
+    if len(steps) == 0:
+      raise TypeError("steps cannot be empty")
+    if not all(isinstance(item, int) for item in steps):
+      raise TypeError("steps must be a list of ints")
+    if any(item < 0 for item in steps):
+      raise ValueError("steps must be a list of non-negative ints")
+
+    if alpha < 0:
+      raise ValueError("alpha (learning rate) must be a positive number")
+    if actValueAlpha < 0 or actValueAlpha >= 1:
+      raise ValueError("actValueAlpha be a number between 0 and 1")
+
     # Save constructor args
     self.steps = steps
     self.alpha = alpha
@@ -176,6 +187,8 @@ class SDRClassifier(object):
                    'actualValues': [1.5, 3,5, 5,5, 7.6],
                   }
     """
+    if learn is False and infer is False:
+      raise ValueError("learn and infer cannot be both False")
 
     # Save the offset between recordNum and learnIteration if this is the first
     #  compute
@@ -241,7 +254,9 @@ class SDRClassifier(object):
       if self._actualValues[bucketIdx] is None:
         self._actualValues[bucketIdx] = actValue
       else:
-        if isinstance(actValue, int) or isinstance(actValue, float):
+        if (isinstance(actValue, int) or
+              isinstance(actValue, float) or
+              isinstance(actValue, long)):
           self._actualValues[bucketIdx] = ((1.0 - self.actValueAlpha)
                                            * self._actualValues[bucketIdx]
                                            + self.actValueAlpha * actValue)
@@ -322,9 +337,15 @@ class SDRClassifier(object):
 
 
   def inferSingleStep(self, patternNZ, weightMatrix):
-    outputActivation = numpy.zeros(self._maxBucketIdx + 1)
-    for bit in patternNZ:
-      outputActivation += weightMatrix[bit, :]
+    """
+    Perform inference for a single step. Given an SDR input and a weight
+    matrix, return a predicted distribution.
+
+    @param patternNZ  list of the active indices from the output below
+    @param weightMatrix numpy array of the weight matrix
+    @return numpy array of the predicted class label distribution
+    """
+    outputActivation = weightMatrix[patternNZ].sum(axis=0)
 
     # softmax normalization
     expOutputActivation = numpy.exp(outputActivation)
--- a/src/nupic/algorithms/sdr_classifier.py
+++ b/src/nupic/algorithms/sdr_classifier.py
@@ -54,7 +54,7 @@ class SDRClassifier(object):
 
   Example Usage:
 
-  c = SDRClassifier(steps=[0], alpha=0.1, actValueAlpha=0.1, verbosity=0)
+  c = SDRClassifier(steps=[1], alpha=0.1, actValueAlpha=0.1, verbosity=0)
 
   # learning
   c.compute(recordNum=0, patternNZ=[1, 5, 9],
@@ -65,6 +65,13 @@ class SDRClassifier(object):
   result = c.compute(recordNum=1, patternNZ=[1, 5, 9],
                      classification={"bucketIdx": 4, "actValue": 34.7},
                      learn=False, infer=True)
+
+  # Print the top three predictions for 1 steps out.
+  topPredictions = sorted(zip(result[1],
+                          result["actualValues"]), reverse=True)[:3]
+  for probability, value in topPredictions:
+    print "Prediction of {} has probability of {}.".format(value,
+                                                           probability*100.0)
 
   References:
     Alex Graves. Supervised Sequence Labeling with Recurrent Neural Networks
--- a/src/nupic/algorithms/sdr_classifier.py
+++ b/src/nupic/algorithms/sdr_classifier.py
@@ -27,17 +27,9 @@ that takes SDRs as input and outputs a predicted distribution of classes.
 """
 
 from collections import deque
-import itertools
 
 import numpy
 
-g_debugPrefix = "SDRClassifier"
-
-
-def _pFormatArray(array_, fmt="%.2f"):
-  """Return a string with pretty-print of a numpy array using the given format
-  for each element"""
-  return "[ " + " ".join(fmt % x for x in array_) + " ]"
 
 
 class SDRClassifier(object):
@@ -57,6 +49,23 @@ class SDRClassifier(object):
   During learning, the connection weights between input units and output units
   are adjusted to maximize the likelihood of the model
 
+  The SDR Classifier is a variation of the previous CLAClassifier which was
+  not based on the references below.
+
+  Example Usage:
+
+  c = SDRClassifier(steps=[0], alpha=0.1, actValueAlpha=0.1, verbosity=0)
+
+  # learning
+  c.compute(recordNum=0, patternNZ=[1, 5, 9],
+            classification={"bucketIdx": 4, "actValue": 34.7},
+            learn=True, infer=False)
+
+  # inference
+  result = c.compute(recordNum=1, patternNZ=[1, 5, 9],
+                     classification={"bucketIdx": 4, "actValue": 34.7},
+                     learn=False, infer=True)
+
   References:
     Alex Graves. Supervised Sequence Labeling with Recurrent Neural Networks
     PhD Thesis, 2008
@@ -67,7 +76,7 @@ class SDRClassifier(object):
     Architectures and Applications, pp 227-236, Springer-Verlag, 1990
   """
 
-  __VERSION__ = 1
+  VERSION = 1
 
 
   def __init__(self,
@@ -148,7 +157,7 @@ class SDRClassifier(object):
 
     # Set the version to the latest version.
     # This is used for serialization/deserialization
-    self._version = SDRClassifier.__VERSION__
+    self._version = SDRClassifier.VERSION
 
 
   def compute(self, recordNum, patternNZ, classification, learn, infer):
@@ -170,6 +179,7 @@ class SDRClassifier(object):
     @param classification Dict of the classification information:
                     bucketIdx: index of the encoder bucket
                     actValue:  actual value going into the encoder
+                    classification could be None for inference mode
     @param learn (bool) if true, learn this sample
     @param infer (bool) if true, perform inference
 
@@ -264,7 +274,7 @@ class SDRClassifier(object):
           self._actualValues[bucketIdx] = actValue
 
       for (iteration, learnPatternNZ) in self._patternNZHistory:
-        error = self.calculateError(classification)
+        error = self._calculateError(classification)
 
         nSteps = self._learnIteration - iteration
         if nSteps in self.steps:
@@ -320,7 +330,7 @@ class SDRClassifier(object):
 
     # NOTE: If doing 0-step prediction, we shouldn't use any knowledge
     #  of the classification input during inference.
-    if self.steps[0] == 0:
+    if self.steps[0] == 0 or classification is None:
       defaultValue = 0
     else:
       defaultValue = classification["actValue"]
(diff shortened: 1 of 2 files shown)


src/nupic/algorithms/sdr_classifier.py
```diff
@@ -0,0 +1,354 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2014, Numenta, Inc.  Unless you have an agreement
```
subutai: 2016

src/nupic/algorithms/sdr_classifier.py
```diff
@@ -0,0 +1,354 @@
...
+
+    for nSteps in self.steps:
+      predictDist = self.inferSingleStep(patternNZ, self._weightMatrix[nSteps])
+
+      retval[nSteps] = predictDist
+
+    return retval
+
+
+  def inferSingleStep(self, patternNZ, weightMatrix):
```
subutai: Docstring?

src/nupic/algorithms/sdr_classifier.py
```diff
@@ -0,0 +1,354 @@
...
+# See the GNU Affero Public License for more details.
+#
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+"""
+Implementation for a SDR classifier.
```
subutai: "Implementation of a SDR classifier"

src/nupic/algorithms/sdr_classifier.py
```diff
@@ -0,0 +1,354 @@
...
+
+
+class SDRClassifier(object):
+  """
+  The SDR Classifier accepts a binary input pattern from the
+  level below (the "activationPattern") and information from the sensor and
+  encoders (the "classification") describing the true (target) input.
+
+  The SDR classifier maps input patterns to class labels. There are as many
+  output units as the number of class labels (buckets). The output is a
```
subutai: "(or buckets in the case of scalar encoders)"

src/nupic/algorithms/sdr_classifier.py
```diff
@@ -0,0 +1,354 @@
...
+               verbosity=0):
+    """Constructor for the SDR classifier.
+
+    Parameters:
+    ---------------------------------------------------------------------
+    @param numInputs (int) Length of the input activation pattern
+    @param steps (list) Sequence of the different steps of multi-step
+        predictions to learn
+    @param alpha (float) The alpha used to adapt the weight matrix during
+        learning. A lower alpha results in longer term memory.
```
subutai: Does alpha really control how long the memory is stored? (If an SDR is not seen for a really long time it will not forget it regardless of alpha.) It seems like alpha just controls how fast it adapts?

src/nupic/algorithms/sdr_classifier.py
```diff
@@ -0,0 +1,354 @@
...
+
+      # Update rolling average of actual values if it's a scalar. If it's
+      # not, it must be a category, in which case each bucket only ever
+      # sees one category so we don't need a running average.
+      while self._maxBucketIdx > len(self._actualValues) - 1:
+        self._actualValues.append(None)
+      if self._actualValues[bucketIdx] is None:
+        self._actualValues[bucketIdx] = actValue
+      else:
+        if isinstance(actValue, int) or isinstance(actValue, float):
```
subutai: Do we need to check for "long" as well?

src/nupic/algorithms/sdr_classifier.py
```diff
@@ -0,0 +1,354 @@
...
+    @param steps (list) Sequence of the different steps of multi-step
+        predictions to learn
+    @param alpha (float) The alpha used to adapt the weight matrix during
+        learning. A lower alpha results in longer term memory.
+    @param actValueAlpha (float) Used to track the actual value within each
+        bucket. A lower actValueAlpha results in longer term memory
+    @param verbosity (int) verbosity level, can be 0, 1, or 2
+    """
+    # Save constructor args
+    self.steps = steps
```
subutai: We should check that steps, alpha, and actValueAlpha are legal values and expected types. (e.g. alpha should be in [0,1]).

src/nupic/algorithms/sdr_classifier.py
```diff
@@ -0,0 +1,354 @@
...
+
+      retval[nSteps] = predictDist
+
+    return retval
+
+
+  def inferSingleStep(self, patternNZ, weightMatrix):
+    outputActivation = numpy.zeros(self._maxBucketIdx + 1)
+    for bit in patternNZ:
+      outputActivation += weightMatrix[bit, :]
```
subutai: Can we do something like outputActivation += weightMatrix[patternNZ].sum()?  It might be faster than the loop.

tests/unit/nupic/algorithms/sdr_classifier_test.py
```diff
@@ -0,0 +1,445 @@
...
+#
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+"""Unit tests for SDRClassifier module."""
+
+CL_VERBOSITY = 0
```
subutai: This is not used anywhere.

tests/unit/nupic/algorithms/sdr_classifier_test.py
```diff
@@ -0,0 +1,445 @@
...
+    self.assertEqual(pretty[0], "[")
+    self.assertEqual(pretty[-1], "]")
+    self.assertEqual(len(pretty.split(" ")), 12)
+
+
+  def _checkValue(self, retval, index, value):
+    self.assertEqual(retval['actualValues'][index], value)
+
+
+  def _checkProbabilityGreater(self, retval, index, probability):
```
subutai: Not used anywhere.

tests/unit/nupic/algorithms/sdr_classifier_test.py
```diff
@@ -0,0 +1,445 @@
...
+    # Enough times to perform Inference and learn associations
+    retval = []
+    for recordNum in xrange(10):
+      retval = self._compute(classifier, recordNum, [1, 5], 0, 10)
+
+    self.assertEqual(retval['actualValues'][0], 10)
+    self.assertGreater(retval[1][0], 0.9)
+
+
+  def testSingleValue0Steps(self):
```
subutai: Can we have a test with steps=[0] that distinguishes between steps=[1]? If we alternate the classification between two buckets, 0-step should learn one of them and 1-step should learn the other.

tests/unit/nupic/algorithms/sdr_classifier_test.py
```diff
@@ -0,0 +1,445 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero Public License for more details.
+#
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+"""Unit tests for SDRClassifier module."""
```
subutai: We should have tests that ensure exceptions are raised properly when bad values are passed in.

tests/unit/nupic/algorithms/sdr_classifier_test.py
```diff
@@ -0,0 +1,445 @@
...
+  """Unit tests for SDRClassifier class."""
+
+
+  def setUp(self):
+    self._classifier = SDRClassifier
+
+
+  def testInitialization(self):
+    c = self._classifier([1], 0.1, 0.1, 0)
+    self.assertEqual(type(c), self._classifier)
```
subutai: We can add some tests of constructors with bad parameters - they should throw exceptions.

tests/unit/nupic/algorithms/sdr_classifier_test.py
```diff
@@ -0,0 +1,445 @@
...
+              classification={'bucketIdx': 4, 'actValue': 34.7},
+              learn=True, infer=True)
+    result = c.compute(recordNum=1, patternNZ=[1, 5, 9],
+                       classification={'bucketIdx': 4, 'actValue': 34.7},
+                       learn=True, infer=True)
+    self.assertSetEqual(set(result.keys()), set(('actualValues', 1)))
+    self.assertAlmostEqual(result['actualValues'][4], 34.7, places=5)
+
+
+  def testComputeComplex(self):
```
subutai: We don't see to have test cases where learn=False and infer=True, or vice versa.  Also, if both are False, an exception should be raised.

tests/unit/nupic/algorithms/sdr_classifier_test.py
```diff
@@ -0,0 +1,445 @@
...
+              patternNZ=[6, 9],
+              classification={'bucketIdx': 5, 'actValue': 44.9},
+              learn=True, infer=True)
+    c.compute(recordNum=3,
+              patternNZ=[1, 5, 9],
+              classification={'bucketIdx': 4, 'actValue': 42.9},
+              learn=True, infer=True)
+    serialized = pickle.dumps(c)
+    c = pickle.loads(serialized)
+    result = c.compute(recordNum=4,
```
subutai: We should also check that the member variables are set as expected after deserialization.  For example, alpha could be incorrect after unpickling, and this test would not catch that.

subutai:
Done with my initial review. Overall the code looks very clean and well written. Left suggestions for more tests plus some more we discussed on paper.

ywcui1990:
@subutai 

Thanks for the code review. I have incorporated your comments and added 3 additional tests, which are 
(1) test that the classifier captured the probability distribution given enough repetition 
(2) test that the classifier are robust to shared bits in the input SDRs and 
(3) test that the classifier adapts to changes in the data streams and remembers the unchanged part

tests/unit/nupic/algorithms/sdr_classifier_test.py
```diff
@@ -0,0 +1,711 @@
...
+      elif randomNumber < 0.7:
+        bucketIdx = 2
+      else:
+        bucketIdx = 4
+      c.compute(recordNum=recordNum, patternNZ=SDR1,
+                classification={'bucketIdx': bucketIdx, 'actValue': bucketIdx},
+                learn=True, infer=True)
+      recordNum += 1
+
+      randomNumber = random.random()
```
subutai: In this second phase, can we leave out the second mapping completely? The classifier should not touch the bits for SDR2 so it should still map to 50%/50%.

tests/unit/nupic/algorithms/sdr_classifier_test.py
```diff
@@ -0,0 +1,711 @@
...
+      if randomNumber < 0.5:
+        bucketIdx = 1
+      else:
+        bucketIdx = 3
+      c.compute(recordNum=recordNum, patternNZ=SDR2,
+                classification={'bucketIdx': bucketIdx, 'actValue': bucketIdx},
+                learn=True, infer=True)
+      recordNum += 1
+
+    for _ in xrange(30000):
```
subutai: Do we really need 10,000 and 30,000 iterations for this test? It would be good if we could do this much quicker.

tests/unit/nupic/algorithms/sdr_classifier_test.py
```diff
@@ -0,0 +1,711 @@
...
+    random.seed(42)
+    for _ in xrange(10000):
+      randomNumber = random.random()
+      if randomNumber < 0.3:
+        bucketIdx = 0
+      elif randomNumber < 0.6:
+        bucketIdx = 1
+      else:
+        bucketIdx = 2
+      c.compute(recordNum=recordNum, patternNZ=SDR1,
```
subutai: You can probably set infer to False here.

src/nupic/algorithms/sdr_classifier.py
```diff
@@ -0,0 +1,375 @@
...
+    # softmax normalization
+    expOutputActivation = numpy.exp(outputActivation)
+    predictDist = expOutputActivation / numpy.sum(expOutputActivation)
+    return predictDist
+
+
+  def calculateError(self, classification):
+    """
+    Calculate error signal
+    :param classification:
```
subutai: Should use a consistent docstring style, i.e. @param

src/nupic/algorithms/sdr_classifier.py
```diff
@@ -0,0 +1,375 @@
...
+
+from collections import deque
+import itertools
+
+import numpy
+
+g_debugPrefix = "SDRClassifier"
+
+
+def _pFormatArray(array_, fmt="%.2f"):
```
scottpurdy: Put this at the end of the file. It just gets in the way up here

src/nupic/algorithms/sdr_classifier.py
```diff
@@ -0,0 +1,375 @@
...
+  encoders). The output is a probabilistic distribution over all class labels.
+
+  During inference, the output is calculated by first doing a weighted summation
+  of all the inputs, and then perform a softmax nonlinear function to get
+  the predicted distribution of class labels
+
+  During learning, the connection weights between input units and output units
+  are adjusted to maximize the likelihood of the model
+
+  References:
```
scottpurdy: Mention that it is a slight variation of the previous CLAClassifier that was not based on this reference. Also include a simple code snippet showing the usage of the class (instantiation and getting predictions out).

src/nupic/algorithms/sdr_classifier.py
```diff
@@ -0,0 +1,375 @@
...
+The SDR classifier takes the form of a single layer classification network
+that takes SDRs as input and outputs a predicted distribution of classes.
+"""
+
+from collections import deque
+import itertools
+
+import numpy
+
+g_debugPrefix = "SDRClassifier"
```
scottpurdy: Remove this. It is only used once and you can just put the string there (or use `type(self).__name__`)

src/nupic/algorithms/sdr_classifier.py
```diff
@@ -0,0 +1,375 @@
...
+    Alex Graves. Supervised Sequence Labeling with Recurrent Neural Networks
+    PhD Thesis, 2008
+
+    J. S. Bridle. Probabilistic interpretation of feedforward classification
+    network outputs, with relationships to statistical pattern recognition.
+    In F. Fogleman-Soulie and J.Herault, editors, Neurocomputing: Algorithms,
+    Architectures and Applications, pp 227-236, Springer-Verlag, 1990
+  """
+
+  __VERSION__ = 1
```
scottpurdy: All caps without underscores is probably better for this (matches our naming conventions)

src/nupic/algorithms/sdr_classifier.py
```diff
@@ -0,0 +1,375 @@
...
+    """
+    outputActivation = weightMatrix[patternNZ].sum(axis=0)
+
+    # softmax normalization
+    expOutputActivation = numpy.exp(outputActivation)
+    predictDist = expOutputActivation / numpy.sum(expOutputActivation)
+    return predictDist
+
+
+  def calculateError(self, classification):
```
scottpurdy: Is this part of the public interface? If it is only meant to be used internally then include a single underscore prefix: `_ calculateError`

tests/unit/nupic/algorithms/sdr_classifier_test.py
```diff
@@ -0,0 +1,711 @@
...
+
+
+  @staticmethod
+  def _compute(classifier, recordNum, pattern, bucket, value):
+    classification = {'bucketIdx': bucket, 'actValue': value}
+    return classifier.compute(recordNum, pattern, classification, True, True)
+
+
+
+if __name__ == '__main__':
```
scottpurdy: double quotes

tests/unit/nupic/algorithms/sdr_classifier_test.py
```diff
@@ -0,0 +1,711 @@
...
+    self.assertEqual(len(pretty.split(" ")), 12)
+
+
+  def _checkValue(self, retval, index, value):
+    self.assertEqual(retval['actualValues'][index], value)
+
+
+  @staticmethod
+  def _compute(classifier, recordNum, pattern, bucket, value):
+    classification = {'bucketIdx': bucket, 'actValue': value}
```
scottpurdy: double quotes here and elsewhere

scottpurdy:
done reviewing

ywcui1990:
@subutai @scottpurdy Thanks for the code review for both of you. 

I incorporated your feedback. In addition, 
* Added a test for simultaneous multi-step prediction
* Sped up the unit tests. It takes less than 2 secs to run all the tests on my computer
* Allowed classification to be "None" in pure inference mode

src/nupic/algorithms/sdr_classifier.py
```diff
@@ -0,0 +1,393 @@
...
+
+  # learning
+  c.compute(recordNum=0, patternNZ=[1, 5, 9],
+            classification={"bucketIdx": 4, "actValue": 34.7},
+            learn=True, infer=False)
+
+  # inference
+  result = c.compute(recordNum=1, patternNZ=[1, 5, 9],
+                     classification={"bucketIdx": 4, "actValue": 34.7},
+                     learn=False, infer=True)
```
scottpurdy: The tricky part is interpreting `result`. Can you follow this with a loop that prints each prediction and associated predicted value (only do this for one of the steps). For example, something like this:

```
# Print the top three predictions for 5 steps out.
for probability, value in sorted(zip(result[5], result["actualValues"]), reverse=True)[:3]:
  print "Prediction of {} has probability of {}.".format(value, probability*100.0)
```

src/nupic/algorithms/sdr_classifier.py
```diff
@@ -0,0 +1,393 @@
...
+
+  During learning, the connection weights between input units and output units
+  are adjusted to maximize the likelihood of the model
+
+  The SDR Classifier is a variation of the previous CLAClassifier which was
+  not based on the references below.
+
+  Example Usage:
+
+  c = SDRClassifier(steps=[0], alpha=0.1, actValueAlpha=0.1, verbosity=0)
```
subutai: I think it would be better to use steps=[1] in this example - that's the most common case.

subutai:
One minor comment and agree with @scottpurdy 's comment. I have no further issues - all looks good to me! :+1:

ywcui1990:
I have fixed the example code snippet. It should be ready to merge if you have no further comments @scottpurdy

cogmission:
@ywcui1990 Please see https://github.com/numenta/nupic/issues/3067

Issue #3067, opened 2016-03-24
new sdr_classifier (Yay!) has erroneous documentation?


cogmission:
Hi,

This line https://github.com/numenta/nupic/blob/master/src/nupic/algorithms/sdr_classifier.py#L313

mentions a "customCompute()" method that doesn't seem to exist? Will it be added later or is there an inherited class that implements that method?

PR #3068, opened 2016-03-24
Fix erroneous documentation in SDR classifier

ywcui1990:
Fixes https://github.com/numenta/nupic/issues/3067
@cogmission Thanks for catching the erroneous documentation. It was a legacy of the CLAclassifier. I have removed it.


cogmission:
I figured it was something like that. Is this classifier a drop in replacement for the CLAClassifier?

subutai:
The perils of copy/paste. :+1:

rhyolight:
Good catch @cogmission

PR #3082, opened 2016-04-06
Fix to GCE to return the right number of scalars when altitude is missing.

### LINKED ISSUE #3077: GeospatialCoordinateEncoder "altitude" seems optional, but not really
I ran into difficulties generating anomaly likelihood values with a GeospatialCoordinateEncoder (GCE), and I believe the root cause is that the `altitude` portion of the raw input vector.

In all current examples of GCE usage, the input vector is [created like this](https://github.com/numenta/nupic.geospatial/blob/master/model/geospatial_anomaly.py#L136-L138):

```python
modelInput = {
    "vector": (speed, longitude, latitude, altitude)
}
```

... where `altitude` may be `None`. This starts breaking when you use the anomaly likelihood post-process, because it introduces arrays with `None` values, and numpy complains when trying to `mean` the data:

```
  File "/Users/mtaylor/Library/Python/2.7/lib/python/site-packages/nupic/algorithms/anomaly_likelihood.py", line 208, in anomalyProbability
    skipRecords=numSkipRecords)
  File "/Users/mtaylor/Library/Python/2.7/lib/python/site-packages/nupic/algorithms/anomaly_likelihood.py", line 364, in estimateAnomalyLikelihoods
    performLowerBoundCheck=False)
  File "/Users/mtaylor/Library/Python/2.7/lib/python/site-packages/nupic/algorithms/anomaly_likelihood.py", line 578, in estimateNormal
    "mean": numpy.mean(sampleData),
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/numpy/core/fromnumeric.py", line 2716, in mean
    out=out, keepdims=keepdims)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/numpy/core/_methods.py", line 62, in _mean
    ret = um.add.reduce(arr, axis=axis, dtype=dtype, out=out, keepdims=keepdims)
TypeError: unsupported operand type(s) for +: 'float' and 'NoneType'
```

I tried using `numpy.nanmean` but that didn't help. Another workaround is to use `0` for altitude, but this drastically increases processing time by adding another dimension to the input that doesn't really exist.

If you look at how the GCE [encodes input data](https://github.com/numenta/nupic/blob/master/src/nupic/encoders/geospatial_coordinate.py#L81-L90):

```python
altitude = None
if len(inputData) == 4:
  (speed, longitude, latitude, altitude) = inputData
else:
  (speed, longitude, latitude) = inputData
coordinate = self.coordinateForPosition(longitude, latitude, altitude)
radius = self.radiusForSpeed(speed)
super(GeospatialCoordinateEncoder, self).encodeIntoArray(
 (coordinate, radius), output)
```

You can see that `altitude` is **optionally** included in the coordinate. However, [the description](https://github.com/numenta/nupic/blob/master/src/nupic/encoders/geospatial_coordinate.py#L69-L71) **always** returns the altitude:

```python
  def getDescription(self):
    """See `nupic.encoders.base.Encoder` for more information."""
    return [('speed', 0), ('longitude', 1), ('latitude', 2), ('altitude', 3)]
```

I believe this is the root problem, but I'm not sure how to fix it. To get things running for my current project, I have just manually removed the `('altitude', 3)` from the description array, and I can now run NuPIC without an `altitude` and post-process the anomaly likelihood properly.

rhyolight:
Fixes #3077

### DIFF
--- a/src/nupic/encoders/geospatial_coordinate.py
+++ b/src/nupic/encoders/geospatial_coordinate.py
@@ -73,14 +73,7 @@ class GeospatialCoordinateEncoder(CoordinateEncoder):
 
   def getScalars(self, inputData):
     """See `nupic.encoders.base.Encoder` for more information."""
-    # This is a fix for the OPF in order to allow optional 'altitude' in the
-    # input vector. If the input length is less than what is defined in the
-    # getDescription function, it means the user is not providing altitude. So
-    # we just add one more to the length of the [0] we're returning.
-    inputLength = len(inputData)
-    while inputLength < len(self.getDescription()):
-      inputLength += 1
-    return numpy.array([0] * inputLength)
+    return numpy.array([0] * len(self.getDescription()))
 
 
   def encodeIntoArray(self, inputData, output):
--- a/src/nupic/encoders/geospatial_coordinate.py
+++ b/src/nupic/encoders/geospatial_coordinate.py
@@ -71,6 +71,18 @@ class GeospatialCoordinateEncoder(CoordinateEncoder):
     return [('speed', 0), ('longitude', 1), ('latitude', 2), ('altitude', 3)]
 
 
+  def getScalars(self, inputData):
+    """See `nupic.encoders.base.Encoder` for more information."""
+    # This is a fix for the OPF in order to allow optional 'altitude' in the
+    # input vector. If the input length is less than what is defined in the
+    # getDescription function, it means the user is not providing altitude. So
+    # we just add one more to the length of the [0] we're returning.
+    inputLength = len(inputData)
+    while inputLength < len(self.getDescription()):
+      inputLength += 1
+    return numpy.array([0] * inputLength)
+
+
   def encodeIntoArray(self, inputData, output):
     """
     See `nupic.encoders.base.Encoder` for more information.


src/nupic/encoders/geospatial_coordinate.py
```diff
@@ -71,6 +71,18 @@ def getDescription(self):
...
+  def getScalars(self, inputData):
+    """See `nupic.encoders.base.Encoder` for more information."""
+    # This is a fix for the OPF in order to allow optional 'altitude' in the
+    # input vector. If the input length is less than what is defined in the
+    # getDescription function, it means the user is not providing altitude. So
+    # we just add one more to the length of the [0] we're returning.
+    inputLength = len(inputData)
+    while inputLength < len(self.getDescription()):
+      inputLength += 1
+    return numpy.array([0] * inputLength)
```
scottpurdy: How about the entire function just be:

```
return numpy.array([0] * len(self.getDescription()))
```

scottpurdy:
Done reviewing.

rhyolight:
@scottpurdy Done, and no explanation needed.

scottpurdy:
👍

PR #3092, opened 2016-04-16
Fixed Username Regex in ClientJobsDAO

### LINKED ISSUE #3073: ClientJobsDAO DB Name Fix
While building and installing Nupic, faced issue with dot in the user name. Having a dot in the username will not work because most SQL languages use the dot as a relation. After replacing dot with underscore was able to install successfully.

fyi: @rhyolight

badlogicmanpreet:
While building and installing Nupic faced issue with dot in the user name. Having a dot in the username will not work because most SQL languages use the dot as a relation.

cc: @rhyolight 

Fixes #3073


rhyolight:
👍 Very nice, thank you!

PR #3093, opened 2016-04-17
Reserved slots in CategoryEncoder. Fixes #3094

### LINKED ISSUE #3094: Reserved slots to unknown categories in CategoryEncoder
Sometimes you have access to a dataset for training, but are not sure if there will be any new categories in the future. The Adaptative Scalar Encoder is not an ideal option, because it means that the SP and TM will have to re-map the adapted bits, loosing some of the work that has been done. The Random Scalar Encoder does not work well in certain situations, when you need to ensure each category is treated completely separately. I came up with a small update to the Category Encoder. When you create it, you can reserve some slots for potential new categories. When a new category is fed, the encoder maps it to one of the previously reserved slots without affecting any of the other categories. If the encoder runs out of slots, newer categories will be treated as Unknown.

I already made a pull request for this, sorry for not opening the issue before ( #3093 ).

larvasapiens:
Adds reserved slots to categories that are yet unknown at the time of creating the Category Encoder. When a new category is fed, the Encoder will update one of the reserved slots to encode that new category. If the Encoder runs out of reserved slots, it will return the <UNKOWN> category representation, as usual.

Fixes #3094

### DIFF
--- a/src/nupic/encoders/category.py
+++ b/src/nupic/encoders/category.py
@@ -29,7 +29,7 @@ from nupic.encoders.scalar import ScalarEncoder
 
 
 UNKNOWN = "<UNKNOWN>"
-
+DUMMY_CATEGORY = "__dummyCategory{}"
 
 
 class CategoryEncoder(Encoder):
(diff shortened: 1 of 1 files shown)


breznak:
hi @larvasapiens !
you will have to create an issue describing your...well, issue :wink: 

Additionally, is this approach for "unknown number of categories" needed? Why not use an AdaptiveScalar/RDSE encoder that: precision can be set to natural numbers, and each category is encoded as a number, the encoder will adapt to new ones... ?

larvasapiens:
I've opened an Issue explaining my reasons. #3094 Thanks for letting me know the correct procedure.

ywcui1990:
I like the idea to make the CategoryEncoder adaptive. Do we need to really hard code "nAdditionalCategorySlots" upfront? Since SDR has huge capacity, why cannot we dynamically add new categories in run time?

Issue #3054, opened 2016-03-16
Update CLA classifier region to allow the use of SDR classifier


ywcui1990:
The CLA classifier region currently uses the CLA classifier. This task is to add an option to the classifier region to allow the use of SDR classifier. The SDR classifier created in (https://github.com/numenta/nupic/issues/3053) has the same API as the CLA classifier.

PR #3095, opened 2016-04-18
Implement SDRClassifierRegion in NuPIC

ywcui1990:
Implements https://github.com/numenta/nupic/issues/3054
Depends on https://github.com/numenta/nupic.core/pull/923
Summary
* Implement Read/Write methods and tests for sdr_classifier.py
* Implement sdr_classifier_factory.py 
* Implement SDRClassifierRegion.py and SDRClassifierRegion.capnp
* Add SDRClassifierRegion to pyRegions
* Add default implementation of sdrClassifier in nupic_default.xml
* Verify that the OPF network runs and gives expected results after changing modelParams["clParams"]["regionName"] from "CLAClassifierRegion" to "SDRClassifierRegion"

Please review
@subutai @scottpurdy


src/nupic/algorithms/sdr_classifier_factory.py
```diff
@@ -0,0 +1,61 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2013, Numenta, Inc.  Unless you have an agreement
```
scottpurdy: nit: copyright year

scottpurdy:
Sorry for the delay on this. @oxtopus can you review this?

src/nupic/algorithms/sdr_classifier.py
```diff
@@ -369,6 +363,90 @@ def inferSingleStep(self, patternNZ, weightMatrix):
     return predictDist
 
 
+  @classmethod
+  def read(cls, proto):
+    classifier = object.__new__(cls)
+
+    classifier.steps = []
+    for step in proto.steps:
```
oxtopus: Only a performance impact with large values of `proto.steps`, but repeated `list.append()`'s can be slower than alternatives like list comprehensions.  For example:

```python
classifier.steps = [step for step in proto.steps]
```

For that matter, why not `classifier.steps = proto.steps`?

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,500 @@
...
+definition of SDRClassifierRegion for a description.
+"""
+
+import warnings
+
+from nupic.bindings.regions.PyRegion import PyRegion
+from nupic.algorithms.sdr_classifier_factory import SDRClassifierFactory
+from nupic.support.configuration import Configuration
+
+try:
```
oxtopus: If I understand correctly, this block is to gracefully handle the absence of `capnp`, but it should trigger a pylint error since `capnp` isn't used.  It'd be better if there were a first-class API that does this.

src/nupic/algorithms/sdr_classifier_factory.py
```diff
@@ -0,0 +1,61 @@
...
+from nupic.support.configuration import Configuration
+
+
+
+class SDRClassifierFactory(object):
+  """Factory for instantiating SDR classifiers."""
+
+
+  @staticmethod
+  def create(*args, **kwargs):
```
oxtopus: In the absence of well defined args, we should have a complete docstring with details of what is expected and required.

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,500 @@
...
+
+      ),
+      commands=dict()
+    )
+
+    return ns
+
+
+  def __init__(self,
+               steps='1',
```
oxtopus: `str` for number of steps?  Why not `steps=1`?

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,500 @@
...
+               ):
+
+    # Set default implementation
+    if implementation is None:
+      implementation = Configuration.get('nupic.opf.sdrClassifier.implementation')
+
+    # Convert the steps designation to a list
+    self.classifierImp = implementation
+    self.steps = steps
+    self.stepsList = eval("[%s]" % (steps))
```
oxtopus: I don't understand why steps is a `str` in the first place, but previous discussions have resulted in standardizing on `str.format()` for substitution rather than `%`.

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,500 @@
...
+
+    # Flag to know if the compute() function is ever called. This is to 
+    # prevent backward compatibilities issues with the customCompute() method
+    # being called at the same time as the compute() method. Only compute() 
+    # should be called via network.run(). This flag will be removed once we 
+    # get to cleaning up the clamodel.py file.
+    self._computeFlag = False
+
+
+  def _initEphemerals(self):
```
oxtopus: What is this function, and why is it implemented as `pass`.

tests/unit/nupic/algorithms/sdr_classifier_test.py
```diff
@@ -24,13 +24,21 @@
...
 import unittest2 as unittest
 
 import numpy
 
 from nupic.algorithms.sdr_classifier import SDRClassifier
 
+try:
+  import capnp
+except ImportError:
+  capnp = None
```
oxtopus: Repeated pattern as before.  Another reason this should be implemented in a higher level api.

oxtopus:
LGTM aside from a few minor questions.

src/nupic/regions/SDRClassifierRegion.py
scottpurdy: Once we have it working for Windows we can remove it altogether. Open to suggestions in the interim but let's not hold up PRs like this on it.

src/nupic/algorithms/sdr_classifier.py
ywcui1990: I couldn't do classifier.steps = proto.steps because proto.steps is of type "capnp list reader". So we have to unpack it here. I used list comparisons following your suggestions.

src/nupic/regions/SDRClassifierRegion.py
ywcui1990: This is to be consistent with the existing CLA classifier Region. I tried to keep the API same for SDRClassifier and CLAClassifier

src/nupic/regions/SDRClassifierRegion.py
ywcui1990: I added a doc string here. This seems to be a common function that exists in all Regions. The function is supposed to "Initialize ephemeral attributes that are not saved with the checkpoint. These ephemeral attributes may be used by derived classes." 

I don't think there are any ephemeral attributes in SDRClassifierRegion, and implement it as `pass`

src/nupic/regions/SDRClassifierRegion.py
oxtopus: Why is CLAClassifierRegion like that?  Does anyone know?  The underlying CLAClassifier.py implementation takes a sequence of `int`, why not the region?

subutai: > Why is CLAClassifierRegion like that? Does anyone know? The underlying CLAClassifier.py implementation takes a sequence of int, why not the region?

I think it is because you can't pass in a list of ints to a constructor through the Network API, it has to be a string.

oxtopus: We should make a concerted effort to hide that implementation detail -- see https://en.wikipedia.org/wiki/Leaky_abstraction for details on why.  I'd be happy to take that on separately (don't hold this PR on resolution).

scottpurdy: @oxtopus I think that's a great idea to look into more. Just keep in mind that this class is a wrapper around an algorithm that is exclusively here to provide a network-api-interface to the algorithm. So requiring strings like this seems fine to me as long as we parse into a list of ints before passing the the algorithm implementaiton.

src/nupic/regions/SDRClassifierRegion.py
scottpurdy: This can just be `self.stepsList = steps.split(",")`

ywcui1990: `steps.split(",")` gives a list of strings (e.g., `["1", "2"]`). We need a list of integers here.

scottpurdy: `[int(i) for i in steps.split(",")]`

Don't use `eval`.

ywcui1990:
@scottpurdy I have replaced "eval" with string splits. Please merge if there are no further comments.

subutai:
@ywcui1990 I have some comments but have not had a chance to do it yet. So please wait to merge until I've reviewed it.  Thanks.

src/nupic/algorithms/sdr_classifier_factory.py
```diff
@@ -0,0 +1,69 @@
...
+  @staticmethod
+  def read(proto):
+    """
+    proto: SDRClassifierRegionProto capnproto object
+    """
+    impl = proto.classifierImp
+    if impl == 'py':
+      return SDRClassifier.read(proto.claClassifier)
+    elif impl == 'cpp':
+      raise NotImplementedError("FastSDRClassifier.read not implemented")
```
subutai: The error message is not correct - there is no C++ version of SDR Classifier.

src/nupic/algorithms/sdr_classifier_factory.py
```diff
@@ -0,0 +1,69 @@
...
+    elif impl == 'diff':
```
subutai: What is the 'diff' implementation? Remove these lines.

src/nupic/algorithms/sdr_classifier_factory.py
```diff
@@ -0,0 +1,69 @@
...
+    impl = kwargs.pop('implementation', None)
+    if impl is None:
+      impl = Configuration.get('nupic.opf.sdrClassifier.implementation')
+    if impl == 'py':
+      return SDRClassifier(*args, **kwargs)
+    elif impl == 'cpp':
+      raise NotImplementedError("FastSDRClassifier not implemented")
+    else:
+      raise ValueError('Invalid classifier implementation (%r). Value must be '
+                       '"py" or "cpp".' % impl)
```
subutai: The error message is not correct - there is no C++ version of SDR Classifier. As such, there is only one implementation option.

src/nupic/algorithms/sdr_classifier_factory.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    impl = proto.classifierImp
+    if impl == 'py':
+      return SDRClassifier.read(proto.claClassifier)
+    elif impl == 'cpp':
+      raise NotImplementedError("FastSDRClassifier.read not implemented")
+    elif impl == 'diff':
+      raise NotImplementedError("SDRClassifierDiff.read not implemented")
+    else:
+      raise ValueError('Invalid classifier implementation (%r). Value must be '
```
subutai: The error message is not correct - there is no C++ version of SDR Classifier. As such, there is only one implementation option.

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,502 @@
...
+          constraints='',
+          defaultValue=0.001,
+          accessMode='Create'),
+
+        implementation=dict(
+          description='The classifier implementation to use.',
+          accessMode='ReadWrite',
+          dataType='Byte',
+          count=0,
+          constraints='enum: py, cpp'),
```
subutai: The error message is not correct - there is no C++ version of SDR Classifier. As such, there is only one implementation option.

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,502 @@
...
+
+    # Convert the steps designation to a list
+    self.classifierImp = implementation
+    self.steps = steps
+    self.stepsList = [int(i) for i in steps.split(",")]
+    self.alpha = alpha
+    self.verbosity = clVerbosity
+
+    # Initialize internal structures
+    self._sdrClassifier = SDRClassifierFactory.create(
```
subutai: This creation should be in the initialize() method.

src/nupic/regions/SDRClassifierRegion.py
subutai: This _initEphemerals() method is not required in this Region. Please remove.

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,502 @@
...
+  def clear(self):
+    self._sdrClassifier.clear()
+
+
+  def getAlgorithmInstance(self):
+    """Returns instance of the underlying CLAClassifier algorithm object."""
+    return self._sdrClassifier
+
+
+  def getParameter(self, name, index=-1):
```
subutai: This implementation of getParameter will not work for all parameters. For example, "implementation" cannot be retrieved, since the member variable has a different name.

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,502 @@
...
+    instance.alpha = proto.alpha
+    instance.verbosity = proto.verbosity
+    instance.maxCategoryCount = proto.maxCategoryCount
+
+    instance._sdrClassifier = SDRClassifierFactory.read(proto)
+
+    return instance
+
+
+  def reset(self):
```
subutai: This method is not required is it?

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,502 @@
...
+    self._computeFlag = True
+
+    # An input can potentially belong to multiple categories. 
+    # If a category value is < 0, it means that the input does not belong to 
+    # that category.
+    categories = [category for category in inputs["categoryIn"]
+                  if category >= 0]
+
+    activeCells = inputs["bottomUpIn"]
+    patternNZ = activeCells.nonzero()[0]
```
subutai: Why declare a variable just to be used once? Why not just have `patternNZ = inputs["bottomUpIn"].nonzero()[0]`

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,502 @@
...
+    # ==========================================================================
+    # Allow to train on multiple input categories. 
+    # Do inference first, and then train on all input categories.
+
+    # --------------------------------------------------------------------------
+    #   1. Call classifier. Don't train. Just inference. Train after.
+
+    # Dummy classification input, because this param is required. Learning is 
+    # off, so the classifier is not learning this input. Inference only here.
+    classificationIn = {"actValue": 0, "bucketIdx": 0}
```
subutai: I'm confused as to why we have so many calls to compute.

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,502 @@
...
+
+    for category in categories:
+      classificationIn = {"bucketIdx": int(category), "actValue": int(category)}
+
+      # ------------------------------------------------------------------------
+      #   2. Train classifier, no inference
+      self._sdrClassifier.compute(recordNum=self.recordNum,
+                                  patternNZ=patternNZ,
+                                  classification=classificationIn,
+                                  learn=self.learningMode,
```
subutai: What happens if self.learningMode is False? Is it still correct to call compute here?

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,502 @@
...
+                                  infer=False)
+
+    outputs['actualValues'] = clResults["actualValues"]
```
subutai: Requires a copy, i.e. outputs['actualValues'][:] = clResults["actualValues"]

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,502 @@
...
+                    DeprecationWarning)
+
+    return self._sdrClassifier.compute(recordNum,
+                                       patternNZ,
+                                       classification,
+                                       self.learningMode,
+                                       self.inferenceMode)
+
+
+  def getOutputValues(self, outputName):
```
subutai: This method is incorrect and will cause an exception. I don't believe we need this method at all.

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,502 @@
...
+    """
+    Return the dictionary of output values. Note that these are normal Python
+    lists, rather than numpy arrays. This is to support lists with mixed scalars
+    and strings, as in the case of records with categorical variables
+    """
+    return self._outputValues[outputName]
+
+
+  def getOutputElementCount(self, outputName):
+    """Returns the width of dataOut."""
```
subutai: This comment is incorrect.

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,502 @@
...
+
+        maxCategoryCount=dict(
+          description='The maximal number of categories the '
+                      'classifier will distinguish between.',
+          dataType='UInt32',
+          required=True,
+          count=1,
+          constraints='',
+          # arbitrarily large value for backward compatibility 
+          defaultValue=1000,
```
subutai: Why is this default 1000? What backward compatibility do we need to satisfy for SDRClassiferRegion? It's a brand new Region.

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,502 @@
...
+    and strings, as in the case of records with categorical variables
+    """
+    return self._outputValues[outputName]
+
+
+  def getOutputElementCount(self, outputName):
+    """Returns the width of dataOut."""
+
+    # Check if classifier has a 'maxCategoryCount' attribute
+    if not hasattr(self, "maxCategoryCount"):
```
subutai: This if statement is not required - this region will always have maxCategoryCount because it is initialized in the constructor.

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,502 @@
...
+    for step in self.stepsList:
+      stepIndex = self.stepsList.index(step)
+      categoryOut = clResults["actualValues"][clResults[step].argmax()]
+      outputs['categoriesOut'][stepIndex] = categoryOut
+
+      # Flatten the rest of the output. For example:
+      #   Original dict  {1 : [0.1, 0.3, 0.2, 0.7]
+      #                   4 : [0.2, 0.4, 0.3, 0.5]}
+      #   becomes: [0.1, 0.3, 0.2, 0.7, 0.2, 0.4, 0.3, 0.5] 
+      stepProbabilities = clResults[step]
```
subutai: How do we know this works in all cases? There are no tests of this.

subutai:
@ywcui1990 I don't see any tests of the region that demonstrate you can create the network, get and set parameters, and get expected values in various situations. I noticed at least one issue in retrieving parameters, but there may be many others.

subutai:
@ywcui1990 We also need an end-to-end integration test that runs the classifier on some dataset with the OPF and verifies you get expected results.

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,502 @@
...
+  def initialize(self, dims, splitterMaps):
+    pass
+
+
+  def clear(self):
+    self._sdrClassifier.clear()
+
+
+  def getAlgorithmInstance(self):
+    """Returns instance of the underlying CLAClassifier algorithm object."""
```
subutai: Comment is incorrect

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,502 @@
...
+    These ephemeral attributes may be used by derived classes.
+    """
+    pass
+
+
+  def initialize(self, dims, splitterMaps):
+    pass
+
+
+  def clear(self):
```
subutai: When is this method used? Is it necessary?

src/nupic/regions/SDRClassifierRegion.py
ywcui1990: It seems that this just needs to be a large enough value. It is arbitrarily set to be 1000

src/nupic/regions/SDRClassifierRegion.py
ywcui1990: outputs['actualValues'] may have different length than clResults["actualValues"], I used `outputs['actualValues'] = copy.copy(clResults["actualValues"])` instead

src/nupic/regions/SDRClassifierRegion.py
ywcui1990: Inference and learning is separated here to allow training with multiple input categories. The first call of compute is for inference, and the rests are for learning

src/nupic/regions/SDRClassifierRegion.py
ywcui1990: I am not sure what this is for. I have removed it

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,504 @@
...
+    elif outputName == "probabilities":
+      return len(self.stepsList) * self.maxCategoryCount
+    elif outputName == "actualValues":
+      return self.maxCategoryCount
+    else:
+      raise ValueError("Unknown output {}.".format(outputName))
+
+
+
+if __name__ == "__main__":
```
subutai: We don't need this "if" block and any of these lines. The file won't be run in a standalone manner and we have  tests elsewhere.

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,504 @@
...
+          accessMode='Create'),
+
+        implementation=dict(
+          description='The classifier implementation to use.',
+          accessMode='ReadWrite',
+          dataType='Byte',
+          count=0,
+          constraints='enum: py'),
+
+        clVerbosity=dict(
```
subutai: Why not just call this parameter `verbosity`?

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -0,0 +1,504 @@
...
+  def getParameter(self, name, index=-1):
+    """
+    Get the value of the parameter.
+
+    @param name -- the name of the parameter to retrieve, as defined
+            by the Node Spec.
+    """
+    # If any spec parameter name is the same as an attribute, this call
+    # will get it automatically, e.g. self.learningMode
+    if name == "clVerbosity":
```
subutai: Can get rid of this if we just call the parameter `verbosity`

ywcui1990:
@subutai I have fixed your most recent comments and the previously unaddressed comments

tests/integration/nupic/regions/single_step_sdr_classifier_test.py
```diff
@@ -0,0 +1,135 @@
...
+                     "Learning mode is not turned off.")
+    self.assertEqual(classifier.getParameter("inferenceMode"), 1,
+                     "Inference mode is not turned on.")
+
+    # make sure we can access all the parameters with getParameter
+    self.assertEqual(classifier.getParameter("maxCategoryCount"), 100)
+    self.assertAlmostEqual(float(classifier.getParameter("alpha")), 0.001)
+    self.assertEqual(int(classifier.getParameter("steps")), 0)
+    self.assertTrue(classifier.getParameter("implementation") == "py")
+    self.assertFalse(classifier.getParameter("verbosity"))
```
subutai: verbosity is an integer, so this should check that the returned value is 0

subutai:
@ywcui1990 Left one minor comment, otherwise 👍

PR #3104, opened 2016-04-23
Get the C++ TM running in the integration tests

### LINKED ISSUE #3103: The CPP TM isn't running in the integration tests

mrcslws:
Fixes #3103

Main changes:

- Stop using so many static class variables.
- Fix filenames so that they end with "test.py" so that pytest runs them.
- Make TM test classes abstract. Use a derived class to specify the TM class
  and to inherit from unittest.TestCase.

This change backs away from using static class variables. They do bad things when the class is used by multiple tests in a single pytest run. For example, before this change the command

~~~
py.test extensive_temporal_memory_test.py
~~~

succeeds, but

~~~
py.test extensive_temporal_memory_test.py extensive_temporal_memory_test.py
~~~

fails.


rhyolight:
Nice catch.

mrcslws:
@scottpurdy wanna review? (sorry, I forgot to tag you)

tests/integration/nupic/algorithms/extensive_tm_test_base.py
```diff
@@ -188,22 +189,26 @@ class ExtensiveTemporalMemoryTest(AbstractTemporalMemoryTest):
...
-    "seed": 42
-  }
-  PATTERN_MACHINE = PatternMachine(100, range(21, 26), num=300)
 
+  def getPatternMachine(self):
+    return PatternMachine(100, range(21, 26), num=300)
+
+  def getDefaultTMParams(self):
+    return {
+      "columnDimensions": [100],
```
scottpurdy: Pass a tuple instead of list?

tests/integration/nupic/algorithms/tutorial_temporal_memory_test.py
```diff
@@ -22,29 +22,36 @@
...
-    "activationThreshold": 1
-  }
-  PATTERN_MACHINE = ConsecutivePatternMachine(6, 1)
 
+  def getPatternMachine(self):
+    return ConsecutivePatternMachine(6, 1)
+
+  def getDefaultTMParams(self):
+    return {
+      "columnDimensions": [6],
```
scottpurdy: tuple? (same as above)

tests/integration/nupic/algorithms/extensive_tm_test_base.py
```diff
@@ -188,22 +189,26 @@ class ExtensiveTemporalMemoryTest(AbstractTemporalMemoryTest):
...
+      "cellsPerColumn": 1,
+      "initialPermanence": 0.8,
+      "connectedPermanence": 0.7,
+      "minThreshold": 11,
+      "maxNewSynapseCount": 11,
+      "permanenceIncrement": 0.4,
+      "permanenceDecrement": 0,
+      "activationThreshold": 11,
+      "seed": 42
```
scottpurdy: I guess personal preference but I like putting trailing comma on last item in list/dict like this so when you add or remove lines you don't have to maintain the last line differently.

scottpurdy:
Done reviewing.

rhyolight:
@scottpurdy doesn't like it when I merge before he gives a final 👍  so I'll wait for him to get back to the office.

src/nupic/test/abstract_temporal_memory_test.py
```diff
@@ -20,26 +20,41 @@
...
 
 from nupic.data.generators.sequence_machine import SequenceMachine
 from nupic.research.monitor_mixin.temporal_memory_monitor_mixin import (
   TemporalMemoryMonitorMixin)
-from nupic.research.temporal_memory import TemporalMemory
 
 
 
-class AbstractTemporalMemoryTest(unittest.TestCase):
+class AbstractTemporalMemoryTest(object):
```
vitaly-krugl: @mrcslws @scottpurdy 

This change is problematic for those of us who editors/IDEs with tab-completion. After this change, tab-completion on `unittest.TestCase` methods in classes like `TutorialTemporalMemoryTest` will no longer work. Perhaps renaming the files to replace the word "test" in the filename with something else or whatever it is that py.test is looking for (can be configured) would be a better option.

mrcslws: @vitaly-krugl Why won't tab-completion work? `TutorialTemporalMemoryTest` inherits from `unittest.TestCase` via multiple inheritance. IDEs can handle that, right?

src/nupic/test/abstract_temporal_memory_test.py
```diff
@@ -20,26 +20,41 @@
 # ----------------------------------------------------------------------
 
 import unittest
```
scottpurdy: No longer need this import in this file, right?

mrcslws: Good point. It's removed now.

PR #3111, opened 2016-04-27
Stop relying on a nupic.core ASSERT to check parameters

mrcslws:
Fixes https://github.com/numenta/nupic/issues/3110

nupic.core NTA_ASSERTs are no longer in Release builds.

An alternate fix would be to convert these NTA_ASSERTs to NTA_CHECKs. But they're potentially expensive checks. The nupic.core math code was written to become faster with NTA_ASSERTIONS_ON disabled, and it's best to keep it that way.

Since the unit tests expect this param checking, implement it on the Python side.

### DIFF
--- a/src/nupic/algorithms/KNNClassifier.py
+++ b/src/nupic/algorithms/KNNClassifier.py
@@ -407,14 +407,12 @@ class KNNClassifier(object):
                                               cellsPerCol=self.cellsPerCol)
 
     if isSparse > 0:
-      isSorted = all(inputPattern[i] <= inputPattern[i+1]
-                     for i in xrange(len(inputPattern)-1))
-      if not isSorted:
-        raise RuntimeError("Sparse inputPattern must be sorted.")
-
-      if not all(bit < isSparse for bit in inputPattern):
-        raise RuntimeError("Sparse inputPattern contains an index outside of"
-                           "the dense representation's bounds.")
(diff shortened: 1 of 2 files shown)


src/nupic/algorithms/KNNClassifier.py
```diff
@@ -406,6 +406,16 @@ def learn(self, inputPattern, inputCategory, partitionId=None, isSparse=0,
       print "  active inputs:", _labeledInput(inputPattern,
                                               cellsPerCol=self.cellsPerCol)
 
+    if isSparse > 0:
+      isSorted = all(inputPattern[i] <= inputPattern[i+1]
+                     for i in xrange(len(inputPattern)-1))
+      if not isSorted:
+        raise RuntimeError("Sparse inputPattern must be sorted.")
```
scottpurdy: Pick more specific exception types for these cases. Not sure what would be best for this one (`ValueError`?) but the next one could be `IndexError`.

scottpurdy:
This change doesn't seem to directly test the issue. I think we just need to be really explicit about how we handle parameter checking. If the tests don't match the desired behavior then we should update them. If we want to only check parameters in debug builds, that's fine too. I don't have a strong opinion.

mrcslws:
@scottpurdy All green.

scottpurdy:
👍

PR #3126, opened 2016-05-11
Fix default values of DATETIME columns in ClientJobsDAO to be compatible with newer as well as older versions of mysql

### LINKED ISSUE #2983: MySQL version 5.7 issue with ClientJobsDAO.py DATETIME DEFAULT 0
Just installed nupic on OSX 10.9. 
Found MySQL version 5.7 (latest) didn't like like the DATETIME DEFAULT 0 usage in 
python/site-packages/nupic/database/ClientJobsDAO.py
Solution used: changed all instances of
DATETIME DEFAULT 0
to
DATETIME DEFAULT CURRENT_TIMESTAMP
... and MySQL was happy again.
Hope this is useful.
Andy

vitaly-krugl:
Fixes #2983
Default DATETIME columns to NULL in ClientJobsDAO for compatibility across mysql versions.

As of mysql 5.7.8, values of 0 are not allowed for DATETIME columns, and CURRENT_TIMESTAMP
is semantically inappropriate for those columns.

Since the old default of 0 ("0000-00-00 00:00:00") is an invalid value for `datetime.datetime` constructor, pymysql returns None for columns containing that invalid date, which is compatible with the new default DATETIME value of NULL that I used in the fix.

CC @scottpurdy


rhyolight:
Thank you @vitaly-krugl !

scottpurdy:
👍

PR #3131, opened 2016-05-18
Sp profile review

breznak:
* Fix SP profiling to work with updated SP syntax
* run benchmarks on SP 1D vs 2D topology speed
* confirms issue https://github.com/numenta/nupic.core/issues/380#issuecomment-200074408 
* adds benchmark for TM (using TP_shim)
* benchmarks of TM/TP

Addresses: https://github.com/numenta/nupic.core/issues/380
Fixes: #3142

### DIFF
--- a/scripts/profiling/sp_profile.py
+++ b/scripts/profiling/sp_profile.py
@@ -21,9 +21,16 @@
 # ----------------------------------------------------------------------
 
 ## run python -m cProfile --sort cumtime $NUPIC/scripts/profiling/sp_profile.py [nColumns nEpochs]
+# example results (not averaged, on local machine):
+# input size N=10000, columns=2048, epochs=150 
+# PySP | 1D (colDim=(2048,))   | 0.441 s/call (compute)
+# PySP | 2D (colDim=(2048,1))  | 0.295 s/call (compute)
+# CppSP| 1D		       | 0.108 s/call (algorithms.py/compute)
+# CppSP| 2D		       | 0.040 s/call --- 2.5x FASTER! should be same(or worse) as 1D, FIXME
 
 import sys
 import numpy
+import itertools
 # chose desired SP implementation to compare:
 from nupic.research.spatial_pooler import SpatialPooler as PySP
 from nupic.bindings.algorithms import SpatialPooler as CppSP
@@ -40,8 +47,8 @@ def profileSP(spClass, spDim, nRuns):
   @param nRuns number of calls of the profiled code (epochs)
   """
   # you can change dimensionality here, eg to 2D
-  inDim = [10000, 1, 1]
-  colDim = [spDim, 1, 1]
+  inDim = (10000,1) # a rather large input
+  colDim = (spDim,1) # 1D vs 2D just toggle here (spDim,) vs (spDim,1), and do the same change in inDim ^^^
 
 
   # create SP instance to measure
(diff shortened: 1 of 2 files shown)


breznak:
@rhyolight can you review and merge?

mrcslws:
Interesting findings. This tp_profile script is testing a very nonrealistic special case, but you're right that the old TP is faster in this case.

This script:

- Uses random inputs
- Uses ~50% sparsity

So if there are 2048 columns, ~1024 of them are active. So, 1024 different segments have to grow / reinforce synapses every timestep. The inputs are random, so none of the synapses are useful. It's hard to predict the exact outcome, since the 50% sparsity might cause tons of false positives, so segments might actually activate quite often despite the random input.

One reason the old TP is faster in this case is because its synapse cleanup code is better. This experiment stress-tests the heck out of synapse cleanup. The flaw with the TM's code (it's actually buggy, not just slow) is that its synapse cleanup deletes one synapse at a time, rather than doing them in groups like the TP's `freeNSynapses` method. This is slower, and doing it one at a time is wrong because if you create 10 synapses and it triggers one-at-a-time cleanup, you might effectively only add 1 synapse, repeatedly cleaning up the just-added synapse. This is a known issue. I'm planning on making segment/synapse cleanup better, you can see [I've been studying for it](http://mrcslws.com/blocks/2016/05/05/tp-segment-cleanup.html).

One reason the TM/TP become slower than the SP in this case is that the SP is more capable of handling non-sparse inputs. Since it uses connection matrices rather than lists of synapses, the change in sparsity doesn't have a dramatic effect on the overlap calculation.

Usually when we talk about the performance of the algorithms, we use non-random inputs. And since you're feeding patterns directly into the TM, you need to think about sparsity.

breznak:
@mrcslws thank you for your review and for the pointer to your blog! 

> Uses ~50% sparsity

good point, I will adapt to use 2% sparsity, to track the common behavior. Is there available "the code from NAB" that let's one to obtain the "best" params for SP/TM? 

> Uses random inputs

I'm aware of this, consider this like the "worst case scenarion". Is there a better way to mimick "notmal: data? That is overlaps of inputs are more common.

subutai:
From another PR: it looks like the SP parameters are quite off in sp_profile. Why is potentialRadius only 3? It should be much larger to form good SDRs. Same with numActiveColumnsPerInhArea, etc. etc. Why is epochs set to 150 here? 

I think the parameters should be set to realistic numbers and the profile re-run with those numbers.

breznak:
> the SP parameters are quite off ...

thanks @subutai , Marcus explained that, I agree...and don't know where I've taken the params from - from some example. I think it shows 2 another optimization problems: "the time metric" would make sense; and "optimal default params" - do we agree "hotgym" is a source of those opt. params? I'll make another PR for the latter and update SP/TM/TP in nupic/nupic.core constructors. 

I'll update the optimization issue with the updated benchmarks.

subutai:
> do we agree "hotgym" is a source of those opt. params?

Well, it looks like sp_profile is trying to benchmark SP with local inhibition and 2D topology.  If that is the goal, you can't use hotgym parameters, since hotgym uses global inhibition. You need to have parameters that make sense for the 2D toplogy / local inhibition scenario.

breznak:
> Well, it looks like sp_profile is trying to benchmark SP with local inhibition and 2D topology 

It is used to benchmark SP on 1D vs "2D" data (the 2D is fake, it's just {2048} vs {2048,1}) - please see motivation for that: numenta/nupic.core#380

So I'm looking for best params for 1D too.

mrcslws:
For random inputs vs non-random, I'd say it's ambiguous. There are lots of experiments you can run, and it's unclear which is the right one. Maybe it's best to have a distribution of tests, so the result isn't just a single time, but a table of times. So far my perf tests have been hotgym and this script: https://github.com/numenta/nupic/blob/master/scripts/temporal_memory_performance_benchmark.py . (It's definitely not a full perf test suite.)

One problem with random inputs is that it actually doesn't cover all the code. With sparse column activations, random inputs will never activate or even match a segment. You can see this in my 2nd, 3rd, and 4th diagrams in [this blog post](http://mrcslws.com/blocks/2016/04/28/life-and-times-of-dendrite-segment.html). When it starts doing random input, the segments get zero excitation. So it never reinforces or punishes a segment. So although you're right that random is generally the worst case, it's still not a holistic test. It will completely overlook the performance of segment reinforcement / punishment.

breznak:
@rhyolight can this PR be reopened and merged just as is, or with minor modifications? It fixes the profile scripts (and adds TM shim to benchmark). 

From the discussion we conclude that those scripts still benchmark only some "type" of the code/code paths; so e.g. hotgym is still a relevant and likely better benchmark.

rhyolight:
@breznak Per our process, we can't merge this until @numenta/nupic-reviewers gives it a thumbs-up. I don't know this code well enough to make that call.

Issue #2616, opened 2015-09-29
Py and C++ classes use same init params


breznak:
EDIT: This issue is to enforce that py and C++ use the same init params. Doing otherwise is against the principle that those 2 implementations are identical. This is fixing (unfortunately not, but cornering) a real problem where SDR representations differ in SP in Py/C++ implementations. 

- [ ] blocked by: https://github.com/numenta/nupic.core/issues/628

rhyolight:
Needs discussion from other @numenta/nupic-committers.

rhyolight:
This would be a whole lot easier to digest if there were separate issues for the SP and the TM, each with different PRs that change default param values. I don't like them mixed together, because they are completely different algorithms and subjects.

breznak:
> This would be a whole lot easier to digest if there were separate issues for the SP and the TM 

agreed. closing this issue then.

PR #3143, opened 2016-05-25
Optimal default params

breznak:
- uses optimal generic params from `common_models/`
- avoids broken/meaningless combinations of default params

Fixes: #2616 

Blocked until same change is done in nupic.core


breznak:
Please review why the Py unittest is failing

PR #3158, opened 2016-06-07
Added examples illustrating overlap properties of Spatial Pooling

### LINKED ISSUE #3149: Add examples illustrating overlap properties of Spatial Pooling
This task is to add an example to examples/sp that illustrates various overlap properties of Spatial Pooling. In particular we want easy to understand script(s) to replicate Figures 5, 7, and 9 in this paper:

http://arxiv.org/abs/1505.02142

The script should produce image files (using matplotlib), one for each graph. It should be clearly commented and documented so that a new user can understand what is going on.

Should also update the README in that directory accordingly.

hernandezurbina:
Fixes #3149


subutai:
@hernandezurbina Looks like there is a test failure in spatial_pooler_cpp_api_test.py. It might be related to the SP changes made by @mrcslws yesterday. Can you make sure you get and rebuild the latest nupic.core and nupic, and then run the tests locally to verify?

hernandezurbina:
@subutai I moved back the overlap test in the SP back to the original test file due to some issues with the cpp version of the SP. As discussed with @mrcslws this is the best temporary solution. Other than that, no changes to the sp_tutorial file nor its README file.

mrcslws:
@subutai FYI the issue with the API test is tracked here, it's pretty gross: https://github.com/numenta/nupic/issues/3159

subutai:
@mrcslws Wow, tricky...  Thanks.

examples/sp/sp_tutorial.py
```diff
@@ -0,0 +1,312 @@
...
+import random
+import matplotlib.pyplot as plt
+from nupic.research.spatial_pooler import SpatialPooler as SP
+
+def pctOverlap(x1, x2, size):
+  """
+  Computes the percentage of overlap between vectors x1 and x2.
+  Parameters:
+  ----------
+  x1: binary vector whose size is specified in parameter size
```
subutai: We use `@param` to document parameters (picked up by automated documentation tools). See nupic/src/nupic/research/temporal_memory.py for an example.

examples/sp/sp_tutorial.py
```diff
@@ -0,0 +1,312 @@
...
+# ----------------------------------------------------------------------
+
+"""A simple tutorial that shows some features of the Spatial Pooler."""
+
+import numpy as np
+import random
+import matplotlib.pyplot as plt
+from nupic.research.spatial_pooler import SpatialPooler as SP
+
+def pctOverlap(x1, x2, size):
```
subutai: Why not just call this "percentOverlap"?

examples/sp/sp_tutorial.py
```diff
@@ -0,0 +1,312 @@
...
+  if minX1X2 > 0:
+    pctOverlap = float(np.dot(x1, x2))/float(minX1X2)
+  return pctOverlap
+
+
+
+def corruptVector(vector, noiseLevel):
+  """
+  Corrupts a binary vector by swaping its inner values depending
+  on the amount of noise specified by the user.
```
subutai: May be clearer to write "Corrupts a binary vector by inverting noiseLevel percent of its bits."

examples/sp/sp_tutorial.py
```diff
@@ -0,0 +1,312 @@
...
+  numActiveColumnsPerInhArea = int(0.02*columnNumber),
+  globalInhibition = True,
+  seed = 1,
+  synPermActiveInc = 0.01,
+  synPermInactiveDec = 0.008
+   )
+
+# Part 1: 
+# -------
+# A column connects to a region of the input vector (specified
```
subutai: "connects to a subset of the"

examples/sp/sp_tutorial.py
```diff
@@ -0,0 +1,312 @@
...
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+"""A simple tutorial that shows some features of the Spatial Pooler."""
+
+import numpy as np
+import random
+import matplotlib.pyplot as plt
```
subutai: We need to use matplotlib in non-interactive mode so this script can work on linux, automated scripts, etc. To do that put `matplotlib.use('Agg')` here and then use savefig after show() (see below).

examples/sp/sp_tutorial.py
```diff
@@ -0,0 +1,312 @@
...
+print ""
+
+bins = np.linspace(min(overlaps), max(overlaps), 50)
+plt.hist(overlaps, bins, alpha=0.5, label='All cols')
+plt.hist(activeColsScores, bins, alpha=0.5, label='Active cols')
+plt.legend(loc='upper right')
+plt.xlabel("Overlap scores")
+plt.ylabel("Frequency")
+plt.title("Figure 1")
+plt.show()
```
subutai: then put in:
`plt.savefig("figure_1")`

examples/sp/sp_tutorial.py
```diff
@@ -0,0 +1,312 @@
...
+
+# Part 2a: 
+# -------
+# The input overlap between two binary vectors is defined as their dot product. In order
+# to normalize this value we divide it by the number of minimum number of active inputs
```
subutai: "divide by the minimum number of active inputs"

examples/sp/sp_tutorial.py
```diff
@@ -0,0 +1,312 @@
...
+# (in either vector). This means we are considering the sparser vector as reference.
+# Two identical binary vectors will have an input overlap of 1, whereas two complete
```
subutai: "whereas two completely"

examples/sp/sp_tutorial.py
```diff
@@ -0,0 +1,312 @@
...
+
+print ""
+print "---------------------------------"
+print "Figure 3 shows the output overlap between two sparse representations" 
+print "in function of their input overlap. Starting from two identical binary vectors"
+print "(which yield the same active columns) we add noise two one of them"
+print "feed it to the SP, and estimate the output overlap between the two"
+print "representations in terms of the common active columns between them."
+print "As expected, as the input overlap decrease, so does the output overlap."
+print "However, this relationship is sublinear, implying that the SP is very"
```
subutai: I think I would just take out this statement about "very sensitive to noise". Because of the SDR math, even an overlap of 10 is very significant, so even with the fast drop the SP is somewhat tolerant to noise".

examples/sp/sp_tutorial.py
```diff
@@ -0,0 +1,312 @@
...
+
+plt.plot(x, y)
+plt.xlabel("Input overlap")
+plt.ylabel("Output overlap")
+plt.title("Figure 3")
+plt.show()
+
+# Part 3: 
+# -------
+# As we saw in the last section, an untrained SP is very sensitive to noise in a binary vector
```
subutai: Similar to my earlier comment, I would take out the first sentence. Focus on the fact that with training, the SP can become less sensitive to noise.

examples/sp/sp_tutorial.py
```diff
@@ -0,0 +1,312 @@
...
+# in such a way that the output overlap between a vector and a corrupted version of it decreases
+# dramatically as the amount of noise grows. To improve this situation we train the SP by
+# turning learning on, and by exposing it to a variety of random binary vectors.
+# We will expose the SP to a repetition of input patterns in order to make it learn and distinguish
+# them once learning is over. This will result in robustness to noise in the inputs.
+# In this section we will reproduce the plot in the last section after the SP has learned a series
+# of inputs. Here we will see how the SP exhibits some resilience to noise after learning.
```
subutai: "exhibits increased resilience"

examples/sp/sp_tutorial.py
```diff
@@ -0,0 +1,312 @@
...
+columnNumber = np.array(columnDimensions).prod()
+inputArray = np.zeros(inputSize, dtype=uintType)
+
+for i in range(inputSize):
+  inputArray[i] = random.randrange(2)
+  
+activeCols = np.zeros(columnNumber, dtype=uintType)
+sp = SP(inputDimensions, 
+  columnDimensions, 
+  potentialRadius = int(0.5*inputSize),
```
subutai: Why is this not inputSize?

examples/sp/sp_tutorial.py
```diff
@@ -0,0 +1,312 @@
...
+# turning learning on, and by exposing it to a variety of random binary vectors.
+# We will expose the SP to a repetition of input patterns in order to make it learn and distinguish
+# them once learning is over. This will result in robustness to noise in the inputs.
+# In this section we will reproduce the plot in the last section after the SP has learned a series
+# of inputs. Here we will see how the SP exhibits some resilience to noise after learning.
+
+# We will present 10 random vectors to the SP, and repeat this 30 times.
+# Try changing the number of times we do this to see how it changes the last plot.
+# Then, you could also modify the number of examples to see how the SP behaves.
+# Is there a relationship between the number of examples and the number of times that
```
subutai: Is this a question for the reader to answer?

examples/sp/sp_tutorial.py
```diff
@@ -0,0 +1,312 @@
...
+  y.append(pctOverlap(outputColumns[0][:], outputColumnsCorrupted[0][:], columnNumber))
+
+print ""
+print "---------------------------------"
+print "How robust is the SP to noise after learning?"
+print "Figure 4 shows again the output overlap between two binary vectors in function"
+print "of their input overlap. After training, the SP exhibits more robustness to noise"
+print "in its input, resulting in a -almost- sigmoid curve. This implies that even if a"
+print "previous input is presented again with a certain amount of noise its sparse"
+print "representation still resembles somehow its original version."
```
subutai: "still resembles its original"

subutai:
@hernandezurbina Thanks - this looks very clean and simple to follow. The comments and printouts will be helpful to a brand new user. I left some small edits/comments.

examples/sp/sp_tutorial.py
subutai: See http://matplotlib.org/faq/howto_faq.html#generate-images-without-having-a-window-appear

examples/sp/sp_tutorial.py
hernandezurbina: yes, I was thinking that the user could play around with those parameters in order to gain more insight about the SP. What do you think?

subutai: Sounds good - good idea.

hernandezurbina:
@subutai Ok, changes/corrections have been applied.

examples/sp/sp_tutorial.py
```diff
@@ -0,0 +1,316 @@
...
+
+print ""
+print "---------------------------------"
+print "Figure 1 shows an histogram of the overlap scores"
+print "from all the columns in the spatial pooler, as well as the"
+print "overlap scores of those columns that were selected to build a"
+print "sparse representation of the input (shown in green)."
+print "The SP chooses 2% of the columns with the largest overlap score"
+print "to make such sparse representation."
+print "---------------------------------"
```
subutai: @hernandezurbina  do you want to put a printout somewhere stating that the figures have been saved to files?

hernandezurbina: @subutai Yes, it is the very first printout. Lines 115-120. Do you think they are not visible? I could print it out as the very last message of the program.

subutai: Oh, I see. Yes, I think the last message might be better.

hernandezurbina: @subutai Ok, it's fixed now.

examples/sp/sp_tutorial.py
```diff
@@ -0,0 +1,317 @@
...
+  Copies the contents of vector x1 into vector x2.
+  
+  @param x1 (array) binary vector to be copied
+  @param x2 (array) binary vector where x1 is copied
+  """
+  size = len(x1)
+  for i in range(size):
+    x2[i] = x1[i]
+
+# The following program has the purpose of presenting some
```
scottpurdy: Should this information be moved into the module doc string?

subutai: Yes, I think it would be a good idea to move these comments to the top - add it to the intro string on line 23.

hernandezurbina:
Ok. Added the description to the doc string.

hernandezurbina:
@subutai In this PR: changed binning in histogram, added comments to figures and added comment in source file regarding part 2b.

subutai:
👍 Thanks!  I'll merge after tests pass.

PR #3166, opened 2016-06-15
Change verbose variable name for classifiers to 'verbosity'

### LINKED ISSUE #3165: Incompatibility clVerbosity / verbosity for classifiers
The parameter dict in classifiers (SDR and CLA) uses `clVerbosity` as parameter name for verbose level, and the actual parameter name is `verbosity`. This is fine in principle but the getter methods for Regions use this param dict as reference for valid parameter names, and thus are expecting `verbosity`.

Thus a stable fix would be to use verbosity instead of `clVerbosity` in those `param` dicts.

natoromano:
Fixes #3165 by changing the verbosity variable proto in SDRClassifierRegion.py. 

The parameter name for verbose level is now `clVerbosity`rather than `verbosity`, for consistency with the other classifiers and other regions (spatial pooler uses `spVerbosity` for instance).

### DIFF
--- a/examples/NuPIC Walkthrough.ipynb
+++ b/examples/NuPIC Walkthrough.ipynb
@@ -1,1548 +1,1904 @@
 {
- "metadata": {
-  "name": "",
-  "signature": "sha256:d5eb0b572ecbf560e545c3b61fd48aed11d949ba39fcff4392a5486c3fe1e8dd"
- },
- "nbformat": 3,
- "nbformat_minor": 0,
- "worksheets": [
+ "cells": [
   {
-   "cells": [
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Encoders\n",
+    "\n",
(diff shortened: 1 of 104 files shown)


natoromano:
@scottpurdy

scottpurdy:
@natoromano - I changed the PR description from "Fixed" -> "Fixes" to get one of the validators passing. It looks like you still need to complete the contributors license, which you can do digitally here:
http://numenta.org/licenses/cl/

And then ping me again when the CI systems (travis and AV) are passing.

natoromano:
Will actually change all to `verbosity`, cf #3165 for (slightly) more explanation.

rhyolight:
@scottpurdy ping

rhyolight:
Hey @natoromano this broke regression tests: https://discourse.numenta.org/t/broken-numenta-nupic-regression-840-master-8fc2690/816

PR #3197, opened 2016-07-06
Python column segment walk

### LINKED ISSUE #3116: From nupic.core [#917] Replace Phases w/columnSegmentWalk
(see: [nupic.core 917](https://github.com/numenta/nupic.core/pull/917))

This is a **HIGH PRIORITY** issue which needs follow up here, and also in HTM.java (see: [#421](https://github.com/numenta/htm.java/issues/421))

The TM algorithm implementation from nupic.core preserves the algorithm but radically changes the methodology to provide performance increases but will affect testing... Therefore, the carried-over changes must be vetted rigorously and tests should reflect the same determinant results because they are used as porting tools also...

andrewmalta13:
fixes #3116 

Implements the column segment walk described <a href="https://github.com/numenta/nupic.core/pull/917"> here </a>.

The implementations are very similar but not exact as there are differences in connections.py and connections.cpp which affect the implementation. (perhaps we can discuss if we eventually want these to agree to ensure complete alignment between the implementations).

Some of the differences that exist due to this discrepancy are:
1. Cpp version delete segments when its last synapse is destroyed and the python version does not.
2. When adding segments or synapses, if the number of created segments or synapses exceeds the parameters determining the maximum number of synapses on a segment or maximum number of segments on a cell:
  i) Cpp version deletes the least recently used segments and the minimum permanence synapses to make room for a new segment/synapse.
  ii) Python version ignores the parameters entirely (this does not seem right, but was in the old implementation) and creates them anyway without removing anything.
3. Less critical, but the data structures used to represent segments and cells are different. CPP has SegmentOverlap struct and uses a struct to represent a segment. Python version just uses an int to index into some dicts to represent a segment.  Cpp uses a dumb Cell struct wrapper that just contains an int, where python just represents a cell by an int. (effectively the same, but a bit annoying on the cpp side to always have to do something like ```int cellindex = cell.idx;```

Furthermore, this implementation does not seem to work with the old serialization methods as it does not preserve the numbering of segments.  If this is confirmed to be a result of the new algorithm, I will submit an github issue to address this, but as of right now I am not 100% certain it isn't an implementation bug.

Please Review:
@mrcslws


src/nupic/research/temporal_memory.py
```diff
@@ -123,416 +214,197 @@ def compute(self, activeColumns, learn=True):
...
-                           learningSegments,
-                           prevActiveCells,
-                           winnerCells,
-                           prevWinnerCells,
-                           self.connections,
-                           predictedInactiveCells,
-                           prevMatchingSegments)
+    """
+    prevActiveCells = sorted(self.activeCells)
+    prevWinnerCells = sorted(self.winnerCells)
```
mrcslws: Sorting these shouldn't be necessary. The cells / columns are walked in order, so these lists are always sorted.

andrewmalta13: 👍

src/nupic/research/temporal_memory.py
```diff
@@ -102,11 +195,9 @@ def __init__(self,
     self._random = Random(seed)
 
     self.activeCells = set()
-    self.predictiveCells = set()
-    self.activeSegments = set()
     self.winnerCells = set()
```
mrcslws: The `activeCells` and `winnerCells` can now be stored in a list or deque. We keep them sorted with the algorithm, so there doesn't have to be a runtime perf hit to keep them sorted.

src/nupic/research/temporal_memory.py
```diff
@@ -23,16 +23,109 @@
...
+  @param cellsPerColumn   (int)    Number of cells per column in the tm
+  @param connections      (Object) Connections object of the tm
+
+  @return (dict){
+                  `column`               (int),
+                  `isActiveColumn`       (bool),
+                  `activeSegmentsBegin`  (int),
+                  `activeSegmentsEnd`    (int),
+                  `matchingSegmentsBegin`(int),
+                  `matchingSegmentsEnd`  (int)
```
mrcslws: Rather than Begin and End values, maybe it would be better to use an `xrange`? This would continue the trend of "Make our Python code more readable than our C++ by using generators instead of iterators". In other words, this would return a dict with 'activeSegments' and 'matchingSegments' set to an xrange.

src/nupic/research/temporal_memory.py
```diff
@@ -123,416 +214,197 @@ def compute(self, activeColumns, learn=True):
...
 
-  # ==============================
-  # Phases
-  # ==============================
-
-  def activateCorrectlyPredictiveCells(self,
-                                       prevPredictiveCells,
-                                       prevMatchingCells,
-                                       activeColumns):
+  def activatePredictedColumn(self, excitedColumn, learn, prevActiveCells):
```
mrcslws: Here's a conversation topic: maybe these should be functions with lots of parameters, rather than methods that access the `self`.

That's how I did it in C++, not because it's necessary, but because it makes it easy to reason about. For example, it's really useful being able to see from the `compute` method where the random number generator gets consumed: https://github.com/numenta/nupic.core/blob/a83ed19fc2dcf901352da865d0b9f42892b8fa60/src/nupic/algorithms/TemporalMemory.cpp#L609

`activatePredictedColumn` might take in parameters `connections, excitedColumn, learn, prevActiveCells, permanenceIncrement, permanenceDecrement` and it would not take in a `self`. It would just be a function. In many ways this is what the Python `TemporalMemory` does currently, it just uses methods instead of functions... probably to support overriding.

My coding style was to put the output params at the beginning of the list.

src/nupic/research/temporal_memory.py
```diff
@@ -123,416 +214,197 @@ def compute(self, activeColumns, learn=True):
...
-    @param prevActiveSegments           (set)         Indices of active segments in `t-1`
-    @param learningSegments             (set)         Indices of learning segments in `t`
-    @param prevActiveCells              (set)         Indices of active cells in `t-1`
-    @param winnerCells                  (set)         Indices of winner cells in `t`
-    @param prevWinnerCells              (set)         Indices of winner cells in `t-1`
-    @param connections                  (Connections) Connectivity of layer
-    @param predictedInactiveCells       (set)         Indices of predicted inactive cells
-    @param prevMatchingSegments         (set)         Indices of segments with
+                      `cells`         (set),
+                      `bestCell`      (set),
```
mrcslws: I think this is an int, not a set.

andrewmalta13:
@mrcslws , incorporates the changes you suggested except the more functional change you suggested. We can talk about it when you get in. I think I agree with the change, but we probably should suggest if we still want to be able to support extending TemporalMemory.

src/nupic/research/temporal_memory.py
```diff
@@ -120,419 +209,192 @@ def compute(self, activeColumns, learn=True):
...
     Indicates the start of a new sequence. Resets sequence state of the TM.
     """
-    self.activeCells = set()
-    self.predictiveCells = set()
-    self.activeSegments = set()
-    self.winnerCells = set()
-
+    self.activeCells = []
+    self.activeSegments = []
+    self.winnerCells = []
```
mrcslws: This should also reset the `matchingSegments`.

(I realize the code had this issue before. It might not have caused problems because the old code doesn't make full use of the matchingSegments, I think it only used them for punishment but not for computing the best matching segment -- it essentially computed the matching segments twice.)

src/nupic/research/temporal_memory.py
mrcslws: You made this change, but now the documentation is stale :)

src/nupic/research/temporal_memory.py
```diff
@@ -120,419 +209,192 @@ def compute(self, activeColumns, learn=True):
...
 
-  def bestMatchingCell(self, cells, activeCells, connections):
+  def bestMatchingSegment(self, excitedColumn, prevActiveCells):
     """
-    Gets the cell with the best matching segment
-    (see `TM.bestMatchingSegment`) that has the largest number of active
-    synapses of all best matching segments.
-
-    If none were found, pick the least used cell (see `TM.leastUsedCell`).
+    Gets the segment on a cell with the largest number of activate synapses.
```
mrcslws: Typo, should be "largest number of active synapses"

mrcslws:
Here's a remaining algorithmic difference between the Python and C++. Since the C++ stores a `SegmentOverlap` struct, it doesn't have to recompute these overlaps. This Python code basically computes the overlaps for each matching segment twice (or, rather, the matching segments in bursting columns). The C++ used to do the same thing -- this was the reason I changed it to store segment + overlap pairs.

This is true independent of your change, and it doesn't affect the results, so I'd be fine with us leaving it this way and deciding later whether to keep optimizing.

cogmission:
Hi Guys,

Is it still too early to ask if there's any insight yet on the performance difference this change makes?

andrewmalta13:
Oh @mrcslws, that is a pretty big performance difference between two implementations. I would support doing that change as a separate pr, but I would like to do it. @cogmission the change seems to be only about 15-20% faster judging from the times in the scripts/temporal_memory_benchmark and extensive_temporal_memory_test.  But who knows, the difference that marcus just brought up could be a pretty substantial speed up for the algorithm.

src/nupic/research/temporal_memory.py
```diff
@@ -120,419 +209,192 @@ def compute(self, activeColumns, learn=True):
...
     Gets the cell with the smallest number of segments.
     Break ties randomly.
 
-    @param cells       (set)         Indices of cells
-    @param connections (Connections) Connectivity of layer
+    @param cells       (list)         Indices of cells
 
     @return (int) Cell index
     """
     leastUsedCells = set()
```
mrcslws: Similarly, this can now be a list, since the cells are always stored in sorted lists.

src/nupic/research/temporal_memory.py
```diff
@@ -545,96 +407,77 @@ def leastUsedCell(self, cells, connections):
...
+      presynapticCell = self.connections.dataForSynapse(synapse).presynapticCell
 
-    for synapse in connections.synapsesForSegment(segment):
-      synapseData = connections.dataForSynapse(synapse)
+      if presynapticCell in candidates:
+        candidates.remove(presynapticCell)
 
-      if synapseData.presynapticCell in activeCells:
-        synapses.add(synapse)
+    candidates = sorted(candidates)
```
mrcslws: This method is definitely fine, but it's basically using two strategies for removing candidates, and I think it would make sense to just use one strategy. Do you think it would be hard to do swapping for the first half of this method, rather than set `remove`? Then it would be like the C++, which does all of this with a single candidates vector, with no extra sorting.

src/nupic/research/temporal_memory.py
```diff
@@ -545,96 +407,77 @@ def leastUsedCell(self, cells, connections):
...
-    # Pick n cells randomly
-    for _ in range(n):
-      i = self._random.getUInt32(len(candidates))
-      cells.add(candidates[i])
-      del candidates[i]
-
-    return cells
-
+  @staticmethod
+  def binSearch(arr, val):
```
mrcslws: This seems a little bit out of place as a TemporalMemory method. Maybe just make it a function? I don't have a strong stance on this, but that's what I'd prefer.

src/nupic/research/temporal_memory.py
```diff
@@ -120,419 +209,192 @@ def compute(self, activeColumns, learn=True):
...
-
-    for cell in prevPredictiveCells:
-      column = self.columnForCell(cell)
 
-      if column in activeColumns:
-        activeCells.add(cell)
-        winnerCells.add(cell)
-        predictedActiveColumns.add(column)
+    cellsToAdd = []
+    newCell = True
```
mrcslws: I don't think this `newCell = True` line is necessary.

mrcslws: And really the `newCell = False` line isn't necessary either.

src/nupic/research/temporal_memory.py
```diff
@@ -120,419 +209,192 @@ def compute(self, activeColumns, learn=True):
...
-    # Sort segments before iterating for compatibility with C++
-    # Sort with primary key = cell idx, secondary key = segment idx
-    segments = sorted(
-      segments,
-      key=lambda segment: (connections.cellForSegment(segment), segment))
-
-    for segment in segments:
-      isLearningSegment = segment in learningSegments
-      isFromWinnerCell = connections.cellForSegment(segment) in winnerCells
+    cells = self.cellsForColumn(excitedColumn["column"])
```
mrcslws: This line can be moved to the `else`. That will help perf. Really we could just pass the column number into `leastUsedCell`.

Also, `cellsForColumn` is slow since it doesn't use an xrange, but that's independent of this change. I'd like to see us stop using `cellsForColumn` as part of the algorithm, but we don't have to do that now.

mrcslws: Hey, sorry, I told you wrong. I didn't notice the method was returning `cells`.

src/nupic/research/temporal_memory.py
```diff
@@ -120,394 +222,200 @@ def compute(self, activeColumns, learn=True):
...
-        if permanence > 0 and self.predictedSegmentDecrement > 0:
-          numActiveSynapsesForSegment[segment] += 1
-
-          if numActiveSynapsesForSegment[segment] >= self.minThreshold:
-            matchingSegments.add(segment)
-            matchingCells.add(connections.cellForSegment(segment))
+                      `cells`         (list),
+                      `bestCell`      (int),
+    """
+    cells = []
```
mrcslws: My bad. This needs to get populated in both paths, since it gets returned by the method.

tests/unit/nupic/research/temporal_memory_test.py
```diff
@@ -41,12 +42,25 @@
...
+  @staticmethod
+  def isSynapseDestroyed(connections, synapse):
+    try:
+      connections.dataForSynapse(synapse)
+      return False
+    except KeyError:
+      return True
 
   def setUp(self):
     self.tm = TemporalMemory()
```
mrcslws: This is now mostly unused. Only a couple tests use it, but every test pays the cost of it. It's probably worth moving this line into the three tests that use `self.tm` and removing this `setUp` method.

tests/unit/nupic/research/temporal_memory_test.py
```diff
@@ -58,369 +72,831 @@ def testInitInvalidParams(self):
...
+    weakInactSynapse = tm.connections.createSynapse(activeSegment,
+                                                    previousActiveCells[3],
+                                                    .009)
+
+    tm.compute(previousActiveColumns, True)
+    tm.compute(activeColumns, True)
+
+    self.assertTrue(self.isSynapseDestroyed(tm.connections, weakInactSynapse))
+
+  # createSynapse in connections.py does not destroy the weaset synapses
```
mrcslws: Typo: "weaset"

rhyolight: Yes, I think you meant "weasette", which is a female weasel.

src/nupic/research/connections.py
```diff
@@ -242,6 +259,48 @@ def updateSynapsePermanence(self, synapse, permanence):
...
+                      `matchingSegments`        (list),
+
+    Notes:
+      activeSegments and matchingSegments are sorted by their corresponding presynaptic cell.
+    """
+
+    numActiveSynapsesForSegment = defaultdict(int)
+    numMatchingSynapsesForSegment = defaultdict(int)
+    activeSegments = set()
+    matchingSegments = set()
```
mrcslws: This is building up a sorted list. It should be faster to just use an unsorted list and then sort it at the end.

i.e. we can just change these to lists.

src/nupic/research/connections.py
```diff
@@ -73,6 +72,17 @@ def __init__(self,
     # Index of the next synapse to be created
     self._nextSynapseIdx = 0
 
+  def segmentCmp(self, a, b):
+    """
+    A simple comparison function for segments to sort them
+    by their corresponding presynapticCell.
```
mrcslws: Segments don't have presynaptic cells. They're on a cell, and they synapse with a bunch of presynaptic cells.

src/nupic/research/connections.py
mrcslws: I see, using a `set` is also doing duplicate detection, since we'll often add a segment to this list multiple times.

In the C++ I do two passes. First I compute all the `numActiveSynapsesForSegment`s, then I walk them once to build up the unsorted activeSegments list. This is necessary since I save the segment + overlap pair, but it should also be faster since we're not doing a bunch of set insertions.

tests/unit/nupic/research/temporal_memory_test.py
```diff
@@ -635,8 +1096,9 @@ def testMapCellsToColumns(self):
     self.assertEqual(columnsForCells[99], set([399]))
 
 
-  @unittest.skipUnless(
-      capnp, "pycapnp is not installed, skipping serialization test.")
+  @unittest.skip("Serialization does not preserve numbering of cells\
+                  and segments, whichis needed in the\
+                  columnSegmentWalk implementation")
   def testWriteRead(self):
```
mrcslws: I don't think this change should affect whether serialization works. It's true that the C++ implementation makes sure segments and synapses always have the same index before and after a `destroySegment` and `destroySynapse`, but I don't think that matters here.

src/nupic/research/connections.py
```diff
@@ -242,6 +259,48 @@ def updateSynapsePermanence(self, synapse, permanence):
     # Update indexes
     self._synapsesForPresynapticCell[newData.presynapticCell][synapse] = newData
 
+  def computeActivity(self, activeInput, activePermanenceThreshold,
```
mrcslws: From the [style guide](https://github.com/numenta/nupic/wiki/Python-Style-Guide): "Separate methods inside classes by two blank lines."

src/nupic/research/connections.py
```diff
@@ -19,8 +19,7 @@
 # http://numenta.org/licenses/
 # ----------------------------------------------------------------------
 
-from collections import defaultdict, namedtuple
-
+from collections import defaultdict
 
 
 class SynapseData(object):
```
mrcslws: I think we want 3 blank lines above this, not 2. "Separate top level functions and classes by three blank lines."

andrewmalta13:
@mrcslws I implemented it using numpy arrays and generated lists using numpy iterators and I must have done something wrong because it basically tripled the call time spent in computeActivity.  It is currently implemented just using normal python arrays.

src/nupic/research/temporal_memory.py
```diff
@@ -518,124 +428,102 @@ def bestMatchingSegment(self, cell, activeCells, connections):
...
-      if synapseData.presynapticCell in activeCells:
-        synapses.add(synapse)
-
-    return synapses
+      presynapticCell = connections.dataForSynapse(synapse).presynapticCell
+      index = binSearch(candidates, presynapticCell)
+      if index != -1:
+        candidates[index] = candidates[eligibleEnd]
+        eligibleEnd -= 1
+
```
mrcslws: nit: there's some trailing whitespace in these files.

src/nupic/research/temporal_memory.py
```diff
@@ -23,18 +23,124 @@
...
+                            cellsPerColumn,
+                            connections):
+  """
+  Generator used for iterating over the lists of active columns,
+  active segments, and matching segments, each sorted by the column they
+  correspond to.
+
+  @param activeColumns    (list)   List of currently active columns
+  @param activeSegments   (list)   List of segments active from lateral input
+  @param matchingSegments (list)   List of segments matching from lateral input
```
mrcslws: In general these should probably say "Sorted list of ...".

src/nupic/research/temporal_memory.py
```diff
@@ -23,18 +23,124 @@
...
+                  `column`               (int),
+                  `isActiveColumn`       (bool),
+                  `activeSegmentsBegin`  (int),
+                  `activeSegmentsEnd`    (int),
+                  `matchingSegmentsBegin`(int),
+                  `matchingSegmentsEnd`  (int)
+                }
+
+  Notes:
+    The ints returned (activeSegmentsBegin/End, matchingSegmentsBegin/End)
```
mrcslws: This still talks about activeSegmentsBegin, etc., which is no longer returned

src/nupic/research/temporal_memory.py
```diff
@@ -120,394 +224,218 @@ def compute(self, activeColumns, learn=True):
...
+    @param excitedColumn       (dict)   Excited Column instance from excitedColumnsGenerator
+    @param learn               (bool)   Whether or not learning is enabled
+    @param initialPermanence   (float)  Initial permanence of a new synapse.
+    @param matchingSegments    (list)   List of segments matching from lateral input
+    @param maxNewSynapseCount  (int)    The maximum number of synapses added to a segment during learning.
+    @param permanenceDecrement (float)  Amount by which permanences of synapses are decremented during learning.
+    @param permanenceIncrement (float)  Amount by which permanences of synapses are incremented during learning.
+    @param prevActiveCells     (list)   Active cells in `t-1`
+    @param prevWinnerCells     (list)   Winner cells in `t-1`
+    @param random              (object) Tm object used to generate random numbers
```
mrcslws: It might be best to just say "Random number generator" rather than specifying that it's associated with a "Tm". (Ditto for the other occurrences of "param random")

mrcslws:
I left a couple more documentation comments, but I'm otherwise done reviewing. It looks good.

The one open issue for me: I still don't love the static methods. It makes the invocations really long, e.g. `TemporalMemory.activatePredictedColumn()`. Since these aren't intended to be part of the class's public interface, I think these would make sense as regular functions outside the class, similar to [the C++](https://github.com/numenta/nupic.core/blob/9956a98a4dcd57d689f577ccfe9c16cfc8613f3b/src/nupic/algorithms/TemporalMemory.cpp#L451). I figure we'll bring in @scottpurdy to discuss. :)

tests/unit/nupic/research/temporal_memory_test.py
```diff
@@ -59,368 +72,848 @@ def testInitInvalidParams(self):
...
-
-    self.assertEqual(activeCells, set([]))
-    self.assertEqual(winnerCells, set([]))
-    self.assertEqual(predictedColumns, set([]))
-    self.assertEqual(predictedInactiveCells, set([32,47]))
-
-  def testBurstColumns(self):
+  def testActivateCorrectlyPredictedCells(self):
     tm = TemporalMemory(
+      columnDimensions=[32],
```
scottpurdy: Use a tuple here. And avoid mutable data types as default values since the same mutable value will be shared for all function calls.

tests/unit/nupic/research/temporal_memory_test.py
```diff
@@ -59,368 +72,848 @@ def testInitInvalidParams(self):
...
+
+    tm.compute(previousActiveColumns, True)
+    self.assertEqual(expectedActiveCells, tm.getPredictiveCells())
+    tm.compute(activeColumns, True)
+    self.assertEqual(expectedActiveCells, tm.getActiveCells())
+
+
+  def testBurstUnpredictedColumns(self):
+    tm = TemporalMemory(
+      columnDimensions=[32],
```
scottpurdy: tuple here as well

scottpurdy: nvm read this code wrong

src/nupic/research/connections.py
```diff
@@ -243,6 +265,52 @@ def updateSynapsePermanence(self, synapse, permanence):
     self._synapsesForPresynapticCell[newData.presynapticCell][synapse] = newData
 
 
+  def computeActivity(self, activeInput, activePermanenceThreshold,
+                      activeSynapseThreshold, matchingPermananceThreshold,
+                      matchingSynapseThreshold):
+    """
+    Computes active and matching segments given the current active input.
```
scottpurdy: You can start doc strings on the first line:
https://www.python.org/dev/peps/pep-0257/

src/nupic/research/temporal_memory.py
```diff
@@ -23,18 +23,124 @@
...
+    The ints returned (activeSegmentsBegin/End, matchingSegmentsBegin/End)
+    represent bounds of indices into the lists activeSegments and
+    matchingSegments. Together they form an interval of segments for the given
+    column with the schema (note begin is closed and end is open): [begin, end).
+  """
+  activeColumnsProcessed = 0
+  activeSegmentsProcessed = 0
+  matchingSegmentsProcessed = 0
+
+  activeColumnsNum    = len(activeColumns)
```
scottpurdy: Don't line things up like this. It's super annoying when you make changes later. Some people still do it for doc strings which I guess is personal preference but avoid it for actual code.

src/nupic/research/temporal_memory.py
```diff
@@ -120,394 +224,218 @@ def compute(self, activeColumns, learn=True):
...
-            matchingCells.add(connections.cellForSegment(segment))
-
-    return activeSegments, predictiveCells, matchingSegments, matchingCells
 
+  @staticmethod
+  def punishPredictedColumn(connections, excitedColumn, matchingSegments,
+                            predictedSegmentDecrement, prevActiveCells):
+    """
+    Punishes the Segments that incorrectly predicted a column to be active.
+    @param connections         (Object) Connections instance for the tm
```
scottpurdy: Put a blank line before listing params

andrewmalta13:
@scottpurdy addressed your initial comments.

scottpurdy:
@mrcslws let me know if/when you're happy with this and I'll do final review

mrcslws:
Looks good! 👍 @scottpurdy I'll be curious to see what you think about the static methods and how they're invoked. I can go either way.

src/nupic/research/connections.py
```diff
@@ -243,23 +249,67 @@ def updateSynapsePermanence(self, synapse, permanence):
...
+    matchingSegments = []
+    for i in xrange(self._nextSegmentIdx):
+      if numActiveSynapsesForSegment[i] >= activeSynapseThreshold:
+        activeSegments.append(i)
+
+    for i in xrange(self._nextSegmentIdx):
+      if numMatchingSynapsesForSegment[i] >= matchingSynapseThreshold:
+        matchingSegments.append(i)
+
+    return (sorted(activeSegments, cmp=self.segmentCmp),
```
scottpurdy: Using lambda functions is probably easier here since it's such a simple comparison and used only in these two places:

```
segmentCmp = lambda a, b: self._segments[a] - self._segments[b]
return (sorted(activeSegments, cmp=segmentCmp),
        sorted(activeSegments, cmp=segmentCmp))
```

src/nupic/research/temporal_memory.py
```diff
@@ -23,22 +23,125 @@
...
+from bisect import bisect_left
 
 EPSILON = 0.000001
 
 
 
-class TemporalMemory(object):
-  """
-  Class implementing the Temporal Memory algorithm.
+def binSearch(arr, val):
```
scottpurdy: Add a doc string for this. It's hard for me to tell what it is supposed to do.

tests/unit/nupic/research/temporal_memory_test.py
```diff
@@ -59,368 +68,848 @@ def testInitInvalidParams(self):
...
-                                                                  prevMatchingCells,
-                                                                  activeColumns)
-
-    self.assertEqual(activeCells, set([]))
-    self.assertEqual(winnerCells, set([]))
-    self.assertEqual(predictedColumns, set([]))
-    self.assertEqual(predictedInactiveCells, set([32,47]))
-
-  def testBurstColumns(self):
+  def testActivateCorrectlyPredictedCells(self):
```
scottpurdy: I can't tell exactly from this diff but are you familiar with the distinction between "predictive" and "predicted" in this code? "predicted" is used to refer to predictions from the previous time step that you would compare or use in current activations. "predictive" is used to refer to cells in the current step that are predicted to be active in the next time step.

I don't like the terminology but we have to keep it consistent in this file or it will be impossible to understand. So make sure you use it accordingly.

scottpurdy:
Skimmed through and put a few comments.

tests/unit/nupic/research/temporal_memory_test.py
andrewmalta13: After a bit of confusion I did pick up on that; however, the test suite in question uses the same variable names as the c++ version to ensure that they are easily relatable to someone looking at both of them. So the test suite shouldn't have the "predicted" vs. "predictive" confusion.

scottpurdy:
@andrewmalta13 this looks good to merge except that we can't break the serialization. It seems from your description that the new serialization works but the old serialization breaks. @mrcslws can you advise on this (since I recall we discussed how to handle it in the C++).

mrcslws:
Andrew and I looked into the serialization issue a little bit. Independently of this change, the Connections class's segment numbers get rearranged when serialized and deserialized. This is true independent of this change. So I think the `activeSegments` and `matchingSegments` have always been screwy for one timestep after deserialization, and this new implementation relies even more on them being correct, whereas the previous implementation stored predictive cells, matching cells rather than just using the segments.

Maybe we should figure out the fix for that issue, then decide what to do? I think it will give us some insight.

andrewmalta13:
Yeah Marcus and I were a little thrown that is was working before the changes.  I thought at first that it was breaking because of something specific to the column segment walk algorithm, but it seems that it is doing something equally weird in both implementations.  I think it really comes down to the fact that we store the arrays of matching and active segments in the serialization which both are just arrays of integers corresponding to the segments and then we renumber segments in deserialization. This leads us with **incorrect** matching and active segments for the next time step which is certain to cause trouble with the new implementation.

scottpurdy:
Can you guys propose a resolution to the serialization issue? Then we can decide how to proceed. If this is a problem already then it may be fine to merge this without the fix but we should address that ASAP and make sure there are tests to enforce proper behavior.

andrewmalta13:
@scottpurdy discussed with @mrcslws and we both agree that the best way to approach the issue is not to change how we are serializing, but rather change how we keep track of segments in connections.py.  Right now a segment is just an integer determined by the order it was created. This differs from the C++ implementation which uses a struct that contains a similar id for a segment.  If we make this change, the serialization would not mess up the numbering of these segments as before. The questions I have are:
1. Is this an acceptable resolution in your eyes
2. If so should this be included in this PR or a separate one?

mrcslws:
I think I vote "let's fix that separately". It's not surprising that using a flat list of segments in the Python `Connections`, combined with a Cell -> Segments -> Synapses serialization format, leads to the segment indices getting scrambled on serialize + deserialize.

scottpurdy:
1. Yes that's a good solution.
2. Given that this isn't a new issue and existed before this PR, yes you can do that as a follow up PR. Does that mean this PR is ready to go?

andrewmalta13:
@scottpurdy unless there is anything else you want to be included, it is ready to merge.

scottpurdy:
👍

PR #3213, opened 2016-07-15
DEVOPS-85 OS X build infrastructure for Bamboo CI environment

oxtopus:
Adds a `build-osx.sh` script and associated vagrant configuration for executing in a VM.


ci/bamboo/build-osx.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+set -o errexit
+set -o xtrace
+
+# Fixup $PATH for --user installation
+export PATH=${HOME}/Library/Python/2.7/bin:${PATH}
+export PYTHONPATH=${HOME}/Library/Python/2.7/lib/python/site-packages:${PATH}
```
lscheinkman: It should append `PYTHONPATH` instead of `PATH` here:
```
export PYTHONPATH=${HOME}/Library/Python/2.7/lib/python/site-packages:${PYTHONPATH}
```

oxtopus: D'oh!  Yep.  Though, it seems to be working as is as I assume that $PYTHONPATH is already empty.

PR #3215, opened 2016-07-19
Upgrade pip

scottpurdy:
@oxtopus please review

### DIFF
--- a/ci/bamboo/build-linux.sh
+++ b/ci/bamboo/build-linux.sh
@@ -38,7 +38,7 @@ apt-get install -y \
     python2.7 \
     python2.7-dev \
     python-pip
-pip install --upgrade --ignore-installed setuptools
+pip install --upgrade --ignore-installed setuptools pip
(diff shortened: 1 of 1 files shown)


oxtopus:
👎 installation of pip should be consistent with https://github.com/numenta/nupic.core/blob/master/ci/bamboo/setup-dependencies-linux.sh#L44

I'm currently testing such a change in a branch.

scottpurdy:
Agree re: consistency. I'll be moving this step to a base Docker image for CI though and I'd like to use this method instead of bundling `get-pip`.

oxtopus:
`python-pip` is a MUCH older version of pip (1.5.4, see http://packages.ubuntu.com/trusty/python-pip).

`get-pip.py` is the official, recommended installation method for pip at https://pip.pypa.io/en/stable/installing/ and works universally between nupic.core linux and os x builds.

I fail to see why we should depend on the distribution package given the circumstances.

PR #3216, opened 2016-07-19
Install pip from local copy, other simplifications

oxtopus:
This installs pip from local `get-pip.py`, consistent with os x and nupic.core builds.  Getting passing jobs on a branch at https://ci.numenta.com/browse/NUP-PY80-5

### DIFF
--- a/ci/bamboo/build-linux.sh
+++ b/ci/bamboo/build-linux.sh
@@ -36,13 +36,12 @@ apt-get install -y \
     build-essential \
     python \
     python2.7 \
-    python2.7-dev \
-    python-pip
-pip install --upgrade --ignore-installed setuptools
-pip install wheel
+    python2.7-dev
 
-# Move into root of nupic repository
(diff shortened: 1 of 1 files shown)


BoltzmannBrain:
👍

scottpurdy:
👍 but like I mentioned elsewhere I'd like to remove `get-pip` from the repo and have pip installed in infrastructure instead.

oxtopus:
re:

> I'd like to remove get-pip from the repo and have pip installed in infrastructure instead.

I believe those plans are in motion.  Luiz is working on getting the vagrant image w/ pip installed and you are working on getting it installed in a docker image.  That is a separate concern from the technique used to install it.

Considering we have to work with windows, os x, and linux, since they each need pip.  Installing from `get-pip.py` is the preferred way, since it works consistently across platforms, whereas using apt in ubuntu requires that we do something different in windows and os x.

PR #3220, opened 2016-07-22
anomaly likelihood region

### LINKED ISSUE #3219: Implement Anomaly Likelihood Region
- Creating a simple and clean AnomalyLikelihoodRegion
- Remove `src/nupic/algorithms/Anomaly.py` and `src/nupic/regions/AnomalyRegion.py`

andrewmalta13:
fixes #3219 

- Implementation of an anomaly likelihood region for the network API
- Removes anomaly region, much of `src/nupic/algorithms/anomaly.py`, and the dependencies that other files have with them.

Now computes anomaly score in the TPregion when the anomalyMode parameter is set to true. 

Also updates the network examples in `examples/network` to not rely on the old anomaly region and instead uses the TPRegion for raw anomaly score calculation.  (network_api_demo.py now uses the anomaly likelihood region to provide an example for how to use it).

tests for the region exist in `tests/unit/nupic/regions/anomaly_likelihood_region_test.py` and any additional test ideas are welcome.

Please review @scottpurdy


andrewmalta13:
Still needs to remove anomaly.py from opf model.

andrewmalta13:
ready for first round of review (doesn't seem like travis is building right now)

scottpurdy:
I don't think the raw anomaly score calculation should be in the TPRegion. If we are saying that it is specific to the temporal memory then I'd put it in temporal_memory.py and then expose it in the region. But the region itself shouldn't be responsible for logic like that.

Alternatively, since the raw anomaly score is independent of any specific TM implementation, you could keep that function in `nupic.algorithms.anomaly` and then call that from the region. I lean towards this I think.

andrewmalta13:
@scottpurdy I thought I was supposed to get rid of nupic.algorithms.anomaly, this is where it was before. Are you suggesting we go back to that?

scottpurdy:
@andrewmalta13 we want to replace the region that merges several different anomaly score calculations with one that is responsible just for the anomaly likelihood computation. If we have some non-likelihood anomaly-related logic that doesn't belong to any specific algorithm then `nupic.algorithms.anomaly` may be a good location for it.

src/nupic/regions/AnomalyLikelihoodRegion.py
```diff
@@ -0,0 +1,148 @@
...
+      "inputs": {
+        "rawAnomalyScore": {
+          "description": "The anomaly score whose \
+                          likelihood is to be computed",
+          "dataType": "Real32",
+          "count": 1,
+          "required": True,
+          "isDefaultInput": False
+        },
+        "value": {
```
scottpurdy: `"metricValue"` might be a better name to distinguish it from the anomaly value

src/nupic/regions/AnomalyLikelihoodRegion.py
```diff
@@ -0,0 +1,148 @@
...
+      "outputs": {
+        "anomalyLikelihood": {
+          "description": "The resultant anomaly likelihood",
+          "dataType": "Real32",
+          "count": 1,
+          "isDefaultOutput": True,
+        },
+      },
+      "parameters": {
+        "claLearningPeriod": {
```
scottpurdy: This region shouldn't be specific to HTM (or CLA which we have stopped using). I'd just call this parameter "probationPeriod"

andrewmalta13: should we change this in nupic.algorithms.anomaly_likelihood.py? These are the names of the instance variables of the class.

scottpurdy: Yes unless @subutai has some rationale for keeping it.

subutai: No particular rationale for that term - it was put in years ago. Keep in mind backward compatibility though - various products are using the anomaly likelihood module directly.

In this class we shouldn't use the term CLA.

rhyolight:
As I stated on [the issue topic](https://github.com/numenta/nupic/issues/3219#issuecomment-236432163), I'm 👎  on this PR as-is. I will approve it without the removal of `Anomaly` and `AnomalyRegion`.

rhyolight:
👍 😄

scottpurdy:
@breznak - Does this break anything for you now? I'd like to merge it tomorrow.

rhyolight:
Three days is enough to wait. We left in the code @breznak was concerned about, so 👍 .

PR #3227, opened 2016-07-27
Remove nupic dependency from hypersearch

### LINKED ISSUE #3226: Remove dependency on nupic namespace in hypersearch
As part of hypersearch refactoring efforts, separate NuPIC-specific swarming from _hypersearch_, keeping most of the current implementation and API.

natoromano:
Fixes #3226 by separating code in swarming that is dependent on NuPIC (i.e. NuPIC-specific swarming and adapter code, which will live in nupic.swarming) from generic hypersearch code (which will live in nupic.swarming.nupic.hypersearch for now).

At this point, nothing is changed to the implementation and the API. A lot of generic hypersearch classes such as HypersearchWorker are still un nupic.swarming as some aspects of their implementation rely on NuPIC-specific aspects such as ClientJobsDAO or the OPF framework.


natoromano:
@oxtopus ready for review. Let me know if this is what you expected. I think it's easier to look directly at the code rather than the diff, since there a some classes being moved around.

scottpurdy:
👍  assuming this is just  moving code around and the tests all still pass

natoromano:
Yes it's only moving code around. Some NuPIC utilities have been copied in `hypersearch` for now (such as `Configuration` and `Enum`), and this will be refactored away with the bulk `hypersearch` implementation.

It passes the CI tests, not sure what else to test it to.

PR #3254, opened 2016-08-08
Full sync of C++ temporal memory to Python temporal memory

### LINKED ISSUE #1511: Create python version of C++ Connections data structure
And refactor Temporal Memory python implementation to use it.

### LINKED ISSUE #2980: C++/Py TM compatibility test needs to be fixed to compare against C++ implementation
C++/Py TM compatibility test (#2954) currently just runs the Py TM against itself since it was created before C++ bindings were finished. Now that they are finished, it should be updated. Additionally, `TPRegion` needs to be updated to accept the new C++ TM implementation.

Requires numenta/nupic.core#839, numenta/nupic.core#840
Related to #2949

@chetan51

### LINKED ISSUE #2999: TM compatibility tests fail when predictedSegmentDecrement is nonzero

### LINKED ISSUE #3199: Temporal Memory serialization corrupts connections object
After the changes to temporal_memory.py detailed by <a href="https://github.com/numenta/nupic/pull/3197"> this </a> pr, it became clear that the serialization happening in connections.py is somehow corrupting the state is trying to preserve.

You can observe this failure in <a href="https://github.com/andrewmalta13/nupic/blob/7a443a651f934665bc204b9f3670ff2f156611b5/tests/unit/nupic/research/temporal_memory_test.py#L1104"> this </a> test which is currently skipped.

### LINKED ISSUE #3230: temporal_memory.py excitedColumnsGenerator() Has No Unit Test
Please correct me if I'm wrong, but I'm thinking [excitedColumnsGenerator() method](https://github.com/numenta/nupic/blob/master/src/nupic/research/temporal_memory.py#L55) is [central to the operation](https://github.com/numenta/nupic/blob/master/src/nupic/research/temporal_memory.py#L274) of the new TemporalMemory algorithm implementation - and I didn't find a unit test for it?

### LINKED ISSUE #3239: Ensure no remaining differences between python and c++ temporal memory implementations
- requires a large refactoring of src/nupic/research/connections.py and minimal changes to src/nupic/temporal_memory.py

- would fix #3199

### LINKED ISSUE #3240: Incorrect Documentation of Connections method
The documentation for the enclosed method, states that it, "...returns the column that a segment's presynaptic cell belongs to." What it does is **return the column of the ```Cell``` that _owns_ the segment.** If I'm not mistaken, ```Segments``` don't have presynaptic cells; ```Synapses``` do. I believe there may be many Cells that synapse to a segment, and each of those synapses have "presynaptic" cells (source cells).

The documentation [for the method should](https://github.com/numenta/nupic/blob/master/src/nupic/research/connections.py#L83) probably read:
 **Returns the column of the ```Cell``` which _owns_ the segment.**

```Python
 def columnForSegment(self, segment, cellsPerColumn):
    """ Returns the column that a segment's presynapticCell belongs to

    @param segment        (int) Segment index
    @param cellsPerColumn (int) Number of cells in a column in the tm

    @return (int)
    """
    return self._segments[segment] / cellsPerColumn
```

As a side note, this really messed me up trying figure out how to derive a presynaptic cell from a segment, until I realized it's not possible. :-P

### LINKED ISSUE #3253: In TemporalMemory we shouldn't use binary search in `growSynapses`
https://github.com/numenta/nupic/blob/8e40e7ad16fd3a04cc2a7d3d12174ccf3fa44daa/src/nupic/research/temporal_memory.py#L567

In the C++ we just do a `std::find`, not a `std::binary_search`.

https://github.com/numenta/nupic.core/blob/f670261a5d308b1497ef2991173c400c7d2a1b82/src/nupic/algorithms/TemporalMemory.cpp#L433

This is important because the list stops being sorted after the first candidate is removed.

andrewmalta13:
fixes #3239 
fixes #3253
fixes #3240 
fixes #3199 
fixes #3230 
fixes #1511
fixes #2999 
fixes #2980 

- Rewrite of connections.py and connections_tests.py
- Implements group_by.py and group_by_tests.py to be up to date with <a href="https://github.com/numenta/nupic.core/pull/1018"> this </a> soon to be merged pr to the c++ temporal memory.
- Reconfigures temporal_memory.py to use the new group_by method as opposed to the excitedColumnGenerator.
- Fixes serialization of the connections datastructure.
- bug fixes in temporal_memory.py
- compatibility test now can be run and it should pass when marcus' pr is merged into nupic.core.

Edit: This change is now ~7-10% faster than the current implementation (about 30% faster than the original phases TM).  Compatibility tests pass and fixes a bunch of tm issues in nupic.

Please Review:
@mrcslws @scottpurdy 

CC:
@cogmission


src/nupic/research/connections.py
```diff
@@ -20,23 +20,117 @@
...
 from collections import defaultdict
+from bisect import bisect_left
 
+EPSILON = 0.00001 # constant error threshold to check equality of permanences to
+                  # other floats
 
 
-class SynapseData(object):
+class Segment(object):
+  ''' Class containing mininal information to identify a unique segment '''
```
mrcslws: Typo: minimal

src/nupic/research/connections.py
```diff
@@ -52,92 +146,164 @@ def __init__(self,
...
+    """ Returns the data for a synapse.
+
+    @param synapse (Object) Synapse object
+
+    @return (SynapseData) Synapse data
     """
-    return self._segments[segment] / cellsPerColumn
 
+    return self._cells[synapse.segment.cell].segments\
+                      [synapse.segment.idx].synapses[synapse.idx]
```
mrcslws: I'm not sure if it will make a perf difference, but definitely feel free to do something like:

~~~python
    segment = synapse.segment

    return self._cells[segment.cell].segments\
                      [segment.idx].synapses[synapse.idx]
~~~

for this and for `dataForSegment` to avoid lookups. I didn't do this in the C++ because there's no overhead in C to do `foo.bar.baz` lookups, since it's all just a single compile-time binary offset no matter how long the chain is.

Your call.

cogmission:
Watching this...

scottpurdy:
What is the performance relative to the other version? @mrcslws do you think we should use this version?

tests/integration/nupic/engine/temporal_memory_compatibility_test.py
```diff
@@ -38,9 +38,9 @@ class TemporalMemoryCompatibilityTest(unittest.TestCase):
   """
 
 
-  @unittest.skip("There are slight differences between implementations that"
-                 "prevent this from passing still. See "
-                 "https://github.com/numenta/nupic/issues/2980")
+  # @unittest.skip("There are slight differences between implementations that"
+  #                "prevent this from passing still. See "
+  #                "https://github.com/numenta/nupic/issues/2980")
```
mrcslws: Just in case: remember to remove this.

tests/unit/nupic/research/connections_test.py
```diff
@@ -19,217 +19,416 @@
...
+    segment3 = connections.createSegment(42)
+    self.assertEqual(segment2.idx, segment3.idx)
 
-    self.assertEqual(connections.createSegment(0), 0)
-    self.assertEqual(connections.createSegment(0), 1)
-    self.assertEqual(connections.createSegment(10), 2)
 
-    self.assertEqual(connections.createSynapse(0, 254, 0.1173), 0)
-    self.assertEqual(connections.createSynapse(0, 477, 0.3253), 1)
+  def testSynapseReusue(self):
```
mrcslws: Typo: "reuse"

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,86 @@
...
+
+  if len(args) % 2 == 1:
+    raise ValueError("Must have a key function for every list.")
+
+  # zip up tuples of lists and their functions
+  tups = [(args[i], args[i+1]) for i in xrange(0, len(args), 2)]
+
+  # populate above lists
+  for listn, fn in tups:
+    groupsListEntry = [(k, list(g)) for k, g in groupby(listn, fn)]
```
mrcslws: Is the `list()` necessary? It seems like we could return a tuple of generators that we got from `groupby`. That would avoid creating a bunch of small lists.

When the list is empty (e.g. when there are no active segments in the column) we could return some other special value like `None`. Then callers don't need to know the length of the list, they can just check for `None` (so we don't have to return counts). This wouldn't be inconsistent with `groupby` usage since there's never a case in groupby where the generator is empty.

andrewmalta13: yeah that is what I originally did. Unfortunately the implementation of group_by in itertools make that tricky. http://stackoverflow.com/a/6236257/4882848

mrcslws: Ouch. So basically we can't use `itertools.groupby` if we want the code to be as fast as it was with the ExcitedColumnsGenerator. We'd have to write our own... which might not be too bad?

One nice thing about writing the `groupby` functionality ourselves is that we can give it useful stuff like a `__len__` method on the returned object, so that it's lazy but still works with `len()`. It'd make for a really clean API.

src/nupic/research/connections.py
```diff
@@ -251,50 +484,93 @@ def computeActivity(self, activeInput, activePermanenceThreshold,
...
+    for i in xrange(self._nextFlatIdx):
+      segment = self._segmentForFlatIdx[i]
+      numMatching = numMatchingSynapsesForSegment[i]
+      if numMatching >= matchingSynapseThreshold:
+        segmentOverlap = SegmentOverlap(segment, numMatching)
+        matchingSegments.append(segmentOverlap)
+
+    segCmp = lambda a, b: (a.segment.cell - b.segment.cell
+                           if a.segment.cell - b.segment.cell != 0
+                           else a.segment.idx - b.segment.idx)
```
mrcslws: This is slower than necessary since it always computes the difference twice. I don't know if it can be written fast as a lambda. 

Maybe do `sorted(activeSegments, key = lambda x: x.segment)` and then implement the `__lt__`, `__le__`, etc. methods on the `Segment` class. These methods could be written more efficiently than a lambda, since you can store the `segment.cell - segment.cell` value rather than compute it twice.

src/nupic/research/connections.py
```diff
@@ -52,92 +159,165 @@ def __init__(self,
...
     """
-    return self._segments[segment]
+    segmentData = self.dataForSegment(segment)
+    synapses = []
+    if segmentData.destroyed:
+      raise ValueError("Attempting to access destroyed segment's synapses")
+
+    for i in xrange(len(segmentData.synapses)):
+      if not segmentData.synapses[i].destroyed:
+        synapses.append(Synapse(i, segment))
```
mrcslws: Maybe this is already in your list of things to try, but: maybe use list comprehension? I played with it real quick and I think it's faster:

~~~python
    return [Synapse(i, segment)
            for i in xrange(len(segmentData.synapses))
            if not segmentData.synapses[i].destroyed]
~~~

cogmission:
@mrcslws 
>  One nice thing about writing the groupby functionality ourselves is that we can give it useful stuff like a __len__ method on the returned object, so that it's lazy but still works with len(). It'd make for a really clean API.

One request? Please don't hide the implementation behind a SWIG interface? Please do it in Python?

mrcslws:
We're talking about making `dataForSynapse`'s lookups faster... but maybe we should just make it return `synapse.data`. A `Synapse` instance could store a `SynapseData` reference. A `Segment` instance could store a `SegmentData` reference.

Maybe the C++ code could use a similar improvement. I haven't focused on this part of the code. In C++ `dataForSynapse` is much faster since it's just a couple vector lookups.

This would mean a `Synapse` and `Segment` instance would be invalidated after you serialize and deserialize. I think that's fine.

mrcslws:
(Actually that wouldn't work in C++ since you can't expect a pointer to stay valid when the data lives in a vector. As the vector grows the data structures might get moved around.)

cogmission:
In C++ u don't have pointers to the actual object? Or can you store the pointers in a vector? I don't know what the hell I'm talking about but just wondering?

src/nupic/research/connections.py
```diff
@@ -251,50 +487,90 @@ def computeActivity(self, activeInput, activePermanenceThreshold,
...
+      segment = self._segmentForFlatIdx[i]
+      numMatching = numMatchingSynapsesForSegment[i]
+      if numMatching >= matchingSynapseThreshold:
+        segmentOverlap = SegmentOverlap(segment, numMatching)
+        matchingSegments.append(segmentOverlap)
 
-    segCmp = lambda a, b: self._segments[a] - self._segments[b]
-    return (sorted(activeSegments, cmp = segCmp),
-            sorted(matchingSegments, cmp = segCmp))
+    segCmp = lambda s: s.segment
```
mrcslws: `segCmp` is no longer the right name. I think it'd be fine to just do inline lambdas in the `sorted` call.

src/nupic/research/connections.py
```diff
@@ -251,50 +487,90 @@ def computeActivity(self, activeInput, activePermanenceThreshold,
...
 
-    numActiveSynapsesForSegment = [0] * self._nextSegmentIdx
-    numMatchingSynapsesForSegment = [0] * self._nextSegmentIdx
+    numActiveSynapsesForSegment = [0] * self._nextFlatIdx
+    numMatchingSynapsesForSegment = [0] * self._nextFlatIdx
 
     for cell in activeInput:
-      for synapseData in self.synapsesForPresynapticCell(cell).values():
-        segment = synapseData.segment
+      for synapse in self.synapsesForPresynapticCell(cell):
```
mrcslws: I see the interface to Connections is changing slightly, which will break the PhasesTemporalMemory. So I think we'll need to either copy the old Connections to nupic.research, or change PhasesTemporalMemory to work with the updated Connections interface.

I see `synapsesForPresynapticCell` previously returned a dict, though its docstring says it was a set. Obviously your approach is better.

andrewmalta13: Oh good point, hmm @scottpurdy what do you think should happen?

cogmission:
@andrewmalta13 Interesting...! So now Segments are "marked" as destroyed instead of actually deleted, for more expedient reuse?

andrewmalta13:
@cogmission yeah, that is the biggest change to the connections data structure. It is needed to have complete parity with the c++ version. Also the way the python connections data structure was doing it before was flat out incorrect in my opinion, which is the main reason I wanted to work on this pr.

src/nupic/research/connections.py
```diff
@@ -20,23 +20,133 @@
...
+
+  def __init__(self, presynapticCell, permanence):
+    '''
+       @param presynapticCell (int)   The index of the presynaptic cell of the
+                                      synapse
+       @param permanence      (float) permanence of the synapse from 0.0 to 1.0
+    '''
     self.presynapticCell = presynapticCell
     self.permanence = permanence
+    self.destroyed = False  #boolean destroyed flag so object can be reused
```
mrcslws: nit: put a space in the comment `# boolean`

src/nupic/research/connections.py
```diff
@@ -20,23 +20,133 @@
...
 from collections import defaultdict
+from bisect import bisect_left
 
+EPSILON = 0.00001 # constant error threshold to check equality of permanences to
+                  # other floats
 
 
-class SynapseData(object):
+class Segment(object):
+  ''' Class containing minimal information to identify a unique segment '''
```
mrcslws: nit: Some of these docstrings are using single quotes, others are using double quotes. I think they should all be double.

mrcslws:
@cogmission The `destroyed` flag existed in the C++ before I started working on it. I decided it was worth keeping because it allows segments and synapses to have a constant path. "Synapse 42 on Segment 11 on Cell 129" will always correspond to the same synapse even after you destroy other segments and synapses around it. If destroying a synapse or segment removed the SynapseData or SegmentData from a list, it would change the indices of the remaining synapses or segments.

src/nupic/research/connections.py
```diff
@@ -52,92 +162,181 @@ def __init__(self,
...
+    """ Returns the data for a segment stored in the self._cells array
+
+    @param idx  (int) segment index on a cell
+    @param cell (int) cell index
+
+    @return (SegmentData) Segment data
 
-    @return (set) Synapse indices
+    NOTE: Only used in the read method in temporal memory for re-creating the
+          active and matching segments
```
mrcslws: Instead, maybe there should be a `connections.getSegment` or `connections.segment` method which takes a cell and segment index and returns a `Segment` object. Then `temporal_memory` won't have to instantiate Segments.

src/nupic/research/connections.py
```diff
@@ -52,92 +162,181 @@ def __init__(self,
...
-    return self._synapsesForSegment[segment]
+  def _minPermanenceSynapse(self, segment):
+    """ Internal method for finding the synapse with the smallest permanence
+    on the given segment.
+
+    @param segment (Object) Segment object to search for synapses on
+
+    @return (Object) Synapse object on the segment of the minimal permanence
+
+    Note: On ties it will chose the first occurance of the minimum permanence
```
mrcslws: "occurrence"

src/nupic/research/connections.py
```diff
@@ -52,92 +162,181 @@ def __init__(self,
...
+      synapseData = synapses[i].data
+      if (not synapseData.destroyed) and (synapseData.permanence
+                                          < minPermanence - EPSILON):
+        minIdx = i
+        minPermanence = synapseData.permanence
+
+    return synapses[minIdx]
+
+
+  def mostActiveSegmentForCells(self, cells, inputCells, synapseThreshold):
```
mrcslws: I think this method is dead code. We can remove it and the tests...

I removed them in the C++ here: https://github.com/numenta/nupic.core/pull/998#discussion_r70546162

cogmission:
> If destroying a synapse or segment removed the SynapseData or SegmentData from a list, it would change the indices of the remaining synapses or segments.

This is an advantage of an OO approach because for instance the Segment's index can be a member variable and distinct from the Segment's occurrence in a fever list. So it always has the same identity regardless of position changes in the list. 

But that's part of what makes this stuff fun to port, because you can think of different ways to express this stuff:-)

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,130 @@
...
+    for i in xrange(self.length):
+      self.index = i
+      yield self.iterable[i]
+
+
+  def __len__(self):
+    return self.length - self.index
+
+
+def groupby(lis, fn):
```
mrcslws: It's worth explaining why we can't use the built-in `itertools.groupby`.

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,130 @@
...
+
+  for i in xrange(len(lis)):
+    val = fn(lis[i])
+
+    if i == 0:
+      key = val
+      continue
+
+    if val != key:
+      yield (key, GroupByGenerator(lis[begining : i], length))
```
mrcslws: This `lis[begining : i]` is still creating a new list. Can we just pass in `lis`, `beginning`, and `length` to the GroupByGenerator?

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,130 @@
...
+  indexList = [] # list of [currentIndex, endIndex] pairs
+
+  if len(args) % 2 == 1:
+    raise ValueError("Must have a key function for every list.")
+
+  # populate above lists
+  for i in xrange(0, len(args), 2):
+    listn = args[i]
+    fn = args[i + 1]
+    groupsListEntry = [(k, g) for k, g in groupby(listn, fn)]
```
mrcslws: Calling `groupByN` immediately walks and partitions every list. So it doesn't have the same performance characteristics of a `groupby`. It does initial processing, and its memory usage increases with list length. It should be possible to do something that delays more of the work. This would cause less memory use and it would make better use of CPU cache, since it will tend to access parts of the list only when it's about to access them again immediately afterward.

For each list, we could store a groupby iterator and its next result. Whenever we yield one of the groupby's results, we'd call `next` on the iterator.

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,130 @@
...
+def groupby(lis, fn):
+  length = 1
+  begining = 0
+
+  for i in xrange(len(lis)):
+    val = fn(lis[i])
+
+    if i == 0:
+      key = val
+      continue
```
mrcslws: You could avoid checking for 0 every time by guarding the whole function with the `if len(lis):` from below. Then just have a `key = fn(lis[0])` line above the loop, and in the loop use an `xrange(1, len(lis))`

And that might be clearer if you say `if len(lis) > 0:`, not just `if len(lis)`

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,130 @@
...
+      yield self.iterable[i]
+
+
+  def __len__(self):
+    return self.length - self.index
+
+
+def groupby(lis, fn):
+  length = 1
+  begining = 0
```
mrcslws: "beginning"

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,130 @@
...
+                       their respective key functions. The lists should be
+                       sorted with respect to their key function.
+
+    @return (tuple) a n + 1 dimensional tuple, where the first element is the
+                    key of the group and the other n entries are lists of
+                    objects that are a member of the current group that is being
+                    iterated over in the nth list passed in. Note that this
+                    is a generator and a n+1 dimensional tuple is yielded for
+                    every group.
+
```
mrcslws: nit: There's some trailing whitespace here.

mrcslws:
Done reviewing. I left some groupby feedback, but otherwise I think we'll be ready to merge this soon.

cogmission:
@andrewmalta13 @mrcslws Just a small question and hope that the occurrence and usage of the new ```GroupByGenerator``` will be accompanied by documentation of what it accomplishes within the context of the TemporalMemory? I'm sure it will, but I'm just making sure... ;-)

mrcslws:
@cogmission the GroupByGenerator is an implementation detail. The thing to pay attention to is `groupByN`, which is essentially just `itertools.groupby` extended to work with multiple sequences. As long as the `groupByN` docstring is good, I think we're fine.

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,161 @@
...
+class GroupByGenerator(object):
+  """ A custom generator used by groupby to allow for len() calls on the
+      generator returned.
+  """
+
+  __slots__ = ['iterable', 'beginning', 'length', 'end', 'index']
+
+  def __init__(self, iterable, beginning, length):
+    self.iterable = iterable
```
mrcslws: Is "iterable" an appropriate name? Iterables don't necessarily support index lookup. Just calling it `lis` would be good.

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,161 @@
...
+    self.beginning = beginning
+    self.length = length
+    self.end = self.beginning + self.length
+    self.index = 0
+
+
+  def __iter__(self):
+    for i in xrange(self.beginning, self.end):
+      self.index = i
```
mrcslws: This loop shouldn't change the object, otherwise there will be a bug that you can't iterate over a `groupby` result in two places at once. (Not necessarily in parallel)

We don't need to worry about the `__len__` result updating as you iterate. It should just return the length of the original list subset.

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,161 @@
...
+  """
+
+  __slots__ = ['iterable', 'beginning', 'length', 'end', 'index']
+
+  def __init__(self, iterable, beginning, length):
+    self.iterable = iterable
+    self.beginning = beginning
+    self.length = length
+    self.end = self.beginning + self.length
+    self.index = 0
```
mrcslws: It's worth keeping this as small as possible. I don't think we need `self.end`, and I think `self.index` is not necessary based on my comments about `__len__`.

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,161 @@
...
+      if val != key:
+        yield (key, GroupByGenerator(lis, beginning, length))
+        length = 1
+        beginning = i
+        key = val
+      else:
+        length += 1
+
+  if l > 0: # yield last group
+    yield (key, GroupByGenerator(lis, beginning, l - beginning))
```
mrcslws: No need to check `l > 0` again, just keep the next line indented and piggyback on the previous check. And I think it's a little bit easer to read if we just use `len(lis)` rather than having this `l` variable.

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,161 @@
...
+                  generates the values in the group.
+  """
+  length = 1
+  beginning = 0
+  l = len(lis)
+
+  if l > 0:
+    key = fn(lis[0])
+    for i in xrange(1, len(lis)):
+      val = fn(lis[i])
```
mrcslws: Calling these `key` and `val` can be deceptive since these aren't key-value pairs. Something else would be good, e.g. `key` +`nextKey` or `key` + `key2`.

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,161 @@
...
+        key = val
+      else:
+        length += 1
+
+  if l > 0: # yield last group
+    yield (key, GroupByGenerator(lis, beginning, l - beginning))
+
+
+def groupByN(*args):
+  """ A utility function wrapper based on groupby in itertools. Allows to
```
mrcslws: I don't think it's very clear to call this a "utility function wrapper". It's groupby, but for multiple lists.

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,161 @@
...
+          nextList[i] = None
+
+    # no more values to process in any of the generators
+    if all([entry is None for entry in nextList]):
+      break
+
+    # find the smallest next group and all lists that have an element in it
+    argIndices = [] # array of indices corresponding to the lists that have at
+                    # least one element that is a member of the minKeyVal group
+    minKeyVal = float("inf")
```
mrcslws: This argIndices list is kind of hard to read.

Instead, maybe do

~~~python
minKeyVal = min(nextVal[0] for nextVal in nextList
                if nextVal is not None)
~~~

Then further down just check

~~~python
for i in xrange(n):
  if nextList[i] is not None and nextList[i][0] == minKeyVal:
    ...
  else:
    ...
~~~

Do you see any downsides with that approach?

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,161 @@
...
+    for i in xrange(n):
+      if advanceList[i]:
+        try:
+          nextList[i] = generatorList[i].next()
+        except StopIteration:
+          nextList[i] = None
+
+    # no more values to process in any of the generators
+    if all([entry is None for entry in nextList]):
+      break
```
mrcslws: nit: we might as well do `all(entry is None for entry in nextList)` without the `[]` to avoid creating a list.

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,161 @@
...
+                    objects that are a member of the current group that is being
+                    iterated over in the nth list passed in. Note that this
+                    is a generator and a n+1 dimensional tuple is yielded for
+                    every group.
+
+  Notes: Read up on groupby here:
+         https://docs.python.org/dev/library/itertools.html#itertools.groupby
+
+         itertools.groupby was not used and the justification can be seen
+         here https://github.com/numenta/nupic/pull/3254#discussion_r74164564
```
mrcslws: The docstring for `groupby` should answer this question. We probably shouldn't link to a github discussion. Better to describe it succinctly here.

Fundamentally, I think `itertools.groupby` is more constrained because it has to support arbitrary iterables. You couldn't pass `(i for i in xrange(10))` into our custom `groupby`, but you can pass it into `itertools.groupby`.

And I think now that we're not evaluating the whole groupby result immediately in `groupByN`, we could use the built-in `itertools.groupby`, though I'm not necessarily suggesting this. The benefits we're getting are:

- Support for `len`
- Support for arbitrary evaluation of the nested generators returned by `groupByN` (though we're not currently using this)

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,161 @@
...
+        beginning = i
+        key = val
+      else:
+        length += 1
+
+  if l > 0: # yield last group
+    yield (key, GroupByGenerator(lis, beginning, l - beginning))
+
+
+def groupByN(*args):
```
mrcslws: The name `groupByN` doesn't feel right to me. It makes it sound like it's grouping by N. Lots of people reading this name will think the function takes in an "n" parameter... maybe they'll think it's essentially a [partition](https://clojuredocs.org/clojure.core/partition) function.

It's identical to `itertools.groupby` when given one sequence (arity 2), and it just supports other arities (4, 6, ...). I think `groupby2` would be a good name. The more we piggyback on people's knowledge about `itertools.groupby`, the better. The goal is to make this idea of a "groupby" seem as simple as it actually is.

src/nupic/support/group_by.py
mrcslws: If we can use `itertools.groupby`, I think we should. Then we'll support generators, which is a likely scenario given how often we use list comprehension.

Since we won't have `len` support, this means the outer function "groupByN" will have to return `None` rather than empty generators when a sequence doesn't contain a key. I think that would make sense.

My impression is that itertools.groupby could be swapped into the current code, then we'll just have to switch away from relying on `len` in the calling code and instead check for `None`.

src/nupic/research/temporal_memory.py
```diff
@@ -164,15 +163,15 @@ def compute(self, activeColumns, learn=True):
...
+    for columnData in groupby2(activeColumns, identity,
                                self.activeSegments, segToCol,
                                self.matchingSegments, segToCol):
       (column,
        activeColumns,
        activeSegmentsOnCol,
        matchingSegmentsOnCol) = columnData
-      if len(activeColumns):
-        if len(activeSegmentsOnCol) != 0:
+      if not activeColumns is None:
```
mrcslws: For these it's better to do `is not None`.

I just checked and it's in PEP-8: "Use `is not` operator rather than `not ... is`" https://www.python.org/dev/peps/pep-0008/

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,89 @@
...
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+from itertools import groupby
+
+
+def groupby2(*args):
+  """ An extension to groupby in itertools. Allows to
```
mrcslws: nit: reformat the paragraph now that this line is narrower

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,89 @@
...
+        try:
+          nextList[i] = generatorList[i].next()
+        except StopIteration:
+          nextList[i] = None
+
+    # no more values to process in any of the generators
+    if all(entry is None for entry in nextList):
+      break
+
+    #the miniumum key value in the nextList
```
mrcslws: Put a space after `#`, and typo: "minimum"

src/nupic/research/temporal_memory.py
```diff
@@ -362,35 +261,37 @@ def activatePredictedColumn(connections, excitedColumn, learn,
...
+    @param column              (int)    Index of bursting column
     @param connections         (Object) Connections instance for the tm
-    @param excitedColumn       (dict)   Excited Column instance from
-                                        excitedColumnsGenerator
-    @param learn               (bool)   Whether or not learning is enabled
     @param initialPermanence   (float)  Initial permanence of a new synapse.
+    @param learn               (bool)   Whether or not learning is enabled
+    @param matchingSegments    (list)   A list of SegmentOverlap objects for the
+                                        column compute is operating on that are
+                                        matching
```
mrcslws: This is now an iterable, and the value is None sometimes. It's worth documenting that. Same with the other method docstrings.

mrcslws:
I played with the `nupic.research` code that still uses Connections (`faulty_temporal_memory`, `phases_temporal_memory`, and by extension the `ExtendedTemporalMemory` and its unit/integration tests).

We'll have to make lots of updates to that code if we want it to keep working with this change.

I think we should go ahead and copy-paste the old code into `connections_phases.py` in `nupic.research` rather than trying to keep all these old implementations working with the latest `connections`.

mrcslws:
Done reviewing, very nice work! 👍

scottpurdy:
👍

src/nupic/support/group_by.py
scottpurdy: `lis`? Does not seem like a good name to me. Is it a Python sequence?

src/nupic/support/group_by.py
```diff
@@ -0,0 +1,161 @@
...
+  def __len__(self):
+    return self.length - self.index
+
+
+def groupby(lis, fn):
+  """ A custom implementation of itertools.groupby that doesn't reuse the
+      generator objects, allowing for groupByN to not have to capture the
+      group in a list.
+
+  @param lis (list) the list to perform the groupby on
```
scottpurdy: Does this have to be a list? Can it be any sequence type?

mrcslws: I think you're commenting on old code. The latest is here: https://github.com/numenta/nupic/blob/master/src/nupic/support/group_by.py

scottpurdy: K thx, I was going through old email and saw unresponded-to-comments making me think it was current code.

PR #3262, opened 2016-08-13
NUP-2299 Added platform-conditional dependency on pycapnp==0.5.8 using PEP-508

### LINKED ISSUE #3263: Add conditional dependency to install pycapnp on Linux and OS X, but not Windows
We're going to install pycapnp==0.5.8 when installing nupic on OS X and Linux. pycapnp is used in the new serialization APIs, except on Windows, where pycapnp doesn't install/build presently.

vitaly-krugl:
Fixes #3263

### DIFF
--- a/ci/bamboo/build-osx.sh
+++ b/ci/bamboo/build-osx.sh
@@ -9,6 +9,12 @@ export PYTHONPATH=${HOME}/Library/Python/2.7/lib/python/site-packages:${PYTHONPA
 # Install pip
 python ci/bamboo/get-pip.py --user --ignore-installed
 
+# Upgrade setuptools (for PEP-508 support used in extras_require); also wheel
+pip install --user --upgrade --ignore-installed setuptools wheel
+python -c 'import pip; print "pip version=", pip.__version__'
+python -c 'import setuptools; print "setuptools version=", setuptools.__version__'
+python -c 'import wheel; print "wheel version=", wheel.__version__'
+
 # Build installable python packages
 python setup.py bdist_wheel
 
--- a/ci/bamboo/build-osx.sh
+++ b/ci/bamboo/build-osx.sh
@@ -6,14 +6,8 @@ set -o xtrace
 export PATH=${HOME}/Library/Python/2.7/bin:${PATH}
 export PYTHONPATH=${HOME}/Library/Python/2.7/lib/python/site-packages:${PYTHONPATH}
 
-# Install pip
-python ci/bamboo/get-pip.py --user --ignore-installed
-
-# Upgrade setuptools (for PEP-508 support used in extras_require); also wheel
-pip install --user --upgrade --ignore-installed setuptools wheel
-python -c 'import pip; print "pip version=", pip.__version__'
-python -c 'import setuptools; print "setuptools version=", setuptools.__version__'
-python -c 'import wheel; print "wheel version=", wheel.__version__'
+# Install specific versions of pip, setuptools, and wheel
+./ci/bamboo/install-pip-setuptools-wheel.sh --user
 
 # Build installable python packages
 python setup.py bdist_wheel
--- a/ci/bamboo/install-pip-setuptools-wheel.sh
+++ b/ci/bamboo/install-pip-setuptools-wheel.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2016, Numenta, Inc.  Unless you have purchased from
+# Numenta, Inc. a separate commercial license for this software code, the
+# following terms and conditions apply:
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero Public License for more details.
+#
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# -----------------------------------------------------------------------------
+
+# Wrapper around get-pip.py that sets specific versions of pip, setuptools, and
+# wheel for use on *nix platforms.
+#
+# ARGS: the caller may pass additional args for get-pip.py. This is primarily
+#       intended to facilitate passing of --user when needed, which is
+#       omitted by default.
+
+set -o errexit
+set -o xtrace
+
+# Tool requirements:
+#   Fleshed out PEP-508 support (Dependency Specification)
+_PIP_VER="8.1.2"
+_SETUPTOOLS_VER="25.2.0"
+_WHEEL_VER="0.29.0"
+
+_GET_PIP="$(cd ${BASH_SOURCE%/*} && pwd)/get-pip.py"
+
+python ${_GET_PIP} "$@" --ignore-installed \
+  pip==${_PIP_VER} \
+  setuptools==${_SETUPTOOLS_VER} \
+  wheel==${_WHEEL_VER}
+
+python -c 'import pip; print "pip version=", pip.__version__'
+python -c 'import setuptools; print "setuptools version=", setuptools.__version__'
+python -c 'import wheel; print "wheel version=", wheel.__version__'
--- a/ci/travis/after_success-release-linux.sh
+++ b/ci/travis/after_success-release-linux.sh
@@ -25,11 +25,11 @@ echo "Running after_success-release.sh..."
 echo
 
 echo "Installing boto..."
-pip install boto --user || exit
+pip install boto || exit
 echo "Installing wheel..."
-pip install wheel --user || exit
+pip install wheel || exit
 echo "Installing twine..."
-pip install twine --user || exit
+pip install twine || exit
 
 echo "Creating distribution files..."
 # This release build creates the source distribution. All other release builds
--- a/ci/travis/before_install-linux.sh
+++ b/ci/travis/before_install-linux.sh
@@ -32,21 +32,18 @@ if [ $CC == 'gcc' ]; then
     export CXX='g++-4.8'
 fi
 
-echo ">>> Installing nupic-linux64..."
-git clone https://github.com/numenta/nupic-linux64.git
-(cd nupic-linux64 && git reset --hard 99863c7da8b923c57bb4e59530ab087c91fd3992)
-source nupic-linux64/bin/activate
+# Upgrade setuptools (for PEP-508 support used in extras_require)
+pip install --upgrade --ignore-installed setuptools
 
-# Upgrade the version of pip included with nupic-linux64
-# TODO: remove after nupic-linux64 has been updated
-pip install --upgrade --install-option="--prefix=`pwd`/nupic-linux64" pip
-pip --version
+pip install --upgrade --ignore-installed pip
+
+pip install wheel
+
+python -c 'import pip; print "pip version=", pip.__version__'
+python -c 'import setuptools; print "setuptools version=", setuptools.__version__'
+python -c 'import wheel; print "wheel version=", wheel.__version__'
 
 pip uninstall numpy --yes
-
-# Assuming pip 1.5.X is installed.
-echo "pip install wheel --user"
-pip install wheel --user -q
 
 # Fetch nupic.core build
 export NUPIC_CORE_COMMITISH=`python -c "execfile('.nupic_modules'); print NUPIC_CORE_COMMITISH"`
@@ -57,7 +54,7 @@ tar xzf "nupic_core-${NUPIC_CORE_COMMITISH}-linux64.tar.gz"
 ls home/travis/build/numenta/nupic.core/bindings/py/dist/wheels
 
 # Install nupic.bindings and dependencies from wheels
-pip install --user --no-index --find-links=home/travis/build/numenta/nupic.core/bindings/py/dist/wheels nupic.bindings
+pip install --no-index --find-links=home/travis/build/numenta/nupic.core/bindings/py/dist/wheels nupic.bindings
 
 # Workaround for multiprocessing.Queue SemLock error from run_opf_bechmarks_test.
 # See: https://github.com/travis-ci/travis-cookbooks/issues/155
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,6 +93,11 @@ deploy:
         - "$CC == gcc-4.8"
 
 before_install:
+  # Create a new virtualenv to enable upgrade of setuptools and pip;
+  # see https://github.com/travis-ci/travis-ci/issues/4194
+  - virtualenv --no-site-packages venv
+  - source venv/bin/activate
+
   # Turned pylint checking off until https://github.com/numenta/nupic/issues/3147 is fixed.
   # - ". ./ci/travis/run_pylint.sh"
   - "if [ ${TRAVIS_OS_NAME:-'linux'} = 'linux' ]; then . ./ci/travis/before_install-linux.sh; fi"
--- a/README.md
+++ b/README.md
@@ -20,8 +20,9 @@ NuPIC binaries are available for:
 The following dependencies are required to install NuPIC on all operating systems.
 
 - [Python 2.7 & development headers](https://docs.python.org/devguide/setup.html#build-dependencies)
-- [pip](https://pip.pypa.io/en/stable/installing/)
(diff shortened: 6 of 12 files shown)


oxtopus:
👍 provided https://ci.numenta.com/browse/NUP-PY81/latest passes and whatever travis-ci issues are resolved.

Thanks!

oxtopus:
Seems to be a problem on OS X:

```
+ python setup.py bdist_wheel
error in nupic setup command: Invalid environment marker: platform_system=='Linux' or platform_system=='Darwin'
```

vitaly-krugl:
@oxtopus : I'll work through this. The problem is that pip and setuptools installed by travis are too old. Travis sets up its python packages in a strange way that makes it difficult to upgrade pip and setuptools, because sys.path lists their eggs explicitly at absolute paths. Even though the upgrade claims that it succeeded, the ones that get used are the old ones!

I think I need to use virtualenv myself to get the environment and package versions that I need to make the build work. This works fine with newer pip/setuptools.

I am also struggling to find meaningful documentation on travis's .travis.yml syntax. They are either  making it difficult to find or it just doesn't exist.

oxtopus:
The error I quoted was from the bamboo run at  https://ci.numenta.com/browse/NUP-PY81 in which a recent version of setuptools is properly installed.  If it is a version incompatibility, we should document that and ensure that specific version gets installed.

ci/travis/after_success-release-linux.sh
```diff
@@ -25,11 +25,11 @@ echo "Running after_success-release.sh..."
 echo
 
 echo "Installing boto..."
-pip install boto --user || exit
```
vitaly-krugl: virtualenv activation (see changes in .travis.yml) doesn't allow `--user`

ci/travis/before_install-linux.sh
```diff
@@ -32,21 +32,18 @@ if [ $CC == 'gcc' ]; then
     export CXX='g++-4.8'
 fi
 
-echo ">>> Installing nupic-linux64..."
-git clone https://github.com/numenta/nupic-linux64.git
```
vitaly-krugl: The nupic-linuxXX and nupic-darwinXX environments have gotten hopelessly out of date, and are not needed for the build to succeed.

CC @scottpurdy

vitaly-krugl:
Please review. This addresses the P2 issue NUP-2299 "Bamboo CI pipeline red"

ci/bamboo/build-osx.sh
```diff
@@ -9,6 +9,12 @@ export PYTHONPATH=${HOME}/Library/Python/2.7/lib/python/site-packages:${PYTHONPA
 # Install pip
 python ci/bamboo/get-pip.py --user --ignore-installed
 
+# Upgrade setuptools (for PEP-508 support used in extras_require); also wheel
+pip install --user --upgrade --ignore-installed setuptools wheel
+python -c 'import pip; print "pip version=", pip.__version__'
```
oxtopus: Please remove these lines, or add them to `ci/bamboo/build-linux.sh` for consistency.

oxtopus:
To reiterate, if there's a specific version requirement on setuptools, I'd like to see it documented in this changeset.  You put a lot of work into making the following changes to setup.py work consistently between travis and bamboo -- going above and beyond what was required -- and it's likely someone will encounter the same problems, too, without that version or greater.

```diff
 +    extras_require = {
 +      # Default requirement based on system type
 +      ":platform_system=='Linux' or platform_system=='Darwin'":
 +        ["pycapnp==0.5.8"],
 +
 +      # Superseded by platform_system-conditional requirement, but keeping
 +      # empty extra for compatibility as recommended by setuptools doc.
 +      "capnp": [],
 +    },
```

oxtopus:
In retrospect, `pip install --upgrade --ignore-installed setuptools wheel` should never have made it into `ci/bamboo/build-linux.sh` (I put it there.  I'm sorry).  In the bamboo builds, `get-pip.py` is used to install `pip`, `setuptools`, and `wheel` -- see https://pip.pypa.io/en/stable/installing/#installing-with-get-pip-py for official recommended installation technique.  We're vendoring in `get-pip.py` (see https://github.com/numenta/nupic/tree/master/ci/bamboo) for a number of reasons, one of which is to version lock a known, good version of `pip`, `setuptools`, and `wheel`.  Calling `pip install --upgrade --ignore-installed setuptools wheel` defeats the purpose of bundling `get-pip.py`.

Instead, if we need to update `pip`, `setuptools`, or `wheel`, we should update `get-pip.py` (and document the versions for anyone not using the correct installation method).

vitaly-krugl:
@oxtopus, where is the current location that documents the versions of pip, setuptools, and wheel?

oxtopus:
https://github.com/numenta/nupic/blob/master/README.md#dependencies, https://github.com/numenta/nupic/blob/master/external/common/requirements.txt, and https://github.com/numenta/nupic/blob/master/DEPENDENCIES.md are all means of conveying dependencies in nupic, none of which currently specify any specific version of setuptools.  Since there is now a dependency on a specific version of setuptools, we need to make that clear in the documentation as as been done for other dependencies where it's known that a specific version (or later) is required.

Personally, modifying `setup.py` to assert on a specific or later version, or a try/except block to catch the failure and suggest that the user install the specific version or later of setuptools would be most effective, but is not necessary in this case -- as long as it's documented.

ci/bamboo/install-pip-setuptools-wheel.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
```
vitaly-krugl: Shared script for installing specific versions of pip, setuptools, and wheel on *nix (Linux and OS X) builds.

vitaly-krugl:
@oxtopus, regarding
> Instead, if we need to update pip, setuptools, or wheel, we should update `get-pip.py`

We don't want to modify `get-pip.py`. Instead, since `get-pip.py` accepts all command-line options that you can pass to `pip`, we simply pass the required packages (pip, setuptools, wheel) along with their versions to `get-pip.py` using a script that is shared between the builds.

vitaly-krugl:
This is ready for code review once again. I removed upgrading of pip/setuptools/wheel, installed explicit version numbers of those packages, and updated version requirements in the README.

vitaly-krugl:
Incidentally, the goal for the chosen approach is to isolate the logic of deciding **whether** and **which version** of pycapnp to install in a single location, thus minimizing the likelihood of bugs and simplifying maintenance.

```
 +    extras_require = {
 +      # Default requirement based on system type
 +      ":platform_system=='Linux' or platform_system=='Darwin'":
 +        ["pycapnp==0.5.8"],
 +
 +      # Superseded by platform_system-conditional requirement, but keeping
 +      # empty extra for compatibility as recommended by setuptools doc.
 +      "capnp": [],
 +    },
```

vitaly-krugl:
PR builds were successful at 067978623f0da3525bea06ec11d2b3ea6ba616a1 on Travis, AppVeyor, and Bamboo (https://ci.numenta.com/browse/NUP-PY81-20/log)

oxtopus:
Thanks!  I appreciate the effort you went through here.  `install-pip-setuptools-wheel.sh` adds an unnecessary amount of complexity, but that's a discussion we can take offline and it's not so important that it's worth holding back a solution that would otherwise fix a broken pipeline. 👍 :shipit:

vitaly-krugl:
@oxtopus, I would not wish to merge a change that, according to you,
> adds an unnecessary amount of complexity

What do you recommend in its place that would not add an "unnecessary amount of complexity"?

oxtopus:
I'd recommended earlier (here, or elsewhere) that all you need to do is add `[capnp]` to the `pip install` command.  For example, the line https://github.com/vitaly-krugl/nupic/blob/a7d3c724911332c8702ed7daaa1f768709faed93/ci/bamboo/build-osx.sh#L17 becomes:

`pip install --user --ignore-installed -f wheelhouse/ dist/nupic-`cat VERSION`*.whl[capnp]`

Anecdotal from my mac:

```
Austins-MacBook-Pro-6:Downloads amarshall$ pip install --user nupic-0.5.4-py2-none-any.whl[capnp]
Processing ./nupic-0.5.4-py2-none-any.whl
Requirement already satisfied (use --upgrade to upgrade): unittest2==0.5.1 in /Users/amarshall/Library/Python/2.7/site-packages (from nupic==0.5.4)
Requirement already satisfied (use --upgrade to upgrade): asteval==0.9.1 in /Users/amarshall/Library/Python/2.7/site-packages (from nupic==0.5.4)
Requirement already satisfied (use --upgrade to upgrade): pytest-cov==1.6 in /Users/amarshall/Library/Python/2.7/site-packages (from nupic==0.5.4)
Requirement already satisfied (use --upgrade to upgrade): validictory==0.9.1 in /Users/amarshall/Library/Python/2.7/site-packages (from nupic==0.5.4)
Requirement already satisfied (use --upgrade to upgrade): numpy==1.9.2 in /Users/amarshall/Library/Python/2.7/lib/python/site-packages (from nupic==0.5.4)
Collecting coverage==3.7.1 (from nupic==0.5.4)
Requirement already satisfied (use --upgrade to upgrade): DBUtils==1.1 in /Users/amarshall/Library/Python/2.7/site-packages (from nupic==0.5.4)
Requirement already satisfied (use --upgrade to upgrade): PyYAML==3.10 in /Users/amarshall/Library/Python/2.7/site-packages (from nupic==0.5.4)
Requirement already satisfied (use --upgrade to upgrade): mock==1.0.1 in /Users/amarshall/Library/Python/2.7/site-packages (from nupic==0.5.4)
Collecting nupic.bindings==0.4.4 (from nupic==0.5.4)
  Using cached nupic.bindings-0.4.4-cp27-none-macosx_10_11_intel.whl
Requirement already satisfied (use --upgrade to upgrade): ordereddict==1.1 in /Users/amarshall/Library/Python/2.7/site-packages (from nupic==0.5.4)
Requirement already satisfied (use --upgrade to upgrade): PyMySQL==0.6.2 in /Users/amarshall/Library/Python/2.7/site-packages (from nupic==0.5.4)
Requirement already satisfied (use --upgrade to upgrade): pyproj==1.9.3 in /Users/amarshall/Library/Python/2.7/site-packages (from nupic==0.5.4)
Requirement already satisfied (use --upgrade to upgrade): python-dateutil==2.1 in /Users/amarshall/Library/Python/2.7/lib/python/site-packages/python_dateutil-2.1-py2.7.egg (from nupic==0.5.4)
Requirement already satisfied (use --upgrade to upgrade): pytest==2.5.1 in /Users/amarshall/Library/Python/2.7/lib/python/site-packages (from nupic==0.5.4)
Requirement already satisfied (use --upgrade to upgrade): psutil==1.0.1 in /Users/amarshall/Library/Python/2.7/site-packages (from nupic==0.5.4)
Requirement already satisfied (use --upgrade to upgrade): pytest-xdist==1.8 in /Users/amarshall/Library/Python/2.7/site-packages (from nupic==0.5.4)
Requirement already satisfied (use --upgrade to upgrade): prettytable==0.7.2 in /Users/amarshall/Library/Python/2.7/site-packages (from nupic==0.5.4)
Collecting pycapnp==0.5.5; extra == "capnp" (from nupic==0.5.4)
Requirement already satisfied (use --upgrade to upgrade): cov-core>=1.6 in /Users/amarshall/Library/Python/2.7/lib/python/site-packages (from pytest-cov==1.6->nupic==0.5.4)
Requirement already satisfied (use --upgrade to upgrade): six in /Users/amarshall/Library/Python/2.7/site-packages (from python-dateutil==2.1->nupic==0.5.4)
Requirement already satisfied (use --upgrade to upgrade): py>=1.4.19 in /Users/amarshall/Library/Python/2.7/site-packages (from pytest==2.5.1->nupic==0.5.4)
Requirement already satisfied (use --upgrade to upgrade): execnet>=1.0.8 in /Users/amarshall/Library/Python/2.7/site-packages (from pytest-xdist==1.8->nupic==0.5.4)
Installing collected packages: coverage, nupic.bindings, pycapnp, nupic
Successfully installed coverage-4.0a1 nupic-0.5.5.dev0 nupic.bindings-0.4.5.dev0 pycapnp-0.5.5
```

Note the `Collecting pycapnp==0.5.5; extra == "capnp" (from nupic==0.5.4)` and `Installing collected packages: coverage, nupic.bindings, pycapnp, nupic
Successfully installed coverage-4.0a1 nupic-0.5.5.dev0 nupic.bindings-0.4.5.dev0 pycapnp-0.5.5` from the output.

I do like the approach, but I'm somewhat conflicted about it because A) it requires a new version of `setuptools`, and B) `get-pip.py` is being used in a way I hadn't anticipated to add support for the newer `setuptools`.

scottpurdy:
By the way, we do need to update setuptools to a fairly recent version. I'm not sure the exact version but some dependencies require something fairly recent (and aren't necessarily able to update it themselves beforehand due to the nature of it):

https://github.com/jparyani/pycapnp#common-problems

And I think we should consider alternatives before we have copies of `get-pip.py` sprinkled throughout all of our repos.

vitaly-krugl:
@oxtopus, the problem with relying on something like `nupic-0.5.4-py2-none-any.whl[capnp]` that you recommended is that it spreads nupic's dependency information outside of the nupic package. When relying on the `...[capnp]` feature selection, all of our build scripts (as well as 3rd party users) would then need to know something unnecessary about the dependencies of the nupic package and take explicit action based on it. So, all of our ci/bamboo nupic build scripts would need to know when to pass it and when not to, as would the build scripts in nupic.regression(1) and other build scripts and users.

(1)Note that nupic.regression build presently uses the old way, but soon won't have to.

Regarding your feedback
> get-pip.py is being used in a way I hadn't anticipated to add support for the newer setuptools.

I didn't know about that capability either until I read the documentation and found that it's fully supported. So, it made perfect sense to use it. In https://pip.pypa.io/en/stable/installing/#installing-with-get-pip-py:
> Additionally, get-pip.py supports using the [pip install options](https://pip.pypa.io/en/stable/reference/pip_install/#pip-install-options) and the [general options](https://pip.pypa.io/en/stable/reference/pip/#general-options).

oxtopus:
I disagree with the conclusions you've made here.  The functionality that allows for specifying `[capnp]` in the pip command has been there for a long time.  I merely suggested that using the status quo functionality is better than duplicating a version number in a build script.

To @scottpurdy's point, there is some duplication between nupic.core and nupic and this changeset further diverges the two projects with respect to how the `pycapnp` dependency is handled.

re: `get-pip.py`, I suspect it has a short-lived future.  I know @scottpurdy wants it gone, so having a disagreement about it certainly doesn't help, but here goes the rationale: `get-pip.py` was included because it is the recommended way of installing `pip` according to official documentation, and it works consistently across platforms without having to keep track of platform-specific installation techniques.  Bundling it in the repo was done to avoid intermittent failures or security issues related to downloading it every time we run an automated build.

Also, just because something is in the documentation, doesn't mean it makes sense in context.  If we're going to download specific versions of pip, setuptools, and wheel in every build, that certainly adds support to the argument for its removal (i.e., it's redundant).

vitaly-krugl:
Other than this, I am going to refrain from adding further comments in this now-merged PR.

The way that nupic.core manages the pycapnp dependency is not ideal. I consider this nupic PR a step in the right direction. There is no need to be concerned about diverging from nupic.core's legacy ways to arrive at something simpler/better. nupic.core could also benefit from a similar simplification.

By the way, this is what the pycapnp picture looks like right now in nupic.core, with just about every build script burdened with intimate knowledge about the specific dependency on `pycapnp==0.5.8`. Also, 3rd party and in-house developers need to be aware of the need to install it and of the correct version to use in order to build nupic.core, and all of these things need to be in sync.
```
~/nta/nupic.core (master)$ git grep 'pycapnp=='
Dockerfile:        pycapnp==0.5.8 \
README.md:    pip install pycapnp==0.5.8
bindings/py/setup.py:    extras_require = {"capnp": ["pycapnp==0.5.8"]},
ci/bamboo/build-linux-debug.sh:    pycapnp==0.5.8 \
ci/bamboo/build-linux-release.sh:    pycapnp==0.5.8 \
ci/bamboo/build-manylinux-prototype.sh:${PYBIN}/pip install pycapnp==0.5.8
ci/bamboo/build-osx.sh:    pycapnp==0.5.8 \
ci/travis/before_install-linux.sh:pip install --user pycapnp==0.5.8 || exit
ci/travis/before_install-osx.sh:pip install pycapnp==0.5.8 --user || exit
```

Compare this to the current nupic:
```
~/nta/nupic (master)$ git grep 'pycapnp=='
Dockerfile:        pycapnp==0.5.8 \
setup.py:        ["pycapnp==0.5.8"],
```
... and the one in Dockerfile is technically no longer necessary, except as a build performance speed-up as it avoids the build time of pycapnp.

rhyolight:
@vitaly-krugl @oxtopus I've created a topic in the forum committer lounge for you to continue this discussion (if you wish).

Issue #1450, opened 2014-10-16
Platform-Independent Model Serialization


rhyolight:
This is a super-super-issue, tracking both [Platform-Independent Model Serialization Phase 1](https://github.com/numenta/nupic/issues/1435) and [Platform-Independent Model Serialization Phase 2](https://github.com/numenta/nupic/issues/1449) across the **Alpha** and **Beta** milestones. 

We plan to use [Cap'n Proto](http://kentonv.github.io/capnproto/).

---
- [x] [Platform-Independent Model Serialization Phase 1](https://github.com/numenta/nupic/issues/1435)
- [ ]  [Platform-Independent Model Serialization Phase 2](https://github.com/numenta/nupic/issues/1449)

PR #3266, opened 2016-08-16
Correcting spatial pooler terminology to mini column instead of cortical column

### LINKED ISSUE #3259: Incorrect terminology in spatial_pooler.py
The file spatial_pooler.py contains two helper classes, CorticalColumns and BinaryCorticalColumns. These names are incorrect as these should refer to mini columns.  There are also numerous references to cortical columns in the comments.

The term "cortical column" means something completely different in neuroscience. This issue is particularly relevant since in the new theory work, we rely on the correct meaning of "cortical column".

This should be an easy change for someone to implement. Other than this file and some tests, nothing else in NuPIC uses these helper functions.

arhik:
Incorporated changes to the spatial terminology as mentioned which Fixes #3259.

### DIFF
--- a/src/nupic/research/spatial_pooler.py
+++ b/src/nupic/research/spatial_pooler.py
@@ -43,7 +43,7 @@ class InvalidSPParamValueError(ValueError):
 
 
 
-class _SparseMatrixCorticalColumnAdapter(object):
+class _SparseMatrixMiniColumnAdapter(object):
   """ Many functions in SpatialPooler operate on a columnIndex but use an
   underlying storage implementation based on a Sparse Matrix in which cortical
   columns are represented as rows.  This can be confusing to someone trying to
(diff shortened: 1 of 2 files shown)


oxtopus:
👍

rhyolight:
@scottpurdy Is this going to break our current serialization?

scottpurdy:
@rhyolight I don't know! Not sure if we have tests covering that for this implementation.

scottpurdy:
I can't find a test that covers it. Seems very likely that it would break the pickle serialization.

rhyolight:
> Seems very likely that it would break the pickle serialization.

Yes, it certainly seems like it would. I guess I should not have marked #3259 as ready for work. :(

rhyolight:
I'm sorry @arhik, but we can't merge this PR because it will change the serialization signature of the spatial pooling, which is an incompatible change with all existing versions of NuPIC. If we merged, no one would be able to de-serialize their models they saved before the change. I did not think of this when we were chatting on the original issue. This PR could be easily merged once #1450 is done. But I don't want to keep it open at this point until we're ready to merge it.

scottpurdy:
Here is one place you can learn more about the limits of pickling (and why we will be removing it soon):
http://stackoverflow.com/questions/16269071/python-pickle-dealing-with-updated-class-definitions

PR #3289, opened 2016-08-22
Only activate cols with overlap with input vector.

### LINKED ISSUE #3287: Winner columns chosen with no overlap with input (local inhibition)
> See discussion of this problem at https://discourse.numenta.org/t/sp-columns-activating-without-input-when-using-topology/1267

Given the condition for selecting winner columns during the local inhibition step here:

https://github.com/numenta/nupic/blob/f1dd3ca676891f7370cf5404b3bbe432ac503f75/src/nupic/research/spatial_pooler.py#L1471

I believe there is a bug that allows columns to be activated that have no overlap with the input vector. When the condition is updated to ensure the column overlaps with the input vector, the proper behavior is observed.

Before:

[](https://www.youtube.com/watch?v=w1t4quc1oNs)

After:

[](https://www.youtube.com/watch?v=BiMZvSbGZKs)

rhyolight:
Fixes #3287

Depends upon https://github.com/numenta/nupic.core/pull/1054

Adds a unit test to ensure that active columns selected during the local
inhibition step have overlap with the input vector.

Added additional condition to winner column selection that ensures the
column overlaps with input vector to pass test.

### DIFF
--- a/src/nupic/research/spatial_pooler.py
+++ b/src/nupic/research/spatial_pooler.py
@@ -1464,13 +1464,14 @@ class SpatialPooler(object):
     addToWinners = max(overlaps)/1000.0
     overlaps = numpy.array(overlaps, dtype=realDType)
     for i in xrange(self._numColumns):
-      maskNeighbors = self._getNeighborsND(i, self._columnDimensions, self._inhibitionRadius)
-      overlapSlice = overlaps[maskNeighbors]
-      numActive = int(0.5 + density * (len(maskNeighbors) + 1))
-      numBigger = numpy.count_nonzero(overlapSlice > overlaps[i])
-      if overlaps[i] > 0 and numBigger < numActive:
-        winners.append(i)
-        overlaps[i] += addToWinners
+      if overlaps[i] > 0:
+        maskNeighbors = self._getNeighborsND(i, self._columnDimensions, self._inhibitionRadius)
+        overlapSlice = overlaps[maskNeighbors]
+        numActive = int(0.5 + density * (len(maskNeighbors) + 1))
+        numBigger = numpy.count_nonzero(overlapSlice > overlaps[i])
+        if numBigger < numActive:
+          winners.append(i)
+          overlaps[i] += addToWinners
     return numpy.array(winners, dtype=uintType)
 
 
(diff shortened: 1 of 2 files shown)


src/nupic/research/spatial_pooler.py
```diff
@@ -1468,7 +1468,7 @@ def _inhibitColumnsLocal(self, overlaps, density):
       overlapSlice = overlaps[maskNeighbors]
       numActive = int(0.5 + density * (len(maskNeighbors) + 1))
       numBigger = numpy.count_nonzero(overlapSlice > overlaps[i])
-      if numBigger < numActive:
+      if overlaps[i] > 0 and numBigger < numActive:
```
rhyolight: This is the only functional change in this PR.

tests/unit/nupic/research/spatial_pooler_unit_test.py
```diff
@@ -1337,6 +1337,16 @@ def testInhibitColumnsLocal(self):
...
 
     density = 0.5
+    sp._numColumns = 16
+    sp._columnDimensions = numpy.array([sp._numColumns])
+    sp._inhibitionRadius = 2
+    overlaps = numpy.array([1, 1, 1, 0, 1, 1, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0], dtype=realDType)
+    active = list(sp._inhibitColumnsLocal(overlaps, density))
+    # Assert that only columns with overlap are activated
+    for index in active:
+      self.assertTrue(overlaps[index] > 0)
```
rhyolight: This is the only additional test I added.

rhyolight:
I guess my IDE removed a bunch of trailing whitespace, but other than that there is only a one-line change and a unit test.

rhyolight:
I just noticed that SP compatibility tests fail at:

https://github.com/numenta/nupic/blob/master/tests/unit/nupic/research/spatial_pooler_compatability_test.py#L202

So this is going to need a nupic.core changes as well.

src/nupic/research/spatial_pooler.py
mrcslws: When it's 0, we're never consuming the `maskNeighbors`, etc., so it'd be better to move this check up a little higher. In other words, there will be two if statements, not just one.

rhyolight:
I also verified that changes to `stimulusThreshold` still have an effect in the presence of noise in the input encoding.

rhyolight:
@mrcslws ready again

rhyolight:
I think @mrcslws has a [better solution](https://github.com/numenta/nupic/issues/3287#issuecomment-241805976).

PR #3301, opened 2016-08-26
Add a lot more scenarios to the TM perf benchmark

### LINKED ISSUE #3300: The TM benchmark doesn't cover many scenarios
Currently the TM benchmark just does 2000 steps of hotgym [1]. 

It should cover more scenarios. It should test various sequences, some that are very predictable and others that are not very predictable. And it should test the TM with various parameters, especially with various cellsPerColumn.

[1] And it's a little goofy: It gets the column SDR using a `RandomDistributedScalarEncoder` that only has `n=400`, so 1648 of the 2048 columns are never active.

mrcslws:
Fixes #3300

### DIFF
--- a/scripts/temporal_memory_performance_benchmark.py
+++ b/scripts/temporal_memory_performance_benchmark.py
@@ -20,12 +20,16 @@
 # http://numenta.org/licenses/
 # ----------------------------------------------------------------------
 
-import csv
-import time
-import numpy
-import sys
+"""
+Run various perf scenarios on the Temporal Memory.
+"""
+
 import argparse
+import csv
+import numpy
 import random
+import sys
+import time
 
 from pkg_resources import resource_filename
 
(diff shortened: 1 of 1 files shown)


scottpurdy:
Please add a module doc string and slightly more descriptive help message for arg parser.

scripts/temporal_memory_performance_benchmark.py
```diff
@@ -20,175 +20,380 @@
...
+
+def tmComputeFn(instance, encoding, activeBits):
+  instance.compute(activeBits, learn=True)
+
+
+def tpComputeFn(instance, encoding, activeBits):
+  instance.compute(encoding, enableLearn=True, computeInfOutput=True)
+
+
+def before(args, printMe):
```
scottpurdy: Name this for what it does, not where it is called from. Same for `after`

scripts/temporal_memory_performance_benchmark.py
```diff
@@ -20,175 +20,380 @@
...
-  @staticmethod
-  def _printProgressBar(completed, total, nDots):
-    def numberOfDots(n):
-      return (n * nDots) // total
-    completedDots = numberOfDots(completed)
-    if completedDots != numberOfDots(completed - 1):
-      print "\r|" + ("." * completedDots) + (" " * (nDots - completedDots)) + "|",
-      sys.stdout.flush()
 
+  def addContestant(self, constructor, paramsFn, computeFn, name):
```
scottpurdy: Can you just take an instance here and let the caller deal with params?

scripts/temporal_memory_performance_benchmark.py
```diff
@@ -20,175 +20,380 @@
 # http://numenta.org/licenses/
 # ----------------------------------------------------------------------
 
-import collections
 import csv
 import time
 import numpy
 import sys
+import argparse
```
scottpurdy: nit: alpha sort order pls

scottpurdy:
Done reviewing

scripts/temporal_memory_performance_benchmark.py
mrcslws: Then the "contestant" would only be usable once.

Also, different tests use different params.

I'm trying to think of ways this could be rearranged, but every option I can think of would require a paramsFn to be used.

PR #3303, opened 2016-08-26
[fix] output active cells

marionleborgne:
@scottpurdy Fixes https://github.com/numenta/nupic/issues/3304

@ywcui1990 This should fix the issue encountered in https://github.com/numenta/nupic.research/pull/607


ywcui1990:
What happens if computePredictedActiveCellIndices is False? It seems that  outputs["activeCells"].fill(0)
outputs["activeCells"][activeIndices] = 1

should be moved outside of the if statement. We always want activeCells being returned, don't we?

marionleborgne:
Based on the description of  `computePredictedActiveCellIndices ` (1 if *active* and predicted active indices should be computed) is seems that if this param is set to False, the expected behavior should be that neither active nor predicteActive cells should be computed.

ywcui1990:
I see, that makes sense. The PR looks good to me.

scottpurdy:
👍

PR #3293, opened 2016-08-23
Grow synapses in predicted columns, not just bursting columns

mrcslws:
Fixes #3292

This is the Python counterpart to https://github.com/numenta/nupic.core/pull/1059

This PR does a few things:

- Makes the TM learn on predicted segments. (`activatePredictedColumn`, `groupby2`)
- Reorders the TM function parameters to match the C++ ordering.
- Splits `compute` into `activateCells` and `activateDendrites` phases. The compute method still exists, it just calls these two other methods.
  - This makes it match the ETM, and it aligns with new TemporalMemory pseudocode that we're reviewing.

PR #3307, opened 2016-08-28
Abolish the SegmentOverlap. Embrace references.

### LINKED ISSUE #3306: TM: The SegmentOverlap is making the code awkward
My [recent changes to the algorithm](https://github.com/numenta/nupic.core/pull/1059) have made the [activatePredictedColumn](https://github.com/numenta/nupic/blob/18ef587aba3654440dd8db20e87e2d96894e8b4c/src/nupic/research/temporal_memory.py#L320) function pretty hard to read.

The reason it's so weird is because "overlaps" are stored alongside segments in a `SegmentOverlap` struct. So if we want the number of active potential synapses for an active segment, we have to find its corresponding matching segment. That's messy.

I created the `SegmentOverlap`, but I've never really liked it. Another problem with it is that we're also storing the overlap for every active segment (i.e. the "number of active connected synapses"), but this is never getting consumed. We only use the `.overlap` on matching segments, not active segments. So this is unnecessary memory use.

It would be much better if the TM's `activeSegments` and `matchingSegments` were just lists of `Segment` instances (not `SegmentOverlap`), and if we had a fast way to look up the "number of active potential synapses" for any segment.

(I have a C++ proof of concept working)

mrcslws:
Fixes #3306

This makes two changes to the Connections and TemporalMemory.

### Change 1: Abolish the SegmentOverlap

This problem is described in #3306 

The solution is to keep the `numActivePotentialSynapsesForSegment` list that was already being created inside of `computeActivity`. Now the TM can look up the "overlap" for any segment by simply doing `numActivePotentialSynapsesForSegment[segment.flatIdx]`.

This makes the code simpler and faster. Part of the reason it's faster is because we're not instantiating a `SegmentOverlap` class for every single active / matching segment.

Now the `activeSegments` and `matchingSegments` are just simple lists of segments.

### Change 2: Use `Segment` and `Synapse` instances for both data storage and identity

This gets rid of the `SegmentData` and the `SynapseData`. Now a `Segment` is both the type used inside of the Connections for data storage and the type used to refer to a segment. Same for the `Synapse`.

The separation between a `Segment` and a `SegmentData` makes sense in C++, where structs are passed around by value. In C++ there's an incentive to keep the Segment struct small, and it also wouldn't make sense to include large amounts of data on the `Segment` struct because changing it won't have any effect on the Connection instance's internal copy.

In Python, we pass around references (because that's what you do in all garbage-collected languages). When you pass around references, it doesn't matter how big the data structure is, and the reference always points to the most up-to-date information. Also in Python, every time you create a new instance of any class, you create something that has to be garbage-collected later, so there's a lot more overhead than in C++ where a struct is just a blob of bits, no different from any other value.

With this change, we only instantiate a `Segment` or `Synapse` when creating a new segment or synapse. Afterward we just stop instantiating stuff. When you call `segmentsForCell`, it returns existing Segment instances.

### Perf results

Before:

~~~
> python scripts/temporal_memory_performance_benchmark.py -i tm_py -t hotgym
Test: hotgym

tm_py: 24.044563s
~~~

After:

~~~
> python scripts/temporal_memory_performance_benchmark.py -i tm_py -t hotgym
Test: hotgym

tm_py: 14.601669s
~~~

Measuring in timesteps-per-second, this is ~60% faster.

## TM algorithm results

This does not affect results. It's the same TemporalMemory algorithm.

### DIFF
--- a/src/nupic/research/connections.py
+++ b/src/nupic/research/connections.py
@@ -53,10 +53,11 @@ class Segment(object):
     to be consistent after serialize / deserialize.
 
     """
-    return ((self.idx, self.cell, self._synapses, self._numDestroyedSynapses,
-             self._destroyed, self._lastUsedIteration) ==
-            (other.idx, other.cell, other._synapses, self._numDestroyedSynapses,
-             other._destroyed, other._lastUsedIteration))
+    return (self.idx == other.idx and
+            self.cell == other.cell and
+            self._numDestroyedSynapses == other._numDestroyedSynapses and
+            self._destroyed == other._destroyed and
+            self._lastUsedIteration == other._lastUsedIteration)
 
 
 
@@ -91,10 +92,10 @@ class Synapse(object):
     differences for synapse permanence.
 
     """
-    return ((self.segment.cell, self.segment.idx, self.idx,
-             self.presynapticCell) ==
-            (other.segment.cell, other.segment.idx, other.idx,
-             other.presynapticCell) and
+    return (self.segment.cell == other.segment.cell and
+            self.segment.idx == other.segment.idx and
+            self.idx == other.idx and
+            self.presynapticCell == other.presynapticCell and
             abs(self.permanence - other.permanence) < EPSILON)
 
 
--- a/src/nupic/research/connections.py
+++ b/src/nupic/research/connections.py
@@ -29,15 +29,15 @@ EPSILON = 0.00001 # constant error threshold to check equality of permanences to
 class Segment(object):
   """ Class containing minimal information to identify a unique segment """
 
-  __slots__ = ['idx', 'cell', 'data']
+  __slots__ = ['cell', 'idx', 'data']
 
-  def __init__(self, idx, cell, data):
+  def __init__(self, cell, idx, data):
     """
-       @param idx  (int) Index of the segment on the cell
        @param cell (int) Index of the cell that this segment is on
+       @param idx  (int) Index of the segment on the cell
     """
-    self.idx = idx
     self.cell = cell
+    self.idx = idx
     self.data = data
 
 
@@ -114,25 +114,6 @@ class CellData(object):
     self.numDestroyedSegments = 0
 
 
-class SegmentOverlap(object):
-  """ Class that allows tracking of overlap scores on segments """
-
-  __slots__ = ['segment', 'overlap']
-
-  def __init__(self, segment, overlap):
-    """
-       @param segment (Object) Segment object to keep track of
-       @param overlap (int)    The number of synapses on the segment that
-                               are above either the matching threshold or
-                               the active threshold
-    """
-    self.segment = segment
-    self.overlap = overlap
-
-  def __eq__(self, other):
-    return (self.segment, self.overlap) == (other.segment, other.overlap)
-
-
 
 def binSearch(arr, val):
   """ function for running binary search on a sorted list.
@@ -190,7 +171,7 @@ class Connections(object):
     segmentsList = self._cells[cell].segments
 
 
-    return [Segment(i, cell, segmentsList[i])
+    return [Segment(cell, i, segmentsList[i])
             for i in xrange(len(segmentsList))
             if not segmentsList[i].destroyed]
 
@@ -234,7 +215,7 @@ class Connections(object):
     return segment.data
 
 
-  def getSegment(self, idx, cell):
+  def getSegment(self, cell, idx):
     """ Returns a Segment object of the specified segment using data from the
         self._cells array.
 
@@ -245,7 +226,7 @@ class Connections(object):
 
     """
 
-    return Segment(idx, cell, self._cells[cell].segments[idx])
+    return Segment(cell, idx, self._cells[cell].segments[idx])
 
 
   def _leastRecentlyUsedSegment(self, cell):
@@ -268,7 +249,7 @@ class Connections(object):
         minIdx = i
         minIteration = segment.lastUsedIteration
 
-    return Segment(minIdx, cell, segments[minIdx])
+    return Segment(cell, minIdx, segments[minIdx])
 
 
   def _minPermanenceSynapse(self, segment):
@@ -295,6 +276,25 @@ class Connections(object):
     return synapses[minIdx]
 
 
+  def segmentForFlatIdx(self, flatIdx):
+    """ Get the segment with the specified flatIdx.
+
+    @param flatIdx (int) The segment's flattened list index.
+
+    @return (Segment) segment object
+    """
+    return self._segmentForFlatIdx[flatIdx]
+
+
+  def segmentFlatListLength(self):
+    """ Get the needed length for a list to hold a value for every segment's
+    flatIdx.
+
+    @return (int) Required list length
+    """
+    return self._nextFlatIdx
+
+
   def synapsesForPresynapticCell(self, presynapticCell):
     """ Returns the synapses for the source cell that they synapse on.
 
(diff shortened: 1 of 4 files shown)


cogmission:
@mrcslws 

After spending some time adjusting subtle differences in Java to get exactly the same output as with the Python version, I want to point out that the previous code had a slight functional bug that doesn't get exposed by the unit tests. 

At the end of ```Connections.computeActivity()``` the ```SegmentOverlap```(s) for both the matching and active [segments were being sorted by a "bad" key](https://github.com/numenta/nupic/blob/master/src/nupic/research/connections.py#L526).

The key: ```(s.segment.cell * self.maxSegmentsPerCell
                            + s.segment.idx)```

really was arbitrary and the "s.segment.idx" was actually **always zero** and had **no** effect!

These sorted segments are used in ```t+1``` as the entries to the new ```groupby2()``` function at the beginning of TemporalMemory.compute() - and _totally_ determine the quality and efficiency of groupby2 groupings which then determines what column's segments get marked as "active" and which don't.

At first, I instinctually grouped by the ascending columns of the cells the segments belonged to and I assumed that the above key was doing the same. I was getting superior results but not equal results (my anomaly scores were better) - which was a weird unwanted result because I was shooting for absolute parity. Then I took another look at the key and examined what it was actually producing which led to my discovery of the bug.

**Summary:**
I no longer am a fan of the "groupby2()" technique (instead preferring your first "walking technique with the ```ExcitedColumnGenerator```s; because it **hides** a potential weirdness in the selection of the active and winner cells.

If we "go" with this "groupby2" technique, please be very careful to sort the source lists at the end of ```Connections.computeActivity()``` carefully and prove to yourself that you are getting the "groupings" that allow for the best TM results?

mrcslws:
A couple quick points:

- This PR already has a big description, and **your comment has nothing to do with this PR**...
- That key function is correct. Using the flatIdx would have been incorrect and would cause bugs. If all the of the segments' `idx` is 0, that just means no cell has multiple segments. If you've ever found two segments with identical `cell` and `idx`, that's a bug. The flatIdx is a magic number that is totally uncorrelated with where the segment is. The computeActivity function, as-is, should output segments in the correct sorted order for use in groupby2.
  - That key function is equivalent to checking `a.cell < b.cell` and then `a.idx < b.idx` as the secondary index. Andrew converted this check into a single scalar check because Python magically makes sorting faster when you return an integer from your key function. It uses the `maxSegmentsPerCell` because that guarantees correct key ordering -- I thought this was a pretty clever trick on Andrew's part. (This code still exists after this PR, it's just in the TM's `activateDendrites` method)
- It sounds like the bug you're describing is located in computeActivity. This has nothing to do with `groupby2`. If there were a bug in this key function, it would also affect the `ExcitedColumnsGenerator` which is really just a specific case of `groupby2`. `groupby2` is the generalized `ExcitedColumnsGenerator`.
- The change to `groupby2` had nothing to do with speed. It was a pleasant surprise that it was faster. We changed to it because it's better code, code that describes what to do using reusable primitives. There's no longer this big advanced ExcitedColumnsGenerator thingy that people need to understand. They just have to know what a `groupby` is. "groupby" is built into multiple programming languages.

cogmission:
[continue'd here](https://discourse.numenta.org/t/abolish-segmentoverlap-pr-discussion/1308) so as not to clutter this PR with loosely related commentary 😉

src/nupic/research/connections.py
```diff
@@ -187,12 +161,9 @@ def segmentsForCell(self, cell):
     @return (list) Segment objects representing segments on the given cell
```
mrcslws: I need to change this docstring's return type. It's now a generator.

src/nupic/research/connections.py
```diff
@@ -203,38 +174,41 @@ def synapsesForSegment(self, segment):
     @return (list) Synapse objects representing synapses on the given segment
```
mrcslws: I need to change this docstring's return type. It's now a generator.

mrcslws:
Current status: I'm thinking about the Connections interface. Right now it's a little confusing having direct access to all the data, and having to know which data you can/can't access. For example, right now you're supposed to read `synapse.permanence` and `synapse.presynapticCell` and `segment.cell` but you're not supposed to read `segment.synapses`. Even if we remove destroyed synapses, it will be important that you not manually add to `segment.synapses`, and instead rely on `Connections.createSynapse` so that it can do extra bookkeeping.

I think I've made the Connections interface deceptive. It doesn't guide you to do the right thing.

**This PR is not ready for review.**

mrcslws:
Verdict: I think this is a good change, it just needs to explicitly mark which parts of the `Segment` and `Synapse` are internal state (e.g. `segment._synapses`), and it should avoid editing these objects directly (e.g. `synapse.permanence = 0.7`).

We could add a layer of indirection, using integers to refer to segments and synapses. The benefit of this would be to avoid incorrect use of the API. But I can't justify that -- using direct references to the data will always be faster. We've seen that these lookups are expensive in Python.

mrcslws:
@scottpurdy ready for review.

My comment above:

> I think I've made the Connections interface deceptive.

no longer applies. I reverted parts of my change and then squashed my commits to avoid churning the blame info.

You might prefer to review the two commits individually, rather than the full diff.

tests/unit/nupic/research/temporal_memory_test.py
```diff
@@ -210,36 +210,6 @@ def testReinforceCorrectlyActiveSegments(self):
     self.assertAlmostEqual(.42, tm.connections.dataForSynapse(is1).permanence)
 
 
-  def testNoGrowthOnCorrectlyActiveSegments(self):
```
scottpurdy: Why isn't this a valid test anymore?

scottpurdy:
To make sure I understand this, it doesn't change the algorithm, just how the segment-related data is stored?

tests/unit/nupic/research/temporal_memory_test.py
mrcslws: It stopped being valid with the recent algorithm change https://github.com/numenta/nupic/pull/3293

It was explicitly checking that we don't grow synapses on predicted columns. Now we do.

The test was flawed (my bad) -- it should be failing now, but instead it only sometimes fails, depending on the random number generator. It didn't do a good job of controlling the winner cells, so sometimes no synapse growth happens in today's code. If I were writing this test today, I'd use 1 cell per column to control the winner cells.

mrcslws:
Correct.

src/nupic/research/connections.py
```diff
@@ -26,111 +26,87 @@
...
+    self._lastUsedIteration = -1
 
 
   def __eq__(self, other):
-    return (self.idx, self.cell) == (other.idx, other.cell)
+    """ Explicitly implement this for unit testing. The flatIdx is not designed
+    to be consistent after serialize / deserialize.
+
+    """
+    return ((self.idx, self.cell, self._synapses, self._numDestroyedSynapses,
```
oxtopus: Personally, I find this difficult to read.

I prefer the following, even if it's more verbose:

```python
return (self.idx == other.idx and
        self.cell == other.cell and
        self._synapses == other._synapses and
        self._numDestroyedSynapses == other._numDestroyedSynapses and
        self._destroyed == other._destroyed and
        self._lastUsedIteration == other._lastUsedIterator)
```

or:

```python
return all((self.idx == other.idx,
            self.cell == other.cell,
            self._synapses == other._synapses,
            self._numDestroyedSynapses == other._numDestroyedSynapses,
            self._destroyed == other._destroyed,
            self._lastUsedIteration == other._lastUsedIterator))
```

src/nupic/research/connections.py
```diff
@@ -26,111 +26,87 @@
...
-    self.segment = segment
-    self.data = synapseData
-
-
-  def __eq__(self, other):
-    return ((self.idx, self.segment, self.data) ==
-            (other.idx, other.segment, other.data))
-
+    @param segment
+    (Object) Segment object that the synapse is synapsed to.
```
oxtopus: 😆

scottpurdy:
👍

PR #3322, opened 2016-09-02
Remove the notion of "destroyed" Segments / Synapses

### LINKED ISSUE #3321: TM: Remove "destroyed" Segments and Synapses
This is the Python version of https://github.com/numenta/nupic.core/issues/1064

mrcslws:
Fixes #3321

This depends on a matching C++ change: https://github.com/numenta/nupic.core/pull/1071

## TM output compatibility changes

@cogmission this section is especially for you.

This change affects the TM's output in the following ways:

1. The correct sort order for segments on a cell is by age -- oldest first. This order matters, because when choosing the best matching segment, when there's a tie between two segments we simply use the first one.
2. The correct sort order for synapses on a segment is by age -- oldest first. Similarly, ties have to be broken when choosing a weakest synapse to destroy.
3. When growing synapses, keep a sorted list of presynaptic cells, choose one randomly, remove it from the list, and repeat. We no longer do the complicated swapping in `growSynapses`.

For all of these, I moved toward a more intuitive algorithm. All you have to do is store segments in a list, always append segments to the end of the list, and remove segments from the list when you destroy them. Same with synapses.

## Changes to data storage in the Connections

Now that we don't mark Segments and Synapses as "destroyed", that means that segments and synapses don't have stable "cell -> segmentIndex -> synapseIndex" paths.

In Python, this mostly doesn't matter. At any given time, you can look up a Segment or Synapse by path, and then you can continue holding on to the object. And segments have a `flatIdx` which is stable and is essentially a path.

The order of segments and synapses still matters (see the previous section), and not having a stable "index" makes sorting harder. Rather than keeping the indices up-to-date as segments and synapses are removed, use an always increasing "ordinal". A segment or synapse's ordinal never has to be updated.

The "ordinal" is not serialized. Its information is conveyed in the order of the serialized synapses / segments.

## Perf

**Before**

~~~
python scripts/temporal_memory_performance_benchmark.py -i tm_py -t hotgym
Test: hotgym

tm_py: 15.334798s
~~~

**After**

~~~
python scripts/temporal_memory_performance_benchmark.py -i tm_py -t hotgym
Test: hotgym

tm_py: 14.610597s
~~~

So this quick test says it's ~5% faster.

### DIFF
--- a/src/nupic/research/connections.py
+++ b/src/nupic/research/connections.py
@@ -19,8 +19,8 @@
 # http://numenta.org/licenses/
 # ----------------------------------------------------------------------
 
-from collections import defaultdict
 from bisect import bisect_left
+from collections import defaultdict
 
 EPSILON = 0.00001 # constant error threshold to check equality of permanences to
                   # other floats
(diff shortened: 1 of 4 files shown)


cogmission:
One question.

>  Rather than keeping the indices up-to-date as segments and synapses are removed, use an always increasing "ordinal". A segment or synapse's ordinal never has to be updated.

Is the "ordinal" a global index given during creation? Or is the ordinal of a segment local to the cell; and likewise, the ordinal of a synapse, local to a segment?

mrcslws:
@cogmission In this code, the ordinal is global. It would also work if the ordinal was local. It's only used for comparing segments that are on the same cell, or synapses on the same segment.

cogmission:
@mrcslws  Got it. I looked at the diff and understand what you're doing there. Thanks!

cogmission:
@mrcslws Did we lose track of this? I'm asking because I was going to update once more once this goes in?

scottpurdy:
How are you testing backwards compatibility? We need some way to have confidence that old models can still be loaded. We've had some tests before but I'm not sure that any are running that cover this code.

mrcslws:
I just added some Python backward-compatibility tests. They require some manual effort, similar to the C++ approach in https://github.com/numenta/nupic.core/pull/1103

I verified that a TM serialized in `master` can by loaded by the code in this PR.

mrcslws:
@scottpurdy Alright, the checks on https://github.com/numenta/nupic/pull/3362 have passed. Let me know if you approve this PR and I'll merge it via that larger PR.

src/nupic/research/connections.py
```diff
@@ -342,47 +372,40 @@ def createSynapse(self, segment, presynapticCell, permanence):
...
+    self._nextSynapseOrdinal += 1
+    segment._synapses.add(synapse)
 
-      synapse._destroyed = False
-      segment._numDestroyedSynapses -= 1
+    self._synapsesForPresynapticCell[presynapticCell].add(synapse)
 
-      synapse.presynapticCell = presynapticCell
-      synapse.permanence = permanence
+    self._numSynapses +=1
```
scottpurdy: nit - add space

src/nupic/research/connections.py
```diff
@@ -342,47 +372,40 @@ def createSynapse(self, segment, presynapticCell, permanence):
...
-      synapse = Synapse(segment, idx, presynapticCell, permanence)
-      segment._synapses.append(synapse)
+    return synapse
 
-    self._synapsesForPresynapticCell[presynapticCell].append(synapse)
-    self._numSynapses += 1
 
-    return synapse
+  def _removeSynapseFromPresynapticMap(self, synapse):
+    presynapticSynapses = \
```
scottpurdy: this is really over 80 characters without the break?

mrcslws: Yeah, the words "presynaptic" and "synapse" are long and each appear three times in this statement. I'll change the list name to "inputSynapses".

mrcslws:
@scottpurdy FYI, I addressed your feedback in the other PR #3362 before merging.

cogmission:
Hey Scott, Just curious...

>  How are you testing backwards compatibility? We need some way to have confidence that old models can still be loaded.

Is this really necessary before 1.0 release?

PR #3325, opened 2016-09-06
Specify >= to support active development of nupic.bindings/nupic.core.

### LINKED ISSUE #3324: Specify pre-release dev version of nupic.bindings in requirements.txt
This is a companion issue to https://github.com/numenta/nupic.core/issues/1069, which I can resolve with the following patch:

```patch
diff --git a/requirements.txt b/requirements.txt
index fe3dddd..a5941a3 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,5 +17,5 @@ pyproj==1.9.3
 prettytable==0.7.2
 # versions in sync.
-nupic.bindings==0.4.6
+nupic.bindings>=0.4.6
 numpy==1.9.2
```

I believe fixing this will require modifications to release process in both projects.

oxtopus:
This fixes https://github.com/numenta/nupic/issues/3327, but I'm not sure what will happen in the next release.

### DIFF
--- a/ci/travis/before_install-linux.sh
+++ b/ci/travis/before_install-linux.sh
@@ -47,7 +47,7 @@ python -c 'import wheel; print "wheel version=", wheel.__version__'
 while read line; do
     if [[ $line == nupic.bindings* ]];
     then
-        IFS='=' read -ra ADDR <<< "$line"
+        IFS='>=' read -ra ADDR <<< "$line"
         NUPIC_BINDINGS_VERSION="${ADDR[2]}"
     fi
 done <requirements.txt
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,5 +17,5 @@ pyproj==1.9.3
 prettytable==0.7.2
 # When updating nupic.bindings, also update any shared dependencies to keep
 # versions in sync.
-nupic.bindings==0.4.6
+nupic.bindings>=0.4.6
 numpy==1.9.2
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -64,11 +64,11 @@ install:
   - for /f %%i in ('python ci\appveyor\get_nupic_bindings_version.py') do set BINDINGS_VERSION=%%i
   - for /f %%i in ('type VERSION') do set NUPIC_VERSION=%%i
 
-  - echo Binding version = %BINDINGS_VERSION%
+  - echo Binding version %BINDINGS_VERSION%
   - echo NuPIC version   = %NUPIC_VERSION%
 
   # Install the latest public release from PyPi.
-  - ps: pip install nupic.bindings==$env:BINDINGS_VERSION
+  - ps: pip install nupic.bindings$env:BINDINGS_VERSION
 
   # Check all packages installed correctly
   - cmd: pip list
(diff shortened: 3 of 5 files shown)


rhyolight:
I support this as a quick fix.

rhyolight:
We should discuss this, but it might be the right thing to do moving forward, not just a quick fix.

oxtopus:
I think it is the right thing to do moving forward.  Just need to make sure it's a matter of policy, and that it is implemented in CI.

scottpurdy:
I'm not sure which issue or PR I am suppose to discuss this on so if you prefer a different one then let me know.

The problem with `>=` is that when people `pip install nupic` then we can get into a bad state if there is a new nupic.core version that the latest nupic release is not compatible with. We have a policy that dependencies are version locked and this will break that.

So how can we resolve this issue? I have some ideas but not a concrete proposal. Can we have a quick discussion on this?

requirements.txt
```diff
@@ -17,5 +17,5 @@ pyproj==1.9.3
 prettytable==0.7.2
 # When updating nupic.bindings, also update any shared dependencies to keep
 # versions in sync.
-nupic.bindings==0.4.6
+nupic.bindings>=0.4.6
```
vitaly-krugl: I think that the idea, communicated to me by @scottpurdy IIRC, was that the new scheme would enable nupic.bindings API changes to get ahead of nupic temporarily. The `>=` might create conflicts by picking up a newer, potentially incompatible, version of nupic.bindings when installing a particular version of of nupic.

ci/travis/before_install-linux.sh
```diff
@@ -47,7 +47,7 @@ python -c 'import wheel; print "wheel version=", wheel.__version__'
 while read line; do
     if [[ $line == nupic.bindings* ]];
     then
-        IFS='=' read -ra ADDR <<< "$line"
+        IFS='>=' read -ra ADDR <<< "$line"
         NUPIC_BINDINGS_VERSION="${ADDR[2]}"
```
rhyolight: because IFS only works for one character, not a string, so when it was `=` it was splitting `foo==bar` into 3 strings.

oxtopus:
@scottpurdy please discuss at the original nupic.core [issue](https://github.com/numenta/nupic.core/issues/1069) or in [HTM Forum](https://discourse.numenta.org/t/nupic-now-has-explicit-dependency-upon-nupic-core/1353)

requirements.txt
rhyolight: I don't really think that's a big deal. This will affect developers (not users), and I think we can provide instructions on how to help developers recover. As developers, how else are we going to test nupic.core versions against nupic unless it allows it?

rhyolight:
@oxtopus does this mean anything to you? https://ci.appveyor.com/project/numenta-ci/nupic/build/0.1.0.1853

rcrowder:
@rhyolight The AV python script tries to determine the version number using a `==` [1]. An easy fix is to remove the split() and pop() from that script so that it returns the entire line. And then fix up the pip command construction [2], so that it then just does a `ps: pip install $env:BINDINGS_VERSION`

1 https://github.com/numenta/nupic/blob/master/ci/appveyor/get_nupic_bindings_version.py#L10
2 https://github.com/numenta/nupic/blob/master/appveyor.yml#L71

rhyolight:
Thanks @rcrowder that's what I thought. @oxtopus I would like to get this passing so we can merge it as a quick fix.

vitaly-krugl:
Please don't merge, I will have a simple proposal shortly.

rhyolight:
Holding off on merge. We will discuss this briefly after standup.

oxtopus:
Closing again due to stalemate.

PR #3328, opened 2016-09-08
Removed obsolete package src/nupic/bindings/

### LINKED ISSUE #3329: Remove obsolete package src/nupic/bindings/
It appears to be a redundant vestige of the former all-in-one nupic + nupic.core + nupic.bindings repository

vitaly-krugl:
Fixes #3329

Per offline discussion with @scottpurdy.

I already tested installing this change on my dev laptop via `python setup.py install` and all nupic tests passed (-i, -u, -w)


vitaly-krugl:
@scottpurdy, I was thinking about this change a bit, and it made me wonder if this directory existed in order to support the use case of running only pure-python algorithms in nupic without the real nupic.bindings? Is that even probable/possible?

scottpurdy:
> @scottpurdy, I was thinking about this change a bit, and it made me wonder if this directory existed in order to support the use case of running only pure-python algorithms in nupic without the real nupic.bindings? Is that even probable/possible? Or is it merely a vestige of the former all-in-one nupic + nupic.core + nupic.bindings repository?

Just related to the namespace packages. If we created pure-Python versions we would probably run into major problems trying to use the same package names.

scottpurdy:
@oxtopus - I think this is just left over from experimentation with namespace packages but I think the `__init__.py` in `src/nupic/` should be sufficient. Does that match your understanding?

vitaly-krugl:
All tests passed, including a local install on MacBookPro.

rhyolight:
👍

vitaly-krugl:
@scottpurdy @oxtopus: I looked into https://setuptools.readthedocs.io/en/latest/setuptools.html#namespace-packages, and indeed we only need the `__init__.py` in `src/nupic` in comibantion with the `namespace_packages = ["nupic"]` arg to `setuptools.setup()`, which we already have. So, this PR should be good to go.

On the other hand, `nupic.binings` appears to do this incorrectly: it has the `declare_namespace` call in both `bindings/py/nupic/__init__.py` and `bindings/py/nupic/bindings/__init__.py`, which makes it appear as if it's trying to define two namespace packages: `nupic` and `nupic.bindings`, but `namespace_packages=["nupic"]` is missing the corresponding `"nupic.bindings"` element. I am guessing that the nupic.bindings package intends to define only the `"nupic"` namespace package, and thus should drop the `declare_namespace` call from `bindings/py/nupic/bindings/__init__.py`.

PR #3331, opened 2016-09-08
Removed unused tie-breaker code from SP

### LINKED ISSUE #3245: Should self._tieBreaker be removed?
It appears this is no longer used?

https://github.com/numenta/nupic/blob/master/src/nupic/research/spatial_pooler.py#L330

It used to exist in the ```inhibitColumns()```  method. Is it ok to just remove it in HTM.Java? ...or was there a substitute put in place?

rhyolight:
Fixes #3245.

### DIFF
--- a/src/nupic/research/spatial_pooler.py
+++ b/src/nupic/research/spatial_pooler.py
@@ -325,12 +325,6 @@ class SpatialPooler(object):
     # non-zero.
     self._permanences = CorticalColumns(numColumns, numInputs)
 
-    # Initialize a tiny random tie breaker. This is used to determine winning
-    # columns where the overlaps are identical.
-    self._tieBreaker = numpy.array([0.01 * self._random.getReal64() for i in
-                                      xrange(self._numColumns)],
-                                    dtype=realDType)
-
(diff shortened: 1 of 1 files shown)


rhyolight:
Damnit my IDE keeps removing trailing whitespace without asking.

cogmission:
One answer when Scott was making me want to cave my head in over whitespace... :-P  

[Sublime](http://www.sublimetext.com)

rhyolight:
I have a sinking feeling about these build failures. I'm going to wait and see what https://github.com/numenta/nupic/pull/3332 does...

rhyolight:
So the answer to this is not as easy as simply removing the code I removed.

cogmission:
@rhyolight If the problem is in the integration "tm compatibility test" the problem likely is that the tie-breaker functionality is still getting initialized in the C++ SP here: https://github.com/numenta/nupic.core/blob/master/src/nupic/algorithms/SpatialPooler.cpp#L579

...and without that in both the Python and C++ SPs the RNG is getting out of sync and probably causing the compatibility test to fail?

PR #3334, opened 2016-09-08
Assign self.connections value from self.connectionsFactory()

### LINKED ISSUE #3333: Support better extensibility of TemporalMemory
As a developer, I want to subclass `TemporalMemory` and possibly `Connections` or otherwise control some aspect of the creation of the `Connections` instance in my subclass.  Currently, `TemporalMemory.__init__()` references `Connections` directly, requiring the user to either completely reimplement `__init__()` in their subclass, or rely on some `super()` trickery to fixup the `Connections` instance after the fact.

oxtopus:
Fixes #3333

... rather than direct usage of Connections constructor.  Allows better extensibility should
the user want to change some aspect of the creation of the connections instance
in a subclass


scottpurdy:
Fine with me. @mrcslws - look good to you?

mrcslws:
Looks good 👍

PR #3335, opened 2016-09-09
Ignore pre-release dev versions of nupic.bindings during installation

### LINKED ISSUE #3324: Specify pre-release dev version of nupic.bindings in requirements.txt

oxtopus:
Fixes #3324 

Attn. @scottpurdy @vitaly-krugl @rhyolight 

This causes `nupic.bindings` in `requirements.txt` to be ignored if `setup.py` detects a pre-release version of `nupic.bindings` is already installed, which is only possible by installing from source, as opposed to release.

This preserves the version locking effect of specifying `==` in `requirements.txt` without interfering with developer workflow, and prevents the inevitable confusion of the current setup.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,11 +22,12 @@
 """Installation script for Python nupic package."""
 
 import os
-import setuptools
+import pkg_resources
 import sys
 
 from setuptools import setup, find_packages, Extension
 from setuptools.command.test import test as BaseTestCommand
+
 
 
 REPO_DIR = os.path.dirname(os.path.realpath(__file__))
@@ -39,6 +40,26 @@ def getVersion():
   """
   with open(os.path.join(REPO_DIR, "VERSION"), "r") as versionFile:
     return versionFile.read().strip()
+
+
+
+def nupicBindingsPrereleaseInstalled():
+  """
+  Make an attempt to determine if a pre-release version of nupic.bindings is
+  installed already.
+
+  @return: boolean
+  """
+  try:
+    nupicDistribution = pkg_resources.get_distribution("nupic.bindings")
+    if pkg_resources.parse_version(nupicDistribution.version).is_prerelease:
+      # A pre-release dev version of nupic.bindings is installed.
+      return True
+  except pkg_resources.DistributionNotFound:
+    pass  # Silently ignore.  The absence of nupic.bindings will be handled by
+    # setuptools by default
+
+  return False
 
 
 
@@ -88,6 +109,14 @@ def findRequirements():
   """
   requirementsPath = os.path.join(REPO_DIR, "requirements.txt")
   requirements = parse_file(requirementsPath)
+
+  if nupicBindingsPrereleaseInstalled():
+    # User has a pre-release version of nupic.bindings installed, which is only
+    # possible if the user installed and built nupic.bindings from source and
+    # it is up to the user to decide when to update nupic.bindings.  We'll
+    # quietly remove the entry in requirements.txt so as to not conflate the
+    # two.
+    requirements = [req for req in requirements if "nupic.bindings" not in req]
 
   return requirements
 


vitaly-krugl:
At first peek, it looks reasonable. I will have another look in the AM.

setup.py
```diff
@@ -89,6 +110,14 @@ def findRequirements():
   requirementsPath = os.path.join(REPO_DIR, "requirements.txt")
   requirements = parse_file(requirementsPath)
 
+  if nupicBindingsPrereleaseInstalled():
+    # User has a pre-release version of nupic.bindings installed, which is only
+    # possible if the user installed and built nupic.bindings from source and
+    # it is up to the user to decide when to update nupic.bindings.  We'll
+    # quietly remove the entry in requirements.txt so as to not conflate the
```
vitaly-krugl: This comment sounds a bit like the code is going to modify the `requirements.txt` file. "quietly remove the entry **from requirements** so as..." might be less confusing.

vitaly-krugl:
It should lead to a better developer workflow than exists presently. A merge of this PR needs to be accompanied by an update of the developer workflow instructions under https://github.com/numenta/nupic/wiki/Installing-and-Building-NuPIC#compiling.

vitaly-krugl:
I am concerned that it might do the wrong thing in this scenario: they had a pre-release of an older version of nupic.bindings on their system (e.g., 0.4.6.dev0) that perhaps forgot about and the version of nupic they are installing from either PyPi or sources has a dependency on a newer version of nupic.bindings (e.g., 0.4.8), then the wrong thing might happen and silently at that.

setup.py
```diff
@@ -89,6 +110,14 @@ def findRequirements():
   requirementsPath = os.path.join(REPO_DIR, "requirements.txt")
   requirements = parse_file(requirementsPath)
 
+  if nupicBindingsPrereleaseInstalled():
+    # User has a pre-release version of nupic.bindings installed, which is only
```
vitaly-krugl: If I am not mistaken, the solution assumes that we're not going to deploy pre-release versions of nupic.bindings to PyPi. Is this a realistic assumption? My understanding is that we're planning to make more frequent deployments to PyPi. Would they not include pre-release versions?

cc @scottpurdy

vitaly-krugl:
I am done with my review. I have a couple of core concerns:
1. The usage scenario where they had an old prerelease version of nupic.bindings on their system and are now installing a new nupic with dependency on  a newer version of nupic.bindings
 * https://github.com/numenta/nupic/pull/3335#issuecomment-245810260
2. The assumption that we would not deploy pre-release versions of nupic.bindings to PyPi.

setup.py
rhyolight: We will not be releasing any `nupic.bindings` versions with the `.dev0` suffix. Only public releases will go on Pypi.

rhyolight:
@vitaly-krugl said:
> I am concerned that it might do the wrong thing in this realistic scenario: they had a pre-release of an older version of nupic.bindings on their system (e.g., 0.4.6.dev0) that the developer perhaps even forgot about and the version of nupic they are installing from either PyPi or sources has a dependency on a newer version of nupic.bindings (e.g., 0.4.8), then the wrong thing might happen and silently at that.

If there were an option like the pip `--ignore-installed` option, that would be perfect.

scottpurdy:
I think this is a reasonable trade off in line with what we discussed at the meeting so I'd like to merge. We can look into whether we want to allow some setup.py flags or something afterwards. Or we could extend this solution to only accept the pre-release version if it is greater than `nupic`s requirement.

So let's try this out while continuing the discussions.

oxtopus:
Re:

> I am concerned that it might do the wrong thing in this realistic scenario: they had a pre-release of an older version of nupic.bindings on their system (e.g., 0.4.6.dev0) that the developer perhaps even forgot about and the version of nupic they are installing from either PyPi or sources has a dependency on a newer version of nupic.bindings (e.g., 0.4.8), then the wrong thing might happen and silently at that.

That is not a coincidence, and is the ideal situation, imo.  Speaking for myself, if I go through the trouble of downloading, building, and installing nupic.core/nupic.bindings, I don't want some external process to interfere.   The risk of " the wrong thing might happen" is mitigated by tests.  If I find out that there's something wrong with my installation of nupic.bindings, I'd  prefer to address it in a way that makes sense for my specific setup vs relying on nupic installation to do the right thing (it can't, anyway).

As it is, there's a defect in the nupic installation that does the wrong thing -- and it happens every time!  I've attempted to address that here by deliberately short-circuiting that process, preventing the installation of nupic from doing that wrong thing.

vitaly-krugl:
I would have preferred at least a visible message on stderr when the dependency is being suppressed. For example, pip will print a message in case an installed package already satisfies the requirement.

cc @scottpurdy @oxtopus

rhyolight:
> I would have preferred at least a visible message on stderr when the dependency is being suppressed. For example, pip will print a message in case an installed package already satisfies the requirement.

We should [create new issues](https://github.com/numenta/nupic/issues/new) for any improvements we'd like to see. I definitely think this is better than the state it was in before.

PR #3339, opened 2016-09-13
Fix Anomaly Likelihood calc of tail probabilities

### LINKED ISSUE #3336: Bug in recursive AnomalyLikelihood calculation of left tail probabilities
The [`normalProbability` method](https://github.com/numenta/nupic/blob/master/src/nupic/algorithms/anomaly_likelihood.py#L732) covers both tails of the Gaussian distribution by flipping values less than the mean. However, the current logic does this flip twice:

```
xp = 2 * distributionParams["mean"] - x
return 1.0 - normalProbability(xp, distributionParams)
```

should be

```
xp = 2 * distributionParams["mean"] - x
return normalProbability(xp, distributionParams)
```

I've implemented this and tested it on NAB, but oddly don't see different overall NAB scores.

BoltzmannBrain:
Fixes #3336

### DIFF
--- a/src/nupic/algorithms/anomaly_likelihood.py
+++ b/src/nupic/algorithms/anomaly_likelihood.py
@@ -743,7 +743,7 @@ def normalProbability(x, distributionParams):
   if x < distributionParams["mean"]:
     # Gaussian is symmetrical around mean, so flip to get the tail probability
     xp = 2 * distributionParams["mean"] - x
-    return 1.0 - normalProbability(xp, distributionParams)
+    return normalProbability(xp, distributionParams)
 
   # Calculate the Q function with the complementary error function, explained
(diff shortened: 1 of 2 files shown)


BoltzmannBrain:
@scottpurdy please review

src/nupic/algorithms/anomaly_likelihood.py
```diff
@@ -743,7 +743,7 @@ def normalProbability(x, distributionParams):
   if x < distributionParams["mean"]:
     # Gaussian is symmetrical around mean, so flip to get the tail probability
     xp = 2 * distributionParams["mean"] - x
-    return 1.0 - normalProbability(xp, distributionParams)
```
oxtopus: dfgadfgdfg

scottpurdy:
replaced with #3448

PR #3344, opened 2016-09-15
SP: Walk neighbors in the same order as in C++

### LINKED ISSUE #3343: SP: the Python and C++ neighbors aren't returned in the same order
As part of https://github.com/numenta/nupic.core/issues/1080, I changed the order that the neighbors are processed. The nupic code needs an update to process them in the same order.

Note that this was already an issue. We worked around it by sorting. https://github.com/numenta/nupic.core/issues/128

So really this issue is just the Nupic side of https://github.com/numenta/nupic.core/issues/128.

mrcslws:
Fixes #3343

This is the matching Python change for https://github.com/numenta/nupic.core/pull/1081. It will fail the compatibility tests until we have a released nupic.bindings with that change.

Originally I used the same approach as in C++, walking the neighbors without creating lists of them. But this approach is slow in Python, sadly. It's better for the Python to batch process big lists in C (via itertools and numpy) than for Python to try to do the processing in Python.

Here's what this change accomplishes:

- Builds a common interface for handling neighbors (matching the C++) `topology.py`
- Walks the neighbors in the same order as in the C++. The SP compatibility tests pass.
- Removes the need to sort neighbors. https://github.com/numenta/nupic.core/issues/128. So for the first time, the neighbors code is compatible.
- Adds to the `_mapColumn` unit test to cover inputs that were previously broken in C++ https://github.com/numenta/nupic.core/issues/1079
- Updates `testUpdateMinDutyCycleLocal` to do the same thing as the C++, because its mocking strategy no longer works with these changes.
- It's slightly faster. About ~32%, according to the test below.

### Note it's only 10 epochs. The [C++ results](https://github.com/numenta/nupic.core/pull/1081) did 100 epochs.

Before

~~~
> time python nupic.research/projects/sp_paper/train_sp_topology_simple.py --spatialImp py --numEpochs 10

real	0m32.865s
user	0m32.688s
sys	0m0.152s
~~~

After

~~~
> time python nupic.research/projects/sp_paper/train_sp_topology_simple.py --spatialImp py --numEpochs 10

real	0m24.733s
user	0m24.565s
sys	0m0.150s
~~~

### DIFF
--- a/src/nupic/math/topology.py
+++ b/src/nupic/math/topology.py
@@ -115,9 +115,8 @@ def neighborhood(centerIndex, radius, dimensions):
     right = min(dimension - 1, centerPosition[i] + radius)
     intervals.append(xrange(left, right + 1))
 
-  return numpy.ravel_multi_index(
-    numpy.array(list(itertools.product(*intervals))).T,
-    dimensions)
+  coords = numpy.array(list(itertools.product(*intervals)))
+  return numpy.ravel_multi_index(coords.T, dimensions)
 
 
 def wrappingNeighborhood(centerIndex, radius, dimensions):
@@ -149,6 +148,5 @@ def wrappingNeighborhood(centerIndex, radius, dimensions):
     interval = [v % dimension for v in xrange(left, right + 1)]
     intervals.append(interval)
 
-  return numpy.ravel_multi_index(
-    numpy.array(list(itertools.product(*intervals))).T,
-    dimensions)
+  coords = numpy.array(list(itertools.product(*intervals)))
+  return numpy.ravel_multi_index(coords.T, dimensions)
--- a/src/nupic/math/topology.py
+++ b/src/nupic/math/topology.py
@@ -1,0 +1,154 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2016, Numenta, Inc.  Unless you have an agreement
+# with Numenta, Inc., for a separate license for this software code, the
+# following terms and conditions apply:
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero Public License for more details.
+#
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+
+import itertools
+import numpy
+
+
+def coordinatesFromIndex(index, dimensions):
+  """
+  Translate an index into coordinates, using the given coordinate system.
+
+  Similar to numpy.unravel_index.
+
+  @param index (int)
+  The index of the point. The coordinates are expressed as a single index by
+  using the dimensions as a mixed radix definition. For example, in dimensions
+  42x10, the point [1, 4] is index 1*420 + 4*10 = 460.
+
+  @param dimensions (list of ints)
+  The coordinate system.
+
+  @returns
+  A list of coordinates of length len(dimensions).
+  """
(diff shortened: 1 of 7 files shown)


src/nupic/research/spatial_pooler.py
```diff
@@ -1259,25 +1260,28 @@ def _mapPotential(self, index, wrapAround=False):
...
+                                           self._inputDimensions)
+
+    columnInputs = columnInputs.astype(uintType)
 
     # Select a subset of the receptive field to serve as the
     # the potential pool
-    numPotential = int(round(indices.size * self._potentialPct))
-    selectedIndices = numpy.empty(numPotential, dtype=uintType)
-    self._random.sample(indices, selectedIndices)
+    numPotential = int(round(columnInputs.size * self._potentialPct))
```
scottpurdy: `round` returns a float that approximates an int. Depending on the representation, the `int` cast may do what you expect but may also round down to the integer below if the float repr is slightly less than the int, right? I.e. `round(0.99)` will approximate `1.0` but the float representation might be slightly less than `1.0`, resulting in `int(round(0.99))` becoming `0`

So perhaps a better method if you want to round up if >=0.5 and down if <0.5 is to do:

```
numPotential = int(columnInputs.size * self._potentialPct + 0.5)
```

src/nupic/research/spatial_pooler.py
```diff
@@ -1477,173 +1481,22 @@ def _inhibitColumnsLocal(self, overlaps, density):
...
-                                     columnCoords[i]+radius+1))
-        curRange = curRange[
-          numpy.logical_and(curRange >= 0, curRange < dimensions[i])]
-
-      rangeND.append(numpy.unique(curRange))
-
-    neighbors = numpy.ravel_multi_index(
-      numpy.array(list(itertools.product(*rangeND))).T, 
-      dimensions).tolist()
+        numActive = int(0.5 + density * len(neighborhood))
```
scottpurdy: Oh here you do the method I suggested above rather than using `round`

src/nupic/math/topology.py
```diff
@@ -0,0 +1,154 @@
...
+
+  intervals = []
+  for i, dimension in enumerate(dimensions):
+    left = max(0, centerPosition[i] - radius)
+    right = min(dimension - 1, centerPosition[i] + radius)
+    intervals.append(xrange(left, right + 1))
+
+  return numpy.ravel_multi_index(
+    numpy.array(list(itertools.product(*intervals))).T,
+    dimensions)
```
scottpurdy: nit: indent more

src/nupic/research/spatial_pooler.py
```diff
@@ -19,13 +19,12 @@
...
 # ----------------------------------------------------------------------
 
-import itertools
-
 import numpy
 from nupic.bindings.math import (SM32 as SparseMatrix,
                                  SM_01_32_32 as SparseBinaryMatrix,
                                  GetNTAReal,
                                  Random as NupicRandom)
+import nupic.math.topology as topology
```
mrcslws: I think it's useful to show that `topology` is a module, not a class.

src/nupic/research/spatial_pooler.py
mrcslws: Fixed. Thanks, I didn't know about that. I do that elsewhere because the existing code already did it.

src/nupic/math/topology.py
```diff
@@ -0,0 +1,154 @@
...
+  """
+  index = 0
+  for i, dimension in enumerate(dimensions):
+    index *= dimension
+    index += coordinates[i]
+
+  return index
+
+
+def neighborhood(centerIndex, radius, dimensions):
```
mrcslws: I went ahead and added this extensibility to the SP by moving all calls to `neighborhood` to `_getInputNeighborhood` and `_getColumnNeighborhood` on the SP. This is much easier here in Python because there's no static typing, and of course because in Python we're returning simple lists, not a magical iterator class.

mrcslws:
My latest change improved perf considerably by relying more on numpy. Now ~32% faster than today's code. I updated my first comment. (The main perf gains are still in the C++ change)

src/nupic/research/spatial_pooler.py
scottpurdy: I don't know how this accomplishes that but, regardless, we don't use this style of import so please change.

PR #3348, opened 2016-09-16
TM: Support extensibility by using instance methods

### LINKED ISSUE #3309: TM: Change three main functions into traditional methods
In the TemporalMemory's [top-level column loop](https://github.com/numenta/nupic/blob/18ef587aba3654440dd8db20e87e2d96894e8b4c/src/nupic/research/temporal_memory.py#L167), we dispatch columns to one of three functions:

- activatePredictedColumn
- burstColumn
- punishPredictedColumn

These three functions exist for readability. We could paste their contents directly into the top-level loop, and that's originally how my code was written, but it made sense to "name" these procedures and move them out of the main loop.

These are actually static methods, which is essentially the same thing as a function.

There are two things here that I don't like:

- They take enormous argument lists. Almost every `self.*` attribute gets passed in. This adds potential for bugs, since you have to carefully get the parameters in the right order. And it makes the code harder to read. It's easier to understand code that you can fit on your screen all at once, and these long argument lists make it where you can't see the whole top-level column loop at once.
- It's a missed opportunity for extensibility. Even if we switch to overridable `classmethod`s, if the methods don't receive the `self` that severely limits the ability to add functionality. For example, the [FaultyTemporalMemory](https://github.com/numenta/nupic.research/blob/1d144224ad7ae5565c202e20952010e3eb9a5168/htmresearch/algorithms/faulty_temporal_memory.py) needs to access `self.deadCells`.

I'd like to convert these to traditional methods which receive a `self`, and stop passing in `self.*` parameters.

mrcslws:
Fixes #3309

It still uses long argument lists, so it doesn't address everything I brought up in #3309, but it still enables extensibility.

As mentioned in that issue, our policy is that code should be organized as:

- Methods that contain almost no logic. They modify the self.
- Functions that don't receive a self. They receive all of their data explicitly. They can mutate data structures that are explicitly passed in, and/or they can return values.

This change makes `activatePredictedColumn`, `burstColumn`, `punishPredictedColumn` into instance methods that contain no logic. They immediately call functions with long argument lists.

### DIFF
--- a/src/nupic/research/temporal_memory.py
+++ b/src/nupic/research/temporal_memory.py
@@ -219,11 +219,13 @@ class TemporalMemory(object):
                                self.matchingSegments, segToCol):
       (column,
        activeColumns,
-       activeSegmentsOnCol,
-       matchingSegmentsOnCol) = columnData
+       columnActiveSegments,
+       columnMatchingSegments) = columnData
       if activeColumns is not None:
-        if activeSegmentsOnCol is not None:
-          cellsToAdd = self.activatePredictedColumn(activeSegmentsOnCol,
+        if columnActiveSegments is not None:
+          cellsToAdd = self.activatePredictedColumn(column,
+                                                    columnActiveSegments,
+                                                    columnMatchingSegments,
                                                     prevActiveCells,
                                                     prevWinnerCells,
                                                     learn)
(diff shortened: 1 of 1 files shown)


mrcslws:
Ready for review. (@oxtopus, maybe?)

oxtopus:
I'm in favor of such change, but defer to @scottpurdy because history.

scottpurdy:
👍

oxtopus:
I didn't quite catch this before, but `_burstColumn()`, `_adaptSegment()`, etc. were moved outside of the class itself.  I think it would have been better to make them static methods on the class, and for that matter, prefix with an additional `_` to take advantage of name mangling to better enforce scope of visibility. e.g.:

```
class TemporalMemory(object):
  ...
  @staticmethod
  def __burstColumn(...):
    ...
  
  def burstColumn(self):
    return self.__burstColumn()
```

mrcslws:
Hmm. What would be the benefit? This code is already in a `temporal_memory` module. Modules seem like a good way to group a bunch of related functions.

PR #3350, opened 2016-09-19
Revert "TM: Support extensibility by using instance methods"

oxtopus:
Reverts numenta/nupic#3348

In retrospect, this wasn't the right change and breaks extensibility.  Consider the following chain of events:

1. `tm.compute()` is called
1. `tm.compute()` calls `self.activateCells()` (`self` is an instance of the `TemporalMemory` class)
1. `self.activateCells()` calls `self.burstColumn()`
1. `self.burstColumn()` calls the module-level `_burstColumn()` function.
1. `_burstColumn()` calls `_adaptSegment()`, `_growSynapses()`, `_leastUsedCell()`

Neither `_adaptSegment()`, `_growSynapses()`, nor `_leastUsedCell()` exist within `TemporalMemory` class, so when a base class implements `burstColumn()`, the implementation must also account for the full chain of events described above.  While technically it is possible to import those functions directly from the `temporal_memory` module, it doesn't feel right given the intent to keep them private.

### DIFF
--- a/src/nupic/research/temporal_memory.py
+++ b/src/nupic/research/temporal_memory.py
@@ -21,6 +21,14 @@
 
 """
 Temporal Memory implementation in Python.
+
+The static methods in this file use the following parameter ordering convention:
+
+1. Output / mutated params
+2. Traditional parameters to the function, i.e. the ones that would still exist
+   if this function were a method on a class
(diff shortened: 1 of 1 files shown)


oxtopus:
cc: @mrcslws @scottpurdy

PR #3351, opened 2016-09-19
TM: Use class methods, not functions, so that subclasses have access

### LINKED ISSUE #3309: TM: Change three main functions into traditional methods

mrcslws:
An update to #3348

Fixes #3309

### DIFF
--- a/src/nupic/research/temporal_memory.py
+++ b/src/nupic/research/temporal_memory.py
@@ -342,15 +342,12 @@ class TemporalMemory(object):
     A list of predicted cells that will be added to active cells and winner
     cells.
     """
-    return _activatePredictedColumn(self.connections, self._random,
-                                    columnActiveSegments,
-                                    prevActiveCells, prevWinnerCells,
-                                    self.numActivePotentialSynapsesForSegment,
-                                    self.maxNewSynapseCount,
-                                    self.initialPermanence,
-                                    self.permanenceIncrement,
-                                    self.permanenceDecrement,
-                                    learn)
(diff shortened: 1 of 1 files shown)


mrcslws:
@oxtopus ready for review.

PR #3352, opened 2016-09-20
Added missing getter methods to TM initiasiling parameters

### LINKED ISSUE #3342: TM initialization parameters should be accessible via getter methods
Just like the SP, the TM should expose the initialization parameters sent to its constructor through getter methods. This allows users to programmatically inspect the TM object to see how it was created. This is a list of initialization parameters that are not exposed.

### `maxSegmentsPerCell`
This param is used to create an instance of the `Connections` class, which exposes `maxSegmentsPerCell` as a public property. So it is possible to get this value from the TM with `tm.connections.maxSegmentsPerCell`, but a getter would make the TM API more direct.

> **Proposal**: add `tm.getMaxSegmentsPerCell()` function.

### `maxSynapsesPerSegment`
This param is used to create an instance of the `Connections` class, which exposes `maxSynapsesPerSegment` as a public property. So it is possible to get this value from the TM with `tm.connections.maxSynapsesPerSegment`, but a getter would make the TM API more direct.

> **Proposal**: add `tm.getMaxSynapsesPerSegment()` function.

arhik:
Fixes #3342

### DIFF
--- a/src/nupic/research/temporal_memory.py
+++ b/src/nupic/research/temporal_memory.py
@@ -295,7 +295,7 @@ class TemporalMemory(object):
       if numActivePotential[i] >= self.minThreshold
     )
 
-    maxSegmentsPerCell = self.connections.maxSegmentsPerCell
+    maxSegmentsPerCell = self.getMaxSegmentsPerCell()
     segmentKey = lambda segment: (segment.cell * maxSegmentsPerCell
                                   + segment.idx)
     self.activeSegments = sorted(activeSegments, key = segmentKey)
--- a/src/nupic/research/temporal_memory.py
+++ b/src/nupic/research/temporal_memory.py
@@ -849,6 +849,22 @@ class TemporalMemory(object):
     self.predictedSegmentDecrement = predictedSegmentDecrement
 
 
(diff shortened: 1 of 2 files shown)


rhyolight:
ping

rhyolight:
We had a hiccup with our tooling server overnight, so the AppVeyor build did not run against this PR. It should be fine without it.

arhik:
@rhyolight No it doesn't sound nit-picky. It makes sense and I was equally concerned.

I am also thinking if `self.getMaxSegmentsPerCell()` function call is more appropriate now  where `self.connections.maxSegmentsPerCell` is used [ There is only one occasion, but asking out of curiosity] Or do you consider it as an function overhead for each call.  I am not sure if python does inline optimization for one line functions.

[location of such instance](https://github.com/arhik/nupic/blob/90045ef3d54feb1dc333aaa40235ca37945be4a2/src/nupic/research/temporal_memory.py#L298)

rhyolight:
> I am also thinking if `self.getMaxSegmentsPerCell()` function call is more appropriate now where `self.connections.maxSegmentsPerCell` is used [There is only one occasion, but asking out of curiosity]

I agree.

PR #3354, opened 2016-09-20
Change private _burstColumn class method signature to accept a cellsF…

### LINKED ISSUE #3353: Change TemporalMemory _burstColumn signature
`TemporalMemory._burstColumn()` calculates a range of cells on which to operate.  The calculation depends on the `cellsPerColumn` argument, which is derived from the `cellsPerColumn` instance attr, passed to it in the public `TemportalMemory.burstColumn()` instance method.  Since this calculation is a derivation of something that depends on the instance, it might as well occur within the instance method and passed as an argument to the private class method in lieu of the `cellsPerColumn` number.

oxtopus:
…orColumn argument in lieu of a cellsPerColumn argument.  Move the calculation that otherwise depends on cellsPerColumn into the instance method.

Fixes #3353

### DIFF
--- a/src/nupic/research/temporal_memory.py
+++ b/src/nupic/research/temporal_memory.py
@@ -382,10 +382,10 @@ class TemporalMemory(object):
 
     return self._burstColumn(
       self.connections, self._random, column, columnMatchingSegments,
-      prevActiveCells, prevWinnerCells,
+      prevActiveCells, prevWinnerCells, cellsForColumn,
       self.numActivePotentialSynapsesForSegment, self.maxNewSynapseCount,
       self.initialPermanence, self.permanenceIncrement,
-      self.permanenceDecrement, learn, cellsForColumn)
+      self.permanenceDecrement, learn)
 
 
   def punishPredictedColumn(self, column, columnActiveSegments,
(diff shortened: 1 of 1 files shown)


src/nupic/research/temporal_memory.py
```diff
@@ -376,12 +376,16 @@ def burstColumn(self, column, columnMatchingSegments, prevActiveCells,
...
-      self.connections, self._random,
-      column, columnMatchingSegments, prevActiveCells, prevWinnerCells,
-      self.numActivePotentialSynapsesForSegment,
-      self.cellsPerColumn, self.maxNewSynapseCount, self.initialPermanence,
-      self.permanenceIncrement, self.permanenceDecrement, learn)
+      self.connections, self._random, column, columnMatchingSegments,
+      prevActiveCells, prevWinnerCells,
+      self.numActivePotentialSynapsesForSegment, self.maxNewSynapseCount,
+      self.initialPermanence, self.permanenceIncrement,
+      self.permanenceDecrement, learn, cellsForColumn)
```
mrcslws: One quick request: move the parameter up to be after `prevWinnerCells`. This would better follow the scheme:

~~~python
  # These class methods use the following parameter ordering convention:
  #
  # 1. Output / mutated params
  # 2. Traditional parameters to the method, i.e. the ones that would still
  #    exist if this were in instance method.
  # 3. Model state (not mutated)
  # 4. Model parameters (including "learn")
~~~

Oops, typo. "this were in instance method." should be "an", not "in".

Issue #3358, opened 2016-09-21
SP columns near the boundary has less neighbors


ywcui1990:
@mrcslws 
When topology is turned on for spatial pooler, columns near the boundary has less number of columns in their "neighborhood", yet they share the same inhibitionRadius as other columns. 

The issue is whether we should enforce each SP column to have the same number of neighbors, a simple fix could be to replace `topology.neighborhood` with `topology.wrappingNeighborhood` on https://github.com/numenta/nupic/blob/master/src/nupic/research/spatial_pooler.py#L1508

Practically, SP columns preferentially form more connections to inputs near the boundary. To illustrate the problem, I plotted the number of SP columns that are connected to each input bits after training. The training dataset is uniformly distributed, but inputs near the corner/boundary are preferentially connected by SP columns.

![trauma_boosting_with_topology_epoch_149](https://cloud.githubusercontent.com/assets/5067931/18728702/4700ce06-8003-11e6-9026-5e11b6edacca.png)

Issue #2639, opened 2015-10-07
SP: review a commented out inhibition test


breznak:
at https://github.com/numenta/nupic/blob/master/tests/unit/nupic/research/spatial_pooler_unit_test.py#L1318

PR #3361, opened 2016-09-22
Fix logic in getColumnNeighborhood

ywcui1990:
Fixes https://github.com/numenta/nupic/issues/3358
Also fixes https://github.com/numenta/nupic/issues/2639

The change has an impact on the testInhibitColumnsLocal. I have verified that the new result makes sense given periodic neighborhood conditions.

@mrcslws Could you make the same change in nupic.core? I am less familiar with the C++ code there. The compatibility test will break until the logic is updated in nupic.core.

### DIFF
--- a/src/nupic/research/spatial_pooler.py
+++ b/src/nupic/research/spatial_pooler.py
@@ -1157,6 +1157,7 @@ class SpatialPooler(object):
     p = int(p*100000) / 100000.0
     return p
 
+
   def _initPermanence(self, potential, connectedPct):
     """
     Initializes the permanences of a column. The method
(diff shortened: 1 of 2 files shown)


src/nupic/research/spatial_pooler.py
```diff
@@ -1505,9 +1506,15 @@ def _getColumnNeighborhood(self, centerColumn):
     @returns (1D numpy array of integers)
     The columns in the neighborhood.
     """
```
mrcslws: The docstring should be updated to match the other one. `"Simply calls topology.wrappingNeighborhood or topology.neighborhood."`

tests/unit/nupic/research/spatial_pooler_unit_test.py
```diff
@@ -1420,7 +1420,7 @@ def testInhibitColumnsLocal(self):
     sp._inhibitionRadius = 3
     overlaps = numpy.array([1, 1, 1, 1, 1, 1, 1, 1, 1, 1], dtype=realDType)
                         #   W  W  L  L  W  W  L  L  L  W
-    trueActive = [0, 1, 4, 5, 8]
+    trueActive = [0, 1, 4, 5]
```
mrcslws: We should probably test both cases `wrapAround=True` and `wrapAround=False`.

ywcui1990:
Thanks Marcus, I have addressed your comments and updated the SP unit test

PR #3365, opened 2016-09-25
TM missing initialization getters and corresponding test

### LINKED ISSUE #3342: TM initialization parameters should be accessible via getter methods

arhik:
Fixes #3342

### DIFF
--- a/tests/unit/nupic/research/temporal_memory_test.py
+++ b/tests/unit/nupic/research/temporal_memory_test.py
@@ -1005,19 +1005,16 @@ class TemporalMemoryTest(unittest.TestCase):
       cellsPerColumn=32,
       maxSegmentsPerCell=200
     )
-    self.assertEqual(tm.connections.maxSegmentsPerCell, tm.getMaxSegmentsPerCell())
-    self.assertEqual(tm.connections.maxSegmentsPerCell, 200)
     self.assertEqual(tm.getMaxSegmentsPerCell(), 200)
 
+  
   def testMaxSynapsesPerSegmentGetter(self):
-      tm = TemporalMemory(
-        columnDimensions=[32,32],
-        cellsPerColumn=16,
-        maxSynapsesPerSegment=150
-      )
-      self.assertEqual(tm.connections.maxSynapsesPerSegment, 150)
-      self.assertEqual(tm.getMaxSynapsesPerSegment(), 150)
-      self.assertEqual(tm.connections.maxSynapsesPerSegment, tm.getMaxSynapsesPerSegment())
+    tm = TemporalMemory(
+      columnDimensions=[32,32],
+      cellsPerColumn=16,
+      maxSynapsesPerSegment=150
+    )
+    self.assertEqual(tm.getMaxSynapsesPerSegment(), 150)
 
 
   def serializationTestPrepare(self, tm):
--- a/tests/unit/nupic/research/temporal_memory_test.py
+++ b/tests/unit/nupic/research/temporal_memory_test.py
@@ -999,6 +999,27 @@ class TemporalMemoryTest(unittest.TestCase):
     self.assertEqual(columnsForCells[99], set([399]))
 
 
+  def testMaxSegmentsPerCellGetter(self):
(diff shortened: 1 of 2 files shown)


tests/unit/nupic/research/temporal_memory_test.py
```diff
@@ -999,6 +999,27 @@ def testMapCellsToColumns(self):
...
+  def testMaxSegmentsPerCellGetter(self):
+    tm = TemporalMemory(
+      columnDimensions=[64,64],
+      cellsPerColumn=32,
+      maxSegmentsPerCell=200
+    )
+    self.assertEqual(tm.connections.maxSegmentsPerCell, tm.getMaxSegmentsPerCell())
+    self.assertEqual(tm.connections.maxSegmentsPerCell, 200)
+    self.assertEqual(tm.getMaxSegmentsPerCell(), 200)
+
```
rhyolight: Nitpick: the rest of this file has 2 newlines between function definitions.

tests/unit/nupic/research/temporal_memory_test.py
```diff
@@ -999,6 +999,27 @@ def testMapCellsToColumns(self):
...
+  def testMaxSynapsesPerSegmentGetter(self):
+      tm = TemporalMemory(
+        columnDimensions=[32,32],
+        cellsPerColumn=16,
+        maxSynapsesPerSegment=150
+      )
+      self.assertEqual(tm.connections.maxSynapsesPerSegment, 150)
```
rhyolight: I would remove this line. It tests the internal state of the object, which should be allowed to change. That is one of the purposes of the getter function itself.

tests/unit/nupic/research/temporal_memory_test.py
```diff
@@ -999,6 +999,27 @@ def testMapCellsToColumns(self):
...
+      self.assertEqual(tm.getMaxSynapsesPerSegment(), 150)
+      self.assertEqual(tm.connections.maxSynapsesPerSegment, tm.getMaxSynapsesPerSegment())
```
rhyolight: Also remove this line, because it is testing something that you've already tested. I think you only need the one assertion: `self.assertEqual(tm.getMaxSynapsesPerSegment(), 150)`

tests/unit/nupic/research/temporal_memory_test.py
```diff
@@ -999,6 +999,27 @@ def testMapCellsToColumns(self):
...
 
+  def testMaxSegmentsPerCellGetter(self):
+    tm = TemporalMemory(
+      columnDimensions=[64,64],
+      cellsPerColumn=32,
+      maxSegmentsPerCell=200
+    )
+    self.assertEqual(tm.connections.maxSegmentsPerCell, tm.getMaxSegmentsPerCell())
+    self.assertEqual(tm.connections.maxSegmentsPerCell, 200)
+    self.assertEqual(tm.getMaxSegmentsPerCell(), 200)
```
rhyolight: Same comment as the others below. You only need this one assertion: 

    self.assertEqual(tm.getMaxSegmentsPerCell(), 200)

rhyolight:
👍 Nicely done, thanks @arhik!

PR #3388, opened 2016-10-14
Cleanup references to nupic.bindings and old CI code for manually fet…

### LINKED ISSUE #3387: Manual download of nupic.bindings no longer needed
We previously had logic for building against unreleased nupic.bindings versions (via `.nupic_modules`) and after that still required manually fetching nupic.bindings from S3 for Linux. These cases are no longer necessary so we should remove code that manually fetches nupic.bindings or references to the old S3 location for intermediary builds.

scottpurdy:
…ching nupic.bindings since it should be found on PyPI without doing anything special.

@rhyolight please review

fixes #3387


rhyolight:
@scottpurdy I think you just need to install `pytest` and make sure the executable is in the path.

scottpurdy:
@rhyolight thanks, it must have been installing `py.test` as part of the explicit nupic.bindings installation to the default Python package install location. When I removed that, it was still installing `py.test` as part of the nupic installation but that had the `--user` flag and the `PATH` wasn't set up for that. I'm going to try to remove `--user` and get rid of the seemingly useless `PATH` and `PYTHONPATH` manipulation.

scottpurdy:
Assuming the latest passes, this should be ready to go. Cleaned up a bunch of things that were pretty confusing in the setup.

PR #3391, opened 2016-10-26
Makes `anomaly_likelihood.py` compliant to Python3

### LINKED ISSUE #3392: Make `anomaly_likelihood.py` compliant to Python3
This is a proposal from @vaulttech. He already has a PR (#3391).

Relevant discussion: https://discourse.numenta.org/t/should-we-make-nupic-python-3-compliant/1537

vaulttech:
Fixes #3392

I tried using nupic's "anomaly_likelihood" for a small home program in Python3 and realized it was not working because nupic only works for Python2. These changes should make at least this module compatible with both Pythons

### DIFF
--- a/src/nupic/algorithms/anomaly_likelihood.py
+++ b/src/nupic/algorithms/anomaly_likelihood.py
@@ -118,7 +118,7 @@ class AnomalyLikelihood(object):
 
 
     if claLearningPeriod != None:
-      print "claLearningPeriod is deprecated, use learningPeriod instead."
+      print("claLearningPeriod is deprecated, use learningPeriod instead.")
       self._learningPeriod = claLearningPeriod
     else:
       self._learningPeriod = learningPeriod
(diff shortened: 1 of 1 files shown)


vaulttech:
Do I need to create an issue for this? It seems like a very simple (and merely syntactical) change into the code.

[also, I read the licence. I signed it -- but I am confused on if just typing my name with that Adobe Sign is enough -- I can't draw my signature O__o]

rhyolight:
Hi @vaulttech and thanks for the contribution. Yes, we should create an issue for this according to our [development process](https://github.com/numenta/nupic/wiki/Development-Process), but this brings up the bigger question of do we want to do it at all? 

What is the point of changing just this one file in the source code to be Python3 compliant? Wouldn't we want to make a concerted effort to update the entire codebase if we decided to do this?

CC @numenta/nupic-committers

rhyolight:
@vaulttech Also, yes you signed the license correctly 😄

vaulttech:
> What is the point of changing just this one file in the source code to be Python3 compliant? Wouldn't we want to make a concerted effort to update the entire codebase if we decided to do this?

Well... in the same way that I needed it, it may be that other people may also need only a small part of the whole code, and might find it convenient if it simply works smoothly with both Pythons.

I mean... instead of thinking of "why" accepting the change, it would possibly be better to think "why not?": you don't lose anything (if you look at the changes, you'll see that they are as simple as changing the `print` statements and transforming calls like `dictionary.has_key(blah)` into `blah in dictionary`).

vaulttech:
> Yes, we should create an issue for this according to our development process, but this brings up the bigger question of do we want to do it at all? 

Only to make it clear, I have taken a look at the [development process](https://github.com/numenta/nupic/wiki/Development-Process) and I am unsure, in this specific case (and based on the cited sentence), if you will create an issue or if I should do it (I ask because you said "we"). I can totally do it, of course, if you agree that it makes sense =)

PR #3393, opened 2016-10-28
Break the build.

rhyolight:
yeah do not merge this

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@
 
 """Installation script for Python nupic package."""
 
-import os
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -21,7 +21,7 @@
 
 """Installation script for Python nupic package."""
 
-import os
+import OOOOOOOOOOOos
```
scottpurdy: I don't think this will work.

scottpurdy: 😜

rhyolight: I blame python!!

Issue #3446, opened 2017-01-18
Potential API violation(s) in engine/__init__.py


oxtopus:
Consider the following snippet from the `CollectionWrapper` implementation in `engine/__init__.py`:

```python
  def keys(self):
    keys = set()
    for i in range(self.collection.getCount()):
      keys.add(self.collection.getByIndex(i)[0])
    return keys

  def values(self):
    values = set()

    for i in range(self.collection.getCount()):
      p = self.collection.getByIndex(i)
      values.add(self.valueWrapper(p[0], p[1]))
    return values

  def items(self):
    items = set()
    for i in range(self.collection.getCount()):
      p = self.collection.getByIndex(i)
      items.add((p[0], self.valueWrapper(p[0], p[1])))
    return items
```

Now, consider this snippet from the [Python documentation re: dict objects](https://docs.python.org/2/library/stdtypes.html#dict.items)...

> If items(), keys(), values(), iteritems(), iterkeys(), and itervalues() are called with no intervening modifications to the dictionary, the lists will directly correspond. This allows the creation of (value, key) pairs using zip(): pairs = zip(d.values(), d.keys()). The same relationship holds for the iterkeys() and itervalues() methods: pairs = zip(d.itervalues(), d.iterkeys()) provides the same value for pairs. Another way to create the same list is pairs = [(v, k) for (k, v) in d.iteritems()].

In Python, sets are unordered and because of their use in `CollectionWrapper.keys()` and `CollectionWrapper.values()`, and `CollectionWrapper.items()`, there's no guarantee of consistent ordering between the two, potentially resulting in a violation of the assumptions that `zip(d.keys(), d.values()) == d.items()` as is consistent with `dict` and `dict`-like objects in Python.

`CollectionWrapper.keys()` and `CollectionWrapper.values()` should iterate in the same order, producing lists (not sets).  `CollectionWrapper.items()` could then be simplified to something like:

```python
def items(self):
  return zip(self.keys(), self.values())
```

rhyolight:
Very well-explained Austin, thanks.

PR #3447, opened 2017-01-18
Use IterableCollection from engine_internal

oxtopus:
This removes `CollectionIterator` in favor of the recently-added `nupic.bindings.engine_internal.IterableCollection`.  There is one important difference between the two implementations, however.  `CollectionIterator` (removed here) iterates over the names only whereas `IterableCollection` iterates over the key-value pairs.  As a result, some of the tests needed to be updated to reflect this change in api and as such requires careful consideration.

Looking into this exposed some flaws in `CollectionWrapper` with respect to its `dict`-like interface documented at https://github.com/numenta/nupic/issues/3446.  Fixing those flaws are not in scope of this change.  A future change to `nupic.bindings.engine_internal` may obviate `CollectionWrapper` entirely, shifting focus there.  In the meantime, this change is meant only to remove some duplicated code.

cc @scottpurdy @rhyolight

### DIFF
--- a/src/nupic/engine/__init__.py
+++ b/src/nupic/engine/__init__.py
@@ -1,7 +1,6 @@
-
 # ----------------------------------------------------------------------
 # Numenta Platform for Intelligent Computing (NuPIC)
-# Copyright (C) 2013, Numenta, Inc.  Unless you have an agreement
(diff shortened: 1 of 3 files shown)


oxtopus:
FYI, this is not mergable until the changes it depends on in nupic.core are released and reference in requirements.txt updated.

PR #3448, opened 2017-01-19
Fix gaussian tail probability calculation

### LINKED ISSUE #3336: Bug in recursive AnomalyLikelihood calculation of left tail probabilities

scottpurdy:
fixes #3336 

@subutai please review


src/nupic/algorithms/anomaly_likelihood.py
```diff
@@ -731,9 +731,11 @@ def nullDistribution(verbosity=0):
 
 def normalProbability(x, distributionParams):
```
subutai: Should we rename this to tailProbability?

scottpurdy: Done.

PR #3449, opened 2017-01-20
NUP-2310 Foundation for network visualization framework

oxtopus:
@scottpurdy @subutai This lays the foundation for visualizing networks and requires recent changes to nupic.bindings to allow for the introspection of network links (i.e. this will fail until the next release has completed).  With this in place, we can focus on iterating on the details of the visualization moving forward, but for now it has a few naive renderers.


oxtopus:
@scottpurdy The doesn't block on user input, if that's what your suggesting.  Regardless, I'll move it into an examples dir.  Thanks for the suggestion.

scottpurdy:
@oxtopus - yes, I did think that `viz.render` would pop up a window and wait for user input. What actually happens?

oxtopus:
@scottpurdy I removed the test and added an example.  Please let me know if you have any additional comments to close out this PR, otherwise I will assume it's good to merge.

src/nupic/frameworks/viz/examples/visualize_network.py
```diff
@@ -0,0 +1,63 @@
...
+
+  # Render w/ graphviz
+  viz.render(renderer=GraphVizRenderer)
+
+  # Render w/ networkx
+  viz.render(renderer=NetworkXRenderer)
+
+
+
+if __name__ == "__main__":
```
scottpurdy: Alternatively, you could have the setup.py specify `main` in this file to be installed as an executable (like `nupic-network-visualization-example` or something). Up to you.

src/nupic/frameworks/viz/graphviz_renderer.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+
+class GraphVizRenderer(object):
+  """
+  Network visualization "renderer" implementation to render a network with
+  graphviz.
+  """
+  def render(self, G):
+    G = nx.nx_agraph.to_agraph(G)
```
scottpurdy: `graph` or something would be more explanatory (here and elsewhere). And local variables should start with a lowercase letter

oxtopus: Thanks.  All of the networkx documentation and related literature tend to use the `G` convention.  I will change to `graph` per your request.

PR #3461, opened 2017-02-13
Use better parameters in the TM performance benchmark

### LINKED ISSUE #3460: The TM performance benchmark script uses bad TM params
These aren't good parameters: https://github.com/numenta/nupic/blob/40f0b4c40d168be82854057183d935b9cd0be19c/scripts/temporal_memory_performance_benchmark.py#L226

This is another case of the "maxNewSynapseCount" parameter name causing confusion. It's really more of a "desired number of active synapses per segment" -- maybe "sampleSize" would be a better name. Anyway, it should never be lower than the "activatonThreshold".

mrcslws:
Fixes #3460


rhyolight:
Yo yo what's the hold up bro

PR #3468, opened 2017-02-24
Serializable base class

### LINKED ISSUE #2538: Serializable interface/abstract class
Functionality: common logic for serialization, will hold the includes, allow easily override serialization for the whole project (eg turn it off for some fork of nupic,...)

API:

```
import capnp...

__init__(type='capnp')
def serialize():
  if self.type == 'off':
    return
  return write()
def deserialize(obj):
  if self.type == 'off':
    return
  return read()
getType() # 'oldpy', 'capnp', 'off'
def write():
  raise # implment in your code
def read():
  raise # to implement
```

Classes: all that support, or should, serialization. Example: `MyClass: super(MyClass, self).__init__(type='off')`

oxtopus:
Fixes #2538

This is roughly analogous to https://github.com/numenta/nupic.core/blob/master/src/nupic/types/Serializable.hpp

### DIFF
--- a/src/nupic/serializable.py
+++ b/src/nupic/serializable.py
@@ -1,0 +1,79 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2017, Numenta, Inc.  Unless you have an agreement
+# with Numenta, Inc., for a separate license for this software code, the
+# following terms and conditions apply:
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero Public License for more details.
+#
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+from abc import ABCMeta, abstractmethod
+
+
+
+class Serializable(object):
+  """ Serializable base class establishing `read()` and `write()` abstract
+      methods, `readFromFile()` and `writeToFile()` concrete methods to support
+      serialization with Cap'n Proto.
+  """
+
+
+  __metaclass__ = ABCMeta
+
+
+  @classmethod
+  @abstractmethod
+  def getCapnpSchema(cls):
+    pass  # Not implemented here.  Per abc protocol, attempts to subclass without
+    # overriding will fail.
+
+
+  @classmethod
+  @abstractmethod
+  def read(cls, proto):
+    pass # Not implemented here.  Per abc protocol, attempts to subclass without
+         # overriding will fail.
+
+
+  @abstractmethod
+  def write(self, proto):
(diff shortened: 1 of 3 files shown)


src/nupic/serializable.py
```diff
@@ -0,0 +1,79 @@
...
+         # overriding will fail.
+
+
+  @classmethod
+  def readFromFile(cls, inp, packed=True):
+    # Get capnproto schema from instance
+    schema = cls.getCapnpSchema()
+
+    # Read from file
+    proto = getattr(schema, "read_packed" if packed else "read")(inp)
```
scottpurdy: I like that you're included packed and unpacked options. I think this would be easier to read as regular if/else blocks though (and same in `writeToFile`)

src/nupic/serializable.py
```diff
@@ -0,0 +1,79 @@
...
+  @abstractmethod
+  def getCapnpSchema(cls):
+    pass  # Not implemented here.  Per abc protocol, attempts to subclass without
+    # overriding will fail.
+
+
+  @classmethod
+  @abstractmethod
+  def read(cls, proto):
+    pass # Not implemented here.  Per abc protocol, attempts to subclass without
```
scottpurdy: Might be better to turn the comment into a doc string that includes documentation on what the method is supposed to do so it is easy for implementers to know what to do. Same for `write`

src/nupic/serializable.py
```diff
@@ -0,0 +1,79 @@
...
+      serialization with Cap'n Proto.
+  """
+
+
+  __metaclass__ = ABCMeta
+
+
+  @classmethod
+  @abstractmethod
+  def getCapnpSchema(cls):
```
scottpurdy: If we aren't including "capnp" in the `read`/`write` methods then `getSchema` is probably sufficient. But I'm fine either way

tests/unit/nupic/serializable_test.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+  def testReadFromAndWriteToFile(self):
+    """ Test generic usage of serializable mixin class """
+
+    class Bar(object):
+      pass
+
+
+    class Foo(Bar, Serializable):
```
scottpurdy: Is this because you can't subclass exclusively from `Serializable`? Can you just change `Bar` -> `object` and remove the `Bar` class?

tests/unit/nupic/serializable_test.py
```diff
@@ -0,0 +1,80 @@
...
+        foo = object.__new__(cls)
+        foo.bar = proto.bar
+        return foo
+
+
+      def write(self, proto):
+        proto.bar = self.bar
+
+
+    self.addCleanup(os.remove, "foo.data")
```
scottpurdy: If `writeToFile` fails because of a bug and doesn't write a file then `os.remove` will also fail. I'm not sure what this will look like in the test output but it could be confusing.

It is a rare case and the errors would reflect a real bug so I don't think you should address it explicitly but thought I'd point it out in case you think there is any cause for concern.

tests/unit/nupic/serializable_test.py
oxtopus: You can (see testABCProtocolEnforced).  I added it in to show (and provide imolicit test coverage) that it can be used as a mixin.

I will have to verify, but iirc `class Foo(object, Serializable):` fails.

scottpurdy: Got it, and I see the doc string explaining that.

oxtopus:
@scottpurdy I believe I've addressed your feedback and all tests pass.  Can I get a once-over?

PR #3474, opened 2017-03-06
[NUP-2372] Replace doxygen with Sphinx

### LINKED ISSUE #3473: Replace doxygen doc build with sphinx

rhyolight:
Fixes #3473

This work is ready to be merged, because it provides better docs than we currently have, but it is not complete. I want to commit this so I can get it automated and then iterate on docstring updates and styling.

Some example docstring changes are included in this PR as a proof-of-concept.


rhyolight:
@oxtopus

requirements.txt
```diff
@@ -15,7 +15,13 @@ PyMySQL==0.6.2
...
 prettytable==0.7.2
+
 # When updating nupic.bindings, also update any shared dependencies to keep
 # versions in sync.
 nupic.bindings==0.5.1
 numpy==1.11.2
+
+# for documentation
+sphinx==1.5.3
+sphinx-autobuild==0.6.0
```
rhyolight: @scottpurdy @oxtopus I'm thinking about removing the sphinx requirements. No need for most users to have this installed.

scottpurdy: That's fine. Maybe put it in a separate file like requirements-dev.txt

oxtopus: You can add an entry to `setup.py`, a la "capnp" and "viz" (https://github.com/numenta/nupic/blob/master/setup.py#L164-L165).  That would allow someone the option of installing the sphinx/doc dependencies w/ setup.py or pip.

PR #3484, opened 2017-03-09
Cleaned up TM vs TP example code

### LINKED ISSUE #3483: cleanup tm/tp example code

rhyolight:
Fixes #3483

Remove older TP examples to encourage users to use the TM instead.
Updated tests and readme, and related code.


rhyolight:
@scottpurdy Looking to start cleaning up the codebase. I don't think we should provide example code for an API we are not documenting.

scottpurdy:
The examples in `examples/tp` seem relevant. We should just rename to TM. And at least one looked more like a test than an example so perhaps could be moved to the tests dir.

rhyolight:
@scottpurdy The files I removed are never run anywhere I can find. I left one file at `examples/tm/tp_test.py` that is run in Travis-CI, but seemingly not in bamboo.

rhyolight:
@scottpurdy Ok how about I leave the names of the TP scripts the same and leave them in the directory, but change the directory name to `tm`? Just trying to make this less confusing.

scottpurdy:
I see, you are worried about confusion between the old and new TM implementations? We should work with @subutai to decide how to differentiate between the old and new implementations now that we can change the old version name from TP/TP10X2 to something more accurate.

rhyolight:
I will close for now then.

PR #3520, opened 2017-04-05
NUP-2394: network API code example

### LINKED ISSUE #3522: create Network API Quick Start docs

marionleborgne:
Fixes #3522.

The code example for the network API is done but the PR is blocked on `NUP-2396`. Currently the example has to call `customCompute()` on the `classifierRegion` since classification of continuous values is not supported by the network API. I'm moving on to the blocking issue (`NUP-2396`) to fix this.

Todos, for when I get back to this issue:

- [x] remove call to `customCompute()` and make use of new region links created in `NUP-2396`.
- [x] take care of TODOs in `nupic/docs/examples/params/model.yaml` by adding new params documentation based on feedback collected in google doc.

cc @rhyolight


rhyolight:
Only input so far, is if we're going to go all in on YAML model params for this example, we should also change the algorithm quick start to use the same YAML file to load params.

marionleborgne:
OK, I'll update the algorithm example.

marionleborgne:
@rhyolight This PR is blocked by #3526. 

Changelog:
* Use new links between SensorRegion and SDRClassifierRegion that allow to pass `bucketIdx` and `actValue` so that the classifier can predict continuous values - like in the 2 other quick-start examples.
* Update YAML with more comments based on @scottpurdy 's feedback.

Once #3526 is merged, this is ready to be reviewed.

rhyolight:
@marionleborgne I also noticed you are writing results to an output file instead of to the console, so the RST will need to be updated to reflect these changes in the descriptive text.

marionleborgne:
@rhyolight Actually, I am still writing to the console. I just added writing to an output file, so that I can compare the results of the 3 examples. I'll update the descriptive text.

rhyolight:
Sounds good. I think the examples should be as simple as possible. Writing
to the console is fine.

---------
Matt Taylor
OS Community Manager
Numenta

On Mon, Apr 10, 2017 at 10:32 AM, Marion Le Borgne <notifications@github.com

marionleborgne:
After discussion with @rhyolight, the plan is to take out writing to file once we fix the discrepancy between prediction results. Out of the scope of this PR. Task tracked here: https://jira.numenta.com/browse/NUP-2400

docs/examples/params/model.yaml
```diff
@@ -0,0 +1,317 @@
+# Type of model that the rest of these parameters apply to.
```
scottpurdy: I'm not sure that we want to have the full documentation for all parameters in every model configuration file. Perhaps we could centralize this information somewhere and link to that reference page at the top of this file?

rhyolight: Do you mean other places in the codebase where we have model params with detailed comments? If yes, I agree.

I think this file is a great place for it, because it will show up here:

http://docs.numenta.org/quick-start/example-model-params.html

Perhaps we should remove the comments from other places and refer them to the API docs?

scottpurdy: I see, yeah might be fine here. I think the real source of truth should be the algorithm doc strings but those are spread out throughout the code so this can be the place where we include them all. We may want to have references back to the individual algorithms in case they get out of sync.

marionleborgne:
@rhyolight @scottpurdy: can someone review this PR? nothing is blocking it since #3526 was merged last week.

docs/examples/algo/complete-example.py
```diff
@@ -1,144 +1,171 @@
 import csv
```
marionleborgne: Yes

docs/examples/util/check_predictions.py
```diff
@@ -0,0 +1,84 @@
+import os
```
marionleborgne: A script to inspect the prediction results of the 3 examples:
- docs/algo/complete-example.py
- docs/opf/complete-example.py
- docs/networkapi/complete-example.py

Adding this info at the top of the file.

docs/examples/algo/complete-example.py
```diff
@@ -1,144 +1,171 @@
...
+  encodingWidth = (timeOfDayEncoder.getWidth()
+                   + weekendEncoder.getWidth()
+                   + scalarEncoder.getWidth())
 
   sp = SpatialPooler(
     # How large the input encoding will be.
     inputDimensions=(encodingWidth),
     # How many mini-columns will be in the Spatial Pooler.
-    columnDimensions=(2048),
+    columnDimensions=(spParams['columnCount']),
```
marionleborgne: This is @rhyolight 's example. I think he pulled the params one by one so that he can describe them in the the quick-start guide: http://nupic.docs.numenta.org/quick-start/algorithms.html

docs/examples/util/run-all-complete-examples.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
marionleborgne: Agreed. 

1. I'll add the test you are talking about as part of: https://jira.numenta.com/browse/NUP-2401  (Fix results discrepancy between OPF, network API and algorithm API)
2. I'll take out the scripts under `docs/util`.

marionleborgne:
@scottpurdy thanks for the review, I addressed the feedback. Some things to note:
* I added a unit test (as you suggested) to make sure the 3 examples can run and don't throw any exception. This does not check if the prediction results are correct.
* I will add a unit test for the prediction results consistency between network API, OPF and algo API as part of https://jira.numenta.com/browse/NUP-2401.

docs/examples/algo/complete-example.py
```diff
@@ -1,87 +1,106 @@
 import csv
 import datetime
-
 import numpy
+import os
+import yaml
```
oxtopus: why yaml?  i.e. what benefit does it have over a native python module?

marionleborgne: In all 3 examples we are using a single YAML file with all the params. This YAML is heavily commented to explain each param. Using YAML vs native python module makes it a bit more obvious in the docs that it's a configuration file. Another (small) benefit is that it displays in a more compact manner in the docs and it's easier to read than the native python module.

rhyolight:
@marionleborgne Will review today

docs/examples/algo/complete-example.py
rhyolight: Python model params adds the extra requirement of a local import, an extra directory, and an empty `__init__.py` file.

docs/examples/algo/complete-example.py
```diff
@@ -1,87 +1,106 @@
...
+    synPermActiveInc=spParams["synPermActiveInc"],
+    synPermConnected=spParams["synPermConnected"],
     # boostStrength controls the strength of boosting. Boosting encourages
     # efficient usage of SP columns.
-    boostStrength=3.0,
+    boostStrength=spParams["boostStrength"],
     # Random number generator seed.
-    seed=1956,
+    seed=spParams["seed"],
+    # TODO: is this useful?
```
rhyolight: Probably not. Feel free to remove.

rhyolight: @marionleborgne If you choose not to make the change above, let me know so I can merge as-is.

rhyolight: I went ahead and merged. Will revisit.

PR #3526, opened 2017-04-08
NUP-2396: Allow SensorRegion to pass actValue and bucketIdx to SDRClassifierRegion

marionleborgne:
This PR blocks #3520.

The network API does not support continuous inference at the moment. The workaround has been to call `customCompute()` on the classifier region to do inference.

This PR fixes the issue: the change enables the creation of new links between sensor and classifier regions so that the `SensorRegion` can pass `bucketIdx` and `actValue` to the `SDRClassifierRegion`. This way inference can happen within the network, without having to call `customCompute()` in an outer loop.

### DIFF
--- a/src/nupic/regions/RecordSensor.py
+++ b/src/nupic/regions/RecordSensor.py
@@ -108,6 +108,25 @@ class RecordSensor(PyRegion):
       singleNodeOnly=True,
       description="Sensor that reads data records and encodes them for an HTM",
       outputs=dict(
+        actValueOut=dict(
+          description="Actual value of the field to predict. The index of the "
+                      "field to predict must be specified via the parameter "
+                      "predictedFieldIdx. If this parameter is not set, then "
+                      "actValueOut won't be populated.",
+          dataType="Real32",
+          count=0,
+          regionLevel=True,
+          isDefaultOutput=False),
+        bucketIdxOut=dict(
+          description="Active index of the encoder bucket for the "
+                      "actual value of the field to predict. The index of the "
+                      "field to predict must be specified via the parameter "
+                      "predictedFieldIdx. If this parameter is not set, then "
+                      "actValueOut won't be populated.",
+          dataType="UInt64",
+          count=0,
+          regionLevel=True,
+          isDefaultOutput=False),
         dataOut=dict(
           description="Encoded data",
           dataType="Real32",  # inefficient for bits, but that's what we use now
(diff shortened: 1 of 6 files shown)


marionleborgne:
cc @scottpurdy and @rhyolight

src/nupic/regions/RecordSensor.py
```diff
@@ -344,6 +379,28 @@ def compute(self, inputs, outputs):
       # Encode the processed records; populate outputs["dataOut"] in place
       self.encoder.encodeIntoArray(data, outputs["dataOut"])
 
+      # If there is a field to predict, set bucketIdxOut and actValueOut.
+      if self.predictedFieldIdx > 0:
+        fields = self.dataSource.getFieldNames()
+        if self.predictedFieldIdx >= len(fields):
+          raise ValueError("predictedFieldIdx (%s) must be strictly inferior "
```
scottpurdy: "strictly inferior to the" -> "strictly less than the"

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -381,6 +402,22 @@ def compute(self, inputs, outputs):
                                     learn=self.learningMode,
                                     infer=False)
 
+      # If the input does not belong to a category, i.e. len(categories) == 0,
+      # then look for bucketIdx and ActValueIn.
```
scottpurdy: ActValueIn -> actValueIn

src/nupic/regions/SDRClassifierRegion.py
```diff
@@ -381,6 +402,22 @@ def compute(self, inputs, outputs):
...
+      if len(categories) == 0:
+        if "bucketIdxIn" not in inputs:
+          raise KeyError("Network link missing: bucketIdxOut -> bucketIdxIn")
+        if "actValue" not in inputs:
```
scottpurdy: Shouldn't this be "actValueIn"?

scottpurdy:
Also, can you include a test that exercises this behavior? One of my comments is on a line that looks like a bug so a test would help make sure all of this works as expected.

src/nupic/regions/SDRClassifierRegion.py
marionleborgne: Good catch, thanks!

marionleborgne:
Thanks for the review @scottpurdy 

To answer https://github.com/numenta/nupic/pull/3526#issuecomment-293026622, the "test" I am using to check that the new links work lives in this PR: https://github.com/numenta/nupic/pull/3520

Test usage:  `cd nupic/docs/examples/util; python check_predictions.py` to generate predictions on the Hot Gym dataset with OPF, network API, and research API. This compares the RMSE of the predictions for the 3 frameworks.

Would you rather have a unit test in nupic?

scottpurdy:
Yes, a unit test with a simple network that exercises the new functionality would be great. It could be an sdr classifier unit test.

marionleborgne:
@scottpurdy thanks for the review. I integrated the feedback and also added unit tests. It's ready for review again.

marionleborgne:
Fixes https://github.com/numenta/nupic/issues/3547

PR #3527, opened 2017-04-10
Fixes #3446

### LINKED ISSUE #3446: Potential API violation(s) in engine/__init__.py

abhinavrai44:
Fixes #3446

### DIFF
--- a/src/nupic/engine/__init__.py
+++ b/src/nupic/engine/__init__.py
@@ -270,25 +270,21 @@ class CollectionWrapper(object):
     return self.collection.contains(key)
 
   def keys(self):
-    keys = set()
+    keys = list()
     for i in range(self.collection.getCount()):
-      keys.add(self.collection.getByIndex(i)[0])
+      keys.append(self.collection.getByIndex(i)[0])
     return keys
 
   def values(self):
-    values = set()
+    values = list()
 
     for i in range(self.collection.getCount()):
       p = self.collection.getByIndex(i)
-      values.add(self.valueWrapper(p[0], p[1]))
+      values.append(self.valueWrapper(p[0], p[1]))
     return values
 
   def items(self):
-    items = set()
-    for i in range(self.collection.getCount()):
-      p = self.collection.getByIndex(i)
-      items.add((p[0], self.valueWrapper(p[0], p[1])))
-    return items
+    return zip(self.keys(), self.values())
 
   def __cmp__(self, other):
     return self.collection == other.collection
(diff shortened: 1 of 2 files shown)


oxtopus:
To further illustrate, this simple snippet should pass:

```python
from nupic.engine import CollectionWrapper

class Collection(list):

  def getCount(self):
    return len(self)

  def getByIndex(self, idx):
    return self[idx]

collection = Collection([(1, "One"), (2, "Two"), (3, "Three"), (4, "Four")])
wrappedCollection = CollectionWrapper(collection, valueWrapper=lambda x, y: y)

assert zip(wrappedCollection.keys(), wrappedCollection.values()) == collection
```

abhinavrai44:
@oxtopus Thanks for providing that code snippet. The implementation using list does pass the snippet provided.

rhyolight:
@abhinavrai44 Thanks again for the PR. Austin is on vacation until Monday, so he may not approve this soon.

abhinavrai44:
No problem. I'll try and work on some other issue

rhyolight:
@oxtopus please re-review

oxtopus:
Thanks!

PR #3537, opened 2017-04-11
Updates anomaly docstrings

### LINKED ISSUE #3536: Update anomaly related docstrings

rhyolight:
WIP

fixes #3536

### DIFF
--- a/src/nupic/algorithms/anomaly.py
+++ b/src/nupic/algorithms/anomaly.py
@@ -32,9 +32,9 @@ def computeRawAnomalyScore(activeColumns, prevPredictedColumns):
 
   The raw anomaly score is the fraction of active columns not predicted.
 
-  @param activeColumns: array of active column indices
-  @param prevPredictedColumns: array of columns indices predicted in prev step
-  @return anomaly score 0..1 (float)
+  :param activeColumns: array of active column indices
+  :param prevPredictedColumns: array of columns indices predicted in prev step
+  :returns: anomaly score 0..1 (float)
   """
   nActiveColumns = len(activeColumns)
   if nActiveColumns > 0:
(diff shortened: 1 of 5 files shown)


rhyolight:
@marionleborgne feel free to review, but technically I need a 👍 from a reviewer to merge.

src/nupic/algorithms/anomaly.py
```diff
@@ -56,48 +56,53 @@ def computeRawAnomalyScore(activeColumns, prevPredictedColumns):
...
 
 
-  def __init__(self, 
-               slidingWindowSize=None, 
-               mode=MODE_PURE, 
+  def __init__(self,
+               slidingWindowSize=None,
+               mode=MODE_PURE,
                binaryAnomalyThreshold=None):
-    """
```
marionleborgne: Why did you remove the doc string in the __init__?

rhyolight: I have been moving all init docs into the class docs because it works better with Sphinx.

scottpurdy: We should have good doc strings for `__init__` functions. Usually the only info you need here is the parameter documentation. But I don't think we should move that to the class doc string just because of docs tooling.

rhyolight: @scottpurdy Can I add a docstring in each `__init__` that points to the class definition? This seems like a common convention in python projects using Sphinx.

scottpurdy: After looking into the Sphinx issue, I don't have a strong opinion. You might want to check with others since this is something all engineers will have to be aware of. But I'm fine with whatever route you want to go.

PR #3549, opened 2017-04-14
Docs: InferenceShifter,ModelResult,SensorInput,InferenceType

### LINKED ISSUE #3548: Update api docs for InferenceShifter

rhyolight:
fixes #3548


src/nupic/frameworks/opf/opfutils.py
```diff
@@ -172,26 +183,34 @@ def isTemporal(inferenceType):
...
+  :param sequenceReset:  The sensor's "resetOut" signal (0 or 1) emitted by the
+                  sensor's compute logic on the supplied inputRecord; provided
+                  for analysis and diagnostics.
+
+  :param dataDict:
+  :param category:
+
+  HELP: This ``dataRow`` description above is wrong because there is no
+        getDecodedFieldMetaInfo() method anymore. I aso need someone to doc
+        ``dataDict`` and ``category``. -- Matt
```
rhyolight: @scottpurdy can you help here (or suggest someone to help)?

scottpurdy: It looks like this class is just an abstraction over the encoders. The `category` comes from `categoryOut` on the sensor region:

https://github.com/numenta/nupic/blob/914ada4fa719609e1523ca5c54c6e06b5b223bd3/src/nupic/frameworks/opf/htmpredictionmodel.py#L448

And in that code you can also see how `dataRow` comes from the sensor `sourceOut` while dataDict is the input to the sensor. Interestingly, these are both the unencoded values so I think they are just different forms of the same thing?

And the intent of this class isn't fully clear to me so hard to know the right thing to put in the documentation.

rhyolight:
@scottpurdy I'm happy with this if you are

Issue #3543, opened 2017-04-12
Remove old "UNDER_DEVELOPMENT" example code


rhyolight:

rhyolight:
I'm running into weird complications from removing some of these files. More PRs coming, just not sure when.

PR #3551, opened 2017-04-14
Remove old examples Part 1

rhyolight:
WIP

contributes to #3543

Removes:

- examples/opf/experiments/anomaly/spatial/10field_many_balanced
- examples/opf/experiments/anomaly/temporal/noisy_saw
- examples/opf/experiments/anomaly/temporal/saw_200_category
- examples/opf/experiments/classification/category_SP_0
- examples/opf/experiments/classification/category_SP_1


rhyolight:
@subutai or @scottpurdy please approve. I'm doing these a bit at a time.

scottpurdy:
I don't really know what things we may want to keep around. For instance, we may want more than hotgym for our regular testing. It looks like some of this stuff is simple artificial data. We can either dump that or keep it if we want to have a suite of algorithm-unit-tests so we can easily validate that new algorithm changes handle all the basic cases. But that sounds like some work and we can always come bring it back later.

Will defer to @subutai on the general question of whether we want to keep these misc. examples around.

rhyolight:
@scottpurdy Keep in mind these were all marked as `UNDER_DEVELOPMENT` and were not running as a part of our test suite.

Issue #3546, opened 2017-04-13
Update python file naming conventions to use snake_case


rhyolight:

subutai:
Where are our conventions now? They used to be on the wiki....

I believe filenames_are_snake (except Regions) but not sure about modules.  This is basically taken from Python conventions.

rhyolight:
@subutai I recently moved them to the API docs:

http://nupic.docs.numenta.org/contributing/python-style-guide.html#directory-and-file-names

PR #3561, opened 2017-04-24
Snake case module names PART 2

rhyolight:
Work In Progress

Contributes to #3546

- src/nupic/frameworks/opf/predictionmetricsmanager.py ==> src/nupic/frameworks/opf/prediction_metrics_manager.py
- src/nupic/frameworks/opf/previousvaluemodel.py ==> src/nupic/frameworks/opf/previous_value_model.py


marionleborgne:
The only thing is that the travis build gets killed. NUP-2423 should fix that.

PR #3575, opened 2017-05-02
Adds a script for versioned docs

rhyolight:
Used in Bamboo to build versioned docs.


rhyolight:
I am merging this because it is an outside script used by Bamboo to build docs.

PR #3583, opened 2017-05-03
Fix doc build folder copy problem

rhyolight:


rhyolight:
Just a doc script change

PR #3599, opened 2017-05-09
git rm old docs to prevent artifacts

rhyolight:
This change just removes any old versions of docs from git. Otherwise some old artifacts and mess up the doc build because the new version is moved over top of the old one.


rhyolight:
Just a CI script change, merging without review or CI.

PR #3601, opened 2017-05-09
NUP-2420 Renamed tm_shim.py to BacktrackingTM_shim.py

### LINKED ISSUE #3602: rename TM_shim.py
It should be like temporal_memory_shim.py ... so BacktrackingTM_shim.py would be better.

Yes, it doesn't match our naming conventions, but keep it exactly like this and we'll catch it as we standardize with find/replace.

vitaly-krugl:
Fixes #3602

From NUP-2420:
> It should be like temporal_memory_shim.py ... so BacktrackingTM_shim.py would be better.
Yes, it doesn't match our naming conventions, but keep it exactly like this and we'll catch it as we standardize with find/replace.


vitaly-krugl:
@rhyolight, please review

vitaly-krugl:
@rhyolight, regarding first_order_0.csv, it looks like there is a history of small updates, not unlike mine, being checked into it. It's possible that this file should have never been checked into the repo or some tests misbehave and overwrite it.

I am looking into it...

PR #3610, opened 2017-05-10
Moved opf_helpers module to helpers

### LINKED ISSUE #3609: move opf_helpers.py to helpers.py
See [TODO](https://github.com/numenta/nupic/blob/59ba7ee43aafbbd1308eb822fa58b7b221eff869/src/nupic/frameworks/opf/opf_helpers.py#L26-L27).

rhyolight:
fixes #3609


rhyolight:
@scottpurdy I'm not sure this is an idea we should follow through on. I honestly don't care either way. But here it is.

rhyolight:
Good point. Users can always `from nupic.frameworks.opf import helpers as opf_helpers`.

PR #3634, opened 2017-05-19
Updated install link in readme.

rhyolight:
fixes https://github.com/numenta/nupic/issues/3633

Ideally this link would go to the API docs, but I haven't created it yet. For now it's better than a bad link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,13 +50,15 @@ For any other installation issues, please see our [search our forums](https://di
 
 Live Community Chat: [![Gitter](https://img.shields.io/badge/gitter-join_chat-blue.svg?style=flat)](https://gitter.im/numenta/public?utm_source=badge)
 
-### Building NuPIC From Source Code
+### Installing NuPIC From Source
 
-To build from local source code, run from the repository root:
+To install from local source code, run from the repository root:
 
     pip install .
 
 Use the optional `-e` argument for a developer install.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,7 +52,7 @@ Live Community Chat: [![Gitter](https://img.shields.io/badge/gitter-join_chat-bl
 
 ### Building NuPIC From Source Code
 
-For details about checking out this repository and building in your local environment, see the [Installing and Building NuPIC](https://github.com/numenta/nupic/wiki/Installing-and-Building-NuPIC) wiki page.
+For details about checking out this repository and building in your local environment, see the [Installing and Building NuPIC](https://discourse.numenta.org/t/building-nupic/2192) wiki page.
```
scottpurdy: I think we can replace this with the command to install from repo, which is the same as the current install command but replacing "nupic" with "." when you're in the repo directory:

```
pip install .
```

Optionally, `-e` can be specified for a developer install.

rhyolight:
@scottpurdy ready

README.md
```diff
@@ -52,7 +52,11 @@ Live Community Chat: [![Gitter](https://img.shields.io/badge/gitter-join_chat-bl
 
 ### Building NuPIC From Source Code
```
scottpurdy: Perhaps change this to `Installing NuPIC from Source`? There isn't really a build for this pure-Python project.

Also, we might want to mention that if you want to build nupic.bindings from source that you should build and install from nupic.core prior to installing nupic (since a PyPI release will be installed if nupic.bindings isn't yet installed).

rhyolight:
@scottpurdy

PR #3638, opened 2017-05-22
Updated OPF Metric API docs

### LINKED ISSUE #3637: Update Metrics API docs

rhyolight:
fixes #3637


src/nupic/frameworks/opf/metrics.py
```diff
@@ -245,119 +319,116 @@ def _isNumber(value):
...
+    """ 
+    Add one instance consisting of ground truth and a prediction.
+
+    :param groundTruth:
+      The actual measured value at the current timestep
+    
+    :param prediction:
+      The value predicted by the network at the current timestep
+
+    :param record: ?
```
rhyolight: @scottpurdy Can you help me out with this?

src/nupic/frameworks/opf/metrics.py
```diff
@@ -245,119 +319,116 @@ def _isNumber(value):
...
   @abstractmethod
   def getMetric(self):
     """
-        return:
-            {value : <current measurement>, "stats" : {<stat> : <value> ...}}
-            metric name is defined by the MetricIface implementation. stats is expected to contain further
-                information relevant to the given metric, for example the number of timesteps represented in
-                the current measurement. all stats are implementation defined, and "stats" can be None
-
+    Metric name is defined by the MetricIface implementation. ``stats`` is
```
rhyolight: > Metric name is defined by the MetricIface implementation

@scottpurdy Does this statement make sense? I don't see how MetricIface defines a metric name.

src/nupic/frameworks/opf/metrics.py
scottpurdy: This is the input record as fed to the `Model.compute` method by the user. The typical usage is to feed a record to that method and get a `ModelResult`. Then you pass `ModelResult.rawInput` into this function as the `record` parameter.

src/nupic/frameworks/opf/metrics.py
scottpurdy: I have the same confusion as you. What is metric name? It looks safe to me to remove any reference to a metric name since I don't think there is one that is relevant to this function.

The return value is also unclear since you say it is an object but what kind? Is it an instance of a specific class or a dict? This is important because it determines whether you do `returnValue.value`, `returnValue["value"]`, or `returnValue.value()` to get the value form the returned object.

rhyolight:
@scottpurdy ready

PR #3639, opened 2017-05-22
Removed references to obsolete tm_py_fast shim

### LINKED ISSUE #3554: "tm_py_fast" is obsolete
When working on the renaming of TP* -> TM*, I noticed that `tm_py_fast` points to an implementation of TPShim that does not exist anymore (`FastTPShim`). See link below. We should probably remove this from the list.
https://github.com/numenta/nupic/blob/master/src/nupic/regions/TPRegion.py#L49

rhyolight:
fixes #3554


rhyolight:
@mrcslws

PR #3647, opened 2017-05-25
Removed topDownCompute fn from backtracking tm

### LINKED ISSUE #3646: Clean up confusing functions in BacktrackingTM
Discussion at https://discourse.numenta.org/t/backtrackingtm-topdowncompute-confusion/2352

rhyolight:
Fixes #3646.

### DIFF
--- a/examples/network/hierarchy_network_demo.py
+++ b/examples/network/hierarchy_network_demo.py
@@ -170,8 +170,6 @@ def createSpatialPooler(network, name, inputWidth):
 def createTemporalMemory(network, name):
   temporalMemoryRegion = network.addRegion(name, "py.TMRegion",
                                            json.dumps(TM_PARAMS))
-  # Enable topDownMode to get the predicted columns output
-  temporalMemoryRegion.setParameter("topDownMode", True)
   # Make sure learning is enabled (this is the default)
   temporalMemoryRegion.setParameter("learningMode", True)
   # Enable inference mode so we get predictions
--- a/examples/network/network_api_demo.py
+++ b/examples/network/network_api_demo.py
@@ -159,8 +159,6 @@ def createNetwork(dataSource):
 
   temporalPoolerRegion = network.regions["temporalPoolerRegion"]
 
-  # Enable topDownMode to get the predicted columns output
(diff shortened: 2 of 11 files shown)


rhyolight:
I don't think this is going to work.

rhyolight:
@subutai or @scottpurdy 

I'm tracing down the test failures on this one. What the heck is an `FDRTemporal`??

https://github.com/numenta/nupic/blob/master/src/nupic/regions/tm_region.py#L365

I see how `_tfdr` is created here:

https://github.com/numenta/nupic/blob/master/src/nupic/regions/tm_region.py#L420-L428

But doesn't that mean that all these implementations need to have a `topDownCompute` method because it is being called from `tm_region.py` here:

https://github.com/numenta/nupic/blob/master/src/nupic/regions/tm_region.py#L533

subutai:
"FDR" stands for Fixed Distributed Representations which was the name for the "new theory" in 2009. It became obsolete once we switched to CLA. Oh yeah, and that is now obsolete too.  Blame Jeff, not me.

I'm pretty sure we can get rid of 'topDownMode' in this region.

rhyolight:
Standby, a lot of trial and error involved...

scottpurdy:
It is still pretty useful to have the "predicted columns" available. I believe that is what "topDownOut" was providing from the TM. This can be used for custom anomaly detection, reconstruction (particularly useful in image classification for debugging), and other places. So I'd like to keep this functionality but "predictedColumns" might be a better name for the region output.

subutai:
> It is still pretty useful to have the "predicted columns" available.

Yeah.  In the research code we now have a bunch of these available as separate outputs.

https://github.com/numenta/htmresearch/blob/master/htmresearch/regions/TMRegion.py#L104

rhyolight:
> It is still pretty useful to have the "predicted columns" available. I believe that is what "topDownOut" was providing from the TM.

It wasn't actually doing that. It was just returning the `columnConfidences`.

rhyolight:
I'm not sure I'm going to continue with this refactoring in light of the [rabbit](https://github.com/numenta/nupic/issues/3648) [hole](https://github.com/numenta/nupic/pull/3658) I've been going down.

Issue #3225, opened 2016-07-27
SDRClassifierRegion does not play well with scalar value prediction


andrewmalta13:
Discussed with Yuwei and we agreed that the SDRClassifierRegion inherited some problems from the CLAClassifierRegion and that it does not work with multistep scalar prediction.

<a href="https://github.com/numenta/nupic/blob/master/src/nupic/regions/CLAClassifierRegion.py#L380"> This </a>  is the line we discussed that was inherited from the CLAClassifier that creates this problem. 

A related <a href="https://github.com/numenta/nupic/issues/3078"> issue  </a> from someone experiencing the same problem I had.

@ywcui1990 what would be your proposal to making this region work both for traditional classification and scalar value prediction?

@rhyolight I am putting the prediction network api example on hold until this is resolved.

natoromano:
Here is my suggested fix:
- `SDRClassifierRegion` should have an _optional_ additional input, `valueIn`
- the `RecordSensor` region should have an _optional_ additional output, `predictedFieldOut`

The `RecordSensor` can use its reference to its encoder to determine a good `bucketIdx` for each value, and use that as `categoryOut`, as well as send the predicted field in `predictedValueOut`. That would perhaps require adding an `.enablePrediction(predictionArgs)` to the `RecordSensor`, much like is done [here](https://github.com/numenta/nupic/blob/master/src/nupic/frameworks/opf/model.py#L155) for the OPF framework. `predictionArgs` can just be the name of the field, or have additional information (number of buckets to use if no encoder is used?).
Of course, a flag would be set by `enablePrediction`, and nothing is sent to `predictedValueOut` if this flag is not set.

As for the `SDRClassifierRegion`, it can do what it currently does if there is no `predictionInput`, i.e. use `int(category)` for both the bucket ID and actual value. If there is a `predictionInput`, it uses `int(category)` for the bucket ID (which, if connected to a sensor, is optimal as it is the same bucket as in the encoder), and `predictionInput` for the actual value.

@ywcui1990 @scottpurdy @andrewmalta13

scottpurdy:
Alternatively, the main implementation could just be for categories and a subclass could additionally do the bucket->scalar conversion. This or @natoromano 's proposal are fine with me. His is simpler.

andrewmalta13:
The next person who picks up this issue should check out some preliminary work I did on this branch: https://github.com/andrewmalta13/nupic/tree/scalarPrediction

[...]

PR #3698, opened 2017-06-12
Handle scalar values in the sdr classifier region

ywcui1990:
Fixes: https://github.com/numenta/nupic/issues/3225

@scottpurdy @rhyolight  Please review.

I changed the SDR classifier region to handle scalar values when it is available.
I also added a test that uses the SDR classifier region for scalar predictions.


ywcui1990:
@scottpurdy Feel free to merge at any time.

PR #3710, opened 2017-06-15
Added Serializable to all classes with a capnp write function

rhyolight:
There is a problem with this PR, tests will fail with this error, but I'm not sure why:

```
============================================== ERRORS ==============================================
_________________________ ERROR collecting tests/unit/nupic/utils_test.py __________________________
tests/unit/nupic/utils_test.py:28: in <module>
    from nupic.utils import MovingAverage
src/nupic/utils.py:32: in <module>
    class MovingAverage(object, Serializable):
/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/abc.py:87: in __new__
    cls = super(ABCMeta, mcls).__new__(mcls, name, bases, namespace)
E   TypeError: Error when calling the metaclass bases
E       Cannot create a consistent method resolution
E   order (MRO) for bases object, Serializable
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
===================================== 1 error in 0.22 seconds ======================================
```

### DIFF
--- a/src/nupic/utils.py
+++ b/src/nupic/utils.py
@@ -27,6 +27,7 @@ in our codebase.
 import numbers
 
 from nupic.serializable import Serializable
+from nupic.movingaverage_capnp import MovingAverageProto
 
 
 class MovingAverage(Serializable):
@@ -132,5 +133,5 @@ class MovingAverage(Serializable):
 
   @classmethod
   def getSchema(cls):
-    raise NotImplementedError()
+    return MovingAverageProto
 
--- a/src/nupic/utils.py
+++ b/src/nupic/utils.py
@@ -128,3 +128,9 @@ class MovingAverage(object, Serializable):
     proto.windowSize = self.windowSize
     proto.slidingWindow = self.slidingWindow
     proto.total = self.total
+
+
+  @classmethod
+  def getSchema(cls):
+    raise NotImplementedError()
+
--- a/src/nupic/utils.py
+++ b/src/nupic/utils.py
@@ -27,7 +27,12 @@ in our codebase.
 import numbers
 
 from nupic.serializable import Serializable
-from nupic.movingaverage_capnp import MovingAverageProto
+try:
+  import capnp
+except ImportError:
+  capnp = None
+if capnp:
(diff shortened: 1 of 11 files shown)


lscheinkman:
The [MovingAverage](https://github.com/rhyolight/nupic/blob/d94ff7f4f747fbe6d6e994e908cb39fcf51a6880/src/nupic/utils.py#L32) class does not implement the abstract method [getSchema](
https://github.com/rhyolight/nupic/blob/d94ff7f4f747fbe6d6e994e908cb39fcf51a6880/src/nupic/serializable.py#L36-L38)

rhyolight:
@lscheinkman Even after that change, I still get the same error on MovingAverage.

lscheinkman:
remove `object` from the mixin list. `Serializable` already inherits from `object`.
```
class MovingAverage(Serializable):
  """
```

rhyolight:
Thanks @lscheinkman that was it.

src/nupic/utils.py
```diff
@@ -127,3 +128,9 @@ def write(self, proto):
     proto.windowSize = self.windowSize
     proto.slidingWindow = self.slidingWindow
     proto.total = self.total
+
+
+  @classmethod
+  def getSchema(cls):
+    raise NotImplementedError()
```
lscheinkman: Should return `MovingAverageProto` from this import
```
from nupic.movingaverage_capnp import MovingAverageProto
```

rhyolight:
@lscheinkman ready

rhyolight:
investigating test failure 🤷‍♂️

rhyolight:
Windows built error doesn't make sense. Re-running to see what happens.

lscheinkman:
You need to add capnp conditionals for Windows. Something like this:
```
try:
   import capnp
 except ImportError:
   capnp = None
if capnp:
  from nupic.movingaverage_capnp import MovingAverageProto
```

rhyolight:
@lscheinkman thanks for the help

PR #3715, opened 2017-06-16
Add capnp serialization to BacktrackingTMCPP

scottpurdy:
depends on numenta/nupic.core#1351 (plus a nupic.bindings release afterward)

### DIFF
--- a/src/nupic/algorithms/backtracking_tm.py
+++ b/src/nupic/algorithms/backtracking_tm.py
@@ -426,6 +426,10 @@ class BacktrackingTM(ConsolePrinterMixin, Serializable):
 
 
   def write(self, proto):
+    """Populate serialization proto instance.
+
+    :param proto: (BacktrackingTMProto) the proto instance to populate
+    """
     proto.version = TM_VERSION
     self._random.write(proto.random)
     proto.numberOfCols = self.numberOfCols
@@ -470,6 +474,7 @@ class BacktrackingTM(ConsolePrinterMixin, Serializable):
     proto.pamCounter = self.pamCounter
     proto.collectSequenceStats = self.collectSequenceStats
     proto.resetCalled = self.resetCalled
+    # In case of None, use negative value as placeholder for serialization
     proto.avgInputDensity = self.avgInputDensity or -1.0
     proto.learnedSeqLength = self.learnedSeqLength
     proto.avgLearnedSeqLength = self.avgLearnedSeqLength
@@ -525,6 +530,10 @@ class BacktrackingTM(ConsolePrinterMixin, Serializable):
 
   @classmethod
   def read(cls, proto):
+    """Deserialize from proto instance.
+
+    :param proto: (BacktrackingTMProto) the proto instance to read from
+    """
     assert proto.version == TM_VERSION
     obj = object.__new__(cls)
     obj._random = Random()
@@ -573,10 +582,12 @@ class BacktrackingTM(ConsolePrinterMixin, Serializable):
     obj.pamCounter = int(proto.pamCounter)
     obj.collectSequenceStats = proto.collectSequenceStats
     obj.resetCalled = proto.resetCalled
-    if proto.avgInputDensity < 0.0:
+    avgInputDensity = proto.avgInputDensity
+    if avgInputDensity < 0.0:
+      # Negative value placeholder indicates None
       obj.avgInputDensity = None
     else:
-      obj.avgInputDensity = proto.avgInputDensity
+      obj.avgInputDensity = avgInputDensity
     obj.learnedSeqLength = int(proto.learnedSeqLength)
     obj.avgLearnedSeqLength = proto.avgLearnedSeqLength
 
--- a/src/nupic/algorithms/backtracking_tm.py
+++ b/src/nupic/algorithms/backtracking_tm.py
@@ -1396,8 +1396,6 @@ class BacktrackingTM(ConsolePrinterMixin, Serializable):
 
       obj.columnIdx = proto.columnIdx
       obj.cellIdx = proto.cellIdx
-      # TODO: Is it ok for this to be a separate instance than the segment
-      # in the TM segment list?
       obj.segment.read(proto.segment, tm)
       obj.activeSynapses = [syn for syn in proto.activeSynapses]
       obj.sequenceSegment = proto.sequenceSegment
--- a/src/nupic/algorithms/backtracking_tm.py
+++ b/src/nupic/algorithms/backtracking_tm.py
@@ -470,7 +470,7 @@ class BacktrackingTM(ConsolePrinterMixin, Serializable):
     proto.pamCounter = self.pamCounter
     proto.collectSequenceStats = self.collectSequenceStats
     proto.resetCalled = self.resetCalled
-    proto.avgInputDensity = self.avgInputDensity
+    proto.avgInputDensity = self.avgInputDensity or -1.0
     proto.learnedSeqLength = self.learnedSeqLength
     proto.avgLearnedSeqLength = self.avgLearnedSeqLength
 
@@ -563,8 +563,7 @@ class BacktrackingTM(ConsolePrinterMixin, Serializable):
       for cellSegments, cellSegmentsProto in zip(columnSegments,
                                                  columnSegmentsProto):
         for segmentProto in cellSegmentsProto:
-          segment = Segment.read(segmentProto)
-          segment.tm = obj
+          segment = Segment.read(segmentProto, obj)
           cellSegments.append(segment)
 
     obj.lrnIterationIdx = int(proto.lrnIterationIdx)
@@ -574,7 +573,10 @@ class BacktrackingTM(ConsolePrinterMixin, Serializable):
     obj.pamCounter = int(proto.pamCounter)
     obj.collectSequenceStats = proto.collectSequenceStats
     obj.resetCalled = proto.resetCalled
-    obj.avgInputDensity = proto.avgInputDensity
+    if proto.avgInputDensity < 0.0:
+      obj.avgInputDensity = None
+    else:
+      obj.avgInputDensity = proto.avgInputDensity
     obj.learnedSeqLength = int(proto.learnedSeqLength)
     obj.avgLearnedSeqLength = proto.avgLearnedSeqLength
 
@@ -592,7 +594,7 @@ class BacktrackingTM(ConsolePrinterMixin, Serializable):
       key = (cellWrapperProto.columnIdx, cellWrapperProto.cellIdx)
       value = []
       for updateWrapperProto in cellWrapperProto.segmentUpdates:
-        segmentUpdate = SegmentUpdate.read(updateWrapperProto.segmentUpdate)
+        segmentUpdate = SegmentUpdate.read(updateWrapperProto.segmentUpdate, obj)
         value.append((int(updateWrapperProto.lrnIterationIdx), segmentUpdate))
       obj.segmentUpdates[key] = value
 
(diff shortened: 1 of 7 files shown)


src/nupic/algorithms/backtracking_tm.py
```diff
@@ -470,7 +470,7 @@ def write(self, proto):
     proto.pamCounter = self.pamCounter
     proto.collectSequenceStats = self.collectSequenceStats
     proto.resetCalled = self.resetCalled
-    proto.avgInputDensity = self.avgInputDensity
+    proto.avgInputDensity = self.avgInputDensity or -1.0
```
vitaly-krugl: Are you guarding against zero or None? It would be good to have a descriptive comment to help reader/maintainer

src/nupic/algorithms/backtracking_tm.py
```diff
@@ -574,7 +573,10 @@ def read(cls, proto):
     obj.pamCounter = int(proto.pamCounter)
     obj.collectSequenceStats = proto.collectSequenceStats
     obj.resetCalled = proto.resetCalled
-    obj.avgInputDensity = proto.avgInputDensity
+    if proto.avgInputDensity < 0.0:
```
vitaly-krugl: Add a comment that correlates this special case to the `write` method?

vitaly-krugl: Also, not sure if we care in this case, but I think I've seen recommendations to access proto members as few times as possible for performance reasons (I think it parses each time). In typical case, we'll read `proto.avgInputDensity` first to check for negative, then second time to assign it to `obj.avgInputDensity`. IIRC, the recommended practice was to save the value in a variable and then compare, assign, etc.

src/nupic/algorithms/backtracking_tm.py
```diff
@@ -1355,7 +1357,7 @@ def getSegmentOnCell(self, c, i, segIdx):
     return retlist
 
 
-  class _SegmentUpdate(Serializable):
+  class _SegmentUpdate(object):
```
vitaly-krugl: Why don't we want `Serializable` base class here any longer? Doesn't it break our goal of having all such classes based on `Serializable`?

I can see that you changed the methods (+ additional arg), so that they are not compatible with `Serializable` interface. Does this mean that `Serializable` is not sufficiently flexible for our use cases?

Same in `Segment` class...

src/nupic/algorithms/backtracking_tm_cpp.py
```diff
@@ -154,6 +163,44 @@ def __init__(self,
...
+    proto.seed = self.seed
+    proto.checkSynapseConsistency = self.checkSynapseConsistency
+    proto.initArgs = json.dumps(self._initArgsDict)
+
+
+  @classmethod
+  def read(cls, proto):
+    # Use base class to create initial class
+    obj = BacktrackingTM.read(proto.baseTM)
+    obj.__class__ = cls
```
vitaly-krugl: I am not enamored with this.  Messing with its class metadata sounds to me like asking for trouble.

src/nupic/algorithms/backtracking_tm_cpp.py
```diff
@@ -154,6 +163,44 @@ def __init__(self,
...
+
+    # Additional CPP-specific deserialization
+    newCells4 = Cells4.read(proto.cells4)
+    print newCells4
+    obj.cells4 = newCells4
+    obj.makeCells4Ephemeral = proto.makeCells4Ephemeral
+    obj.seed = proto.seed
+    obj.checkSynapseConsistency = proto.checkSynapseConsistency
+    obj._initArgsDict = json.loads(proto.initArgs)
+    obj._initArgsDict["outputType"] = str(obj._initArgsDict["outputType"])
```
vitaly-krugl: Is this casting away unicode? How about a short comment to that or other effect?  I think there may be an arg in `json.loads` that generates strings instead of unicode.

src/nupic/algorithms/backtracking_tm_cpp.py
```diff
@@ -154,6 +163,44 @@ def __init__(self,
                             )
 
 
+  @classmethod
+  def getSchema(cls):
+    return BacktrackingTMCppProto
+
+
+  def write(self, proto):
+    super(BacktrackingTMCPP, self).write(proto.baseTM)
```
vitaly-krugl: Document proto arg's class name?

src/nupic/algorithms/backtracking_tm_cpp.py
```diff
@@ -154,6 +163,44 @@ def __init__(self,
...
+    self.cells4.write(proto.cells4)
+    proto.makeCells4Ephemeral = self.makeCells4Ephemeral
+    proto.seed = self.seed
+    proto.checkSynapseConsistency = self.checkSynapseConsistency
+    proto.initArgs = json.dumps(self._initArgsDict)
+
+
+  @classmethod
+  def read(cls, proto):
```
vitaly-krugl: Document proto args class name?

src/nupic/regions/tm_region.capnp
```diff
@@ -1,11 +1,17 @@
...
 
-# Next ID: 11
+# Next ID: 13
 struct TMRegionProto {
   temporalImp @0 :Text;
-  temporalMemory @1 :TemporalMemoryProto;
+  union {
+    temporalMemory @1 :TemporalMemoryProto;
+    backtrackingTM @11 :BacktrackingTMProto;
+    backtrackingTMCpp @12 :BacktrackingTMCppProto;
```
vitaly-krugl: Field numbers look strange. I think you're trying to leave a gap for existing 2, 3, etc. But do we really need to? Since we're breaking compatibility on the way to NuPIC 1.0, we might as well have straight-forward numbering and not worry about compatibility?

src/nupic/regions/tm_region.py
```diff
@@ -805,7 +805,22 @@ def writeToProto(self, proto):
...
+      tmProto = proto.init("backtrackingTM")
+      self._tfdr.write(tmProto)
+    elif self.temporalImp == "cpp":
+      tmProto = proto.init("backtrackingTMCpp")
+      self._tfdr.write(tmProto)
+    elif self.temporalImp == "tm_py":
+      tmProto = proto.init("temporalMemory")
+      self._tfdr.write(tmProto)
+    elif self.temporalImp == "tm_cpp":
+      tmProto = proto.init("temporalMemory")
```
vitaly-krugl: There is no "temporalMemoryCpp"?

src/nupic/regions/tm_region.py
```diff
@@ -805,7 +805,22 @@ def writeToProto(self, proto):
...
-    self._tfdr.write(proto.temporalMemory)
+    if self.temporalImp == "py":
+      tmProto = proto.init("backtrackingTM")
+      self._tfdr.write(tmProto)
+    elif self.temporalImp == "cpp":
+      tmProto = proto.init("backtrackingTMCpp")
+      self._tfdr.write(tmProto)
+    elif self.temporalImp == "tm_py":
+      tmProto = proto.init("temporalMemory")
+      self._tfdr.write(tmProto)
```
vitaly-krugl: Why not have a single `self._tfdr.write(tmProto)` call, versus repeating in each clause?

src/nupic/regions/tm_region.py
```diff
@@ -805,7 +805,22 @@ def writeToProto(self, proto):
       self.computePredictedActiveCellIndices)
     proto.orColumnOutputs = self.orColumnOutputs
 
-    self._tfdr.write(proto.temporalMemory)
+    if self.temporalImp == "py":
```
vitaly-krugl: Do we want to add tests for these new code paths?

src/nupic/regions/tm_region.py
```diff
@@ -828,7 +843,18 @@ def readFromProto(cls, proto):
       proto.computePredictedActiveCellIndices)
     instance.orColumnOutputs = proto.orColumnOutputs
 
-    instance._tfdr = _getTPClass(proto.temporalImp).read(proto.temporalMemory)
+    if instance.temporalImp == "py":
```
vitaly-krugl: Do we want to add tests for these new code paths?

src/nupic/algorithms/backtracking_tm.py
scottpurdy: Segment and SegmentUpdate are not intended to be used outside BacktrackingTM - they are implementation details. So we don't need the common serialization interface for them. Additionally, they require additional information for deserialization, so the interface is necessarily different. Do you have any suggested changes or just curious?

vitaly-krugl: Curious and no suggestions

vitaly-krugl: Perhaps a short comment to that effect to appease curious Georges in the future?

src/nupic/algorithms/backtracking_tm_cpp.py
scottpurdy: Unless you have some concrete argument against it then I think it is the cleanest method. The alternative is to instantiate a new class instance via `object.__new__(cls)` and copy `obj.__dict__` into it. But that results in the full instantiation of an extra temporary class. Or you can create an instance method for reading results into an existing class but that's a very different approach to how the rest of capnp serialization is implemented.

src/nupic/algorithms/backtracking_tm_cpp.py
scottpurdy: I don't see a way to do this. Happy to change it if you can point me the to the method but I've added a comment for now.

src/nupic/regions/tm_region.capnp
scottpurdy: We don't have to worry about renumbering because we aren't maintaining backwards compatibility. But there is no advantage to renumbering all of them. Actually I worry that perfectly number fields in all of our schema will lead people to think that that order must be maintained. So I could change it but not sure I'm achieving anything positive in doing that.

src/nupic/regions/tm_region.py
scottpurdy: Isn't that what htmpredictionmodel_serialization_test is for? The regions aren't easily tested outside a network.

src/nupic/regions/tm_region.py
scottpurdy: Good idea. I changed it here and in `read` to match.

src/nupic/regions/tm_region.py
scottpurdy: I believe the Python TemporalMemory uses the TemporalMemory.capnp file from nupic.bindings

scottpurdy:
@vitaly-krugl - I've responded to or addressed all your comments. Please review again at your leisure.

src/nupic/algorithms/backtracking_tm_cpp.py
vitaly-krugl: I just didn't have confidence that modifying the class `obj.__class__ = cls` wouldn't cause subtle issues. I will accept this if it's accepted practice in python.

src/nupic/regions/tm_region.capnp
vitaly-krugl: okay

src/nupic/regions/tm_region.py
vitaly-krugl: I think it's testing just cpp and py, not the other 2 variations

src/nupic/algorithms/backtracking_tm_cpp.py
vitaly-krugl: comment is fine

scottpurdy:
@vitaly-krugl - I think we should have some network-level tests for TMRegion that tests each of the four implementation options. I'll save that for another PR (but part of same JIRA).

vitaly-krugl:
@scottpurdy --

> I think we should have some network-level tests for TMRegion that tests each of the four implementation options. I'll save that for another PR (but part of same JIRA).

Yes, I agree

PR #3731, opened 2017-06-28
fix asterisk password bug in test_db.py

pianomanfrazier:
The password is being submitted as all asterisks to the database instead of the actual password.

### DIFF
--- a/examples/swarm/test_db.py
+++ b/examples/swarm/test_db.py
@@ -83,7 +83,7 @@ def dbValidator():
   host = Configuration.get("nupic.cluster.database.host")
   port = int(Configuration.get("nupic.cluster.database.port"))
   user = Configuration.get("nupic.cluster.database.user")
-  passwd = "*" * len(Configuration.get("nupic.cluster.database.passwd"))
+  passwd = Configuration.get("nupic.cluster.database.passwd")
 
 
   print "This script will validate that your MySQL is setup correctly for "
@@ -103,7 +103,7 @@ def dbValidator():
   print "    host   :    ", host
   print "    port   :    ", port
   print "    user   :    ", user
-  print "    passwd :    ", passwd
+  print "    passwd :    ", "*" * len(passwd)
 
 
   if testDbConnection(host, port, user, passwd):
--- a/examples/swarm/test_db.py
+++ b/examples/swarm/test_db.py
@@ -65,10 +65,10 @@ def testDbConnection(host, port, user, passwd):
     cursor.execute("INSERT INTO db_test VALUES ('testing123', 123)")
     cursor.execute("DROP TABLE IF EXISTS db_test")
     cursor.execute("DROP DATABASE IF EXISTS nupic_db_test")
-    return True
 
   except pymysql.err.OperationalError:
-    return False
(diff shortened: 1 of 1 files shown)


rhyolight:
Wow, how could this have been working? cc @scottpurdy @vitaly-krugl ?

vitaly-krugl:
Interns!! :)

vitaly-krugl:
@pianomanfrazier, @rhyolight: Also, the existing logic is not at all pythonic. The call to `testDbConnection` returns True/False instead of raising an exception. So, even if this example test is running in CI, it would simply print a message on failure, but not communicate the failure via process exist code, so higher-level test driver has no clue that something is amiss.

rhyolight:
Thanks Vitaly. @pianomanfrazier let me know of you want to go the extra mile as suggested, otherwise I will merge this as-is.

rhyolight:
Well shit, I screwed up. @pianomanfrazier You are supposed to sign our [contributor license](https://numenta.org/licenses/contrib/) before I merge. Do you mind doing that now?

Just in case, I reverted this accidental merge (all I can do at this point). @pianomanfrazier Once you sign the license I'll take care of this and get it re-merged. 

Sorry for the hassle! Apparently there is something wrong with our automated contributor validator. I'll be looking into it.

pianomanfrazier:
Just signed the license.

rhyolight:
Thanks, it takes a bit to process. I'll check back in later.

examples/swarm/test_db.py
```diff
@@ -65,10 +65,10 @@ def testDbConnection(host, port, user, passwd):
     cursor.execute("INSERT INTO db_test VALUES ('testing123', 123)")
     cursor.execute("DROP TABLE IF EXISTS db_test")
     cursor.execute("DROP DATABASE IF EXISTS nupic_db_test")
-    return True
 
   except pymysql.err.OperationalError:
-    return False
+    print "Error: failed to connect to database"
+    raise pymysql.err.OperationalError
```
vitaly-krugl: This isn't right. We want to re-raise the original exception instance, including the message. Just use `raise` from within the exception handler block without any arg to do this.

examples/swarm/test_db.py
```diff
@@ -103,12 +103,13 @@ def dbValidator():
...
-  print "    passwd :    ", passwd
+  print "    passwd :    ", "*" * len(passwd)
 
 
-  if testDbConnection(host, port, user, passwd):
+  try: 
+    testDbConnection(host, port, user, passwd)
     print "Connection successful!!"
-  else:
+  except pymysql.err.OperationalError:
```
vitaly-krugl: If you don't re-raise the exception after printing the text, then higher level program that invokes this script has no way to find out that something went wrong. Actually, just remove the try/except here and move printing of the test to the `pymysql.err.OperationalError` handler in `testDbConnection` (instead of `print "Error: failed to connect to database"` there).

rhyolight:
@pianomanfrazier I have your CL now. I'm not sure you can re-open this PR, but you can push your changes to a new remote branch and create a new PR to address @vitaly-krugl's changes.

PR #3737, opened 2017-07-06
Added serialization guide to API docs.

rhyolight:
NUP-2467


rhyolight:
@scottpurdy Should we be seeing a CircleCI build now?

scottpurdy:
@rhyolight - I would think we should see circle ci, let me look

PR #3741, opened 2017-07-06
DEVOPS-362 Fix Circle

MaxASchwarzer:
Integrates some infrastructure code from nupic.core for use in our CircleCI builds.


MaxASchwarzer:
Note that this doesn't actually make Circle builds _succeed_, as some tests are failing.  It does, however, allow Circle to run tests.

PR #3754, opened 2017-08-08
Check for "vector" predicted field before record sensor compute.

### LINKED ISSUE #3753: Encoders using vector input like CoordinateEncoder do not work anymore
I found this issue when trying to get https://github.com/numenta/nupic.geospatial running with NuPIC 1.0. 

Looks like this is the origin of the bug:

https://github.com/numenta/nupic/commit/34dd874896b22da217c5bede494f910faf6efb99#diff-d62e7e036bc450e0076245d0839b7ef4R381

In this commit, we start using `self.predictedField` as a condition to get encoder outputs instead of `self.predictedFieldIdx`. The CoordinateEncoder's `predictedFieldIdx` is always `-1`, so it worked before the change. But the CoordinateEncoders' `predictedField` is not `None`, it is `vector`, so the compute function tries to get bucket values from the encoder that do not exist. 

We need to add a special condition to the compute function to check also for `self.predictedField == "vector"`.

rhyolight:
Fixes #3753

### DIFF
--- a/src/nupic/regions/record_sensor.py
+++ b/src/nupic/regions/record_sensor.py
@@ -377,7 +377,10 @@ class RecordSensor(PyRegion):
       self.encoder.encodeIntoArray(data, outputs["dataOut"])
 
       # If there is a field to predict, set bucketIdxOut and actValueOut.
-      if self.predictedField is not None:
+      # There is a special case where a predicted field might be a vector, as in
+      # the CoordinateEncoder. Since this encoder does not provide bucket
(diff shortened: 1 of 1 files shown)


rhyolight:
@scottpurdy This fix feels hacky to me, do you have a better way in mind to fix this?

rhyolight:
test

lscheinkman:
@rhyolight Should we add a test to cover this condition?

PR #3762, opened 2017-08-14
Added conditions for IO for coord encoder

### LINKED ISSUE #3761: CoordinateEncoder usage still broken in some configurations
See this long discussion starting from https://discourse.numenta.org/t/debugging-coordinateencoder-issues-in-nupic-1-0/2656/13?u=rhyolight

The issue is that the CoordinateEncoder doesn't provide a getBucketIndices function like it is supposed to. It was just never implemented. And there are no special cases coded into the HTMPredictionModel like there are for some other encoders to handle things like this.

The HtmPredictionModel in the OPF needs to be updated to do some checking before constructing the I/O objects.

rhyolight:
Fixes #3761

This is bad code practice, but I don't have time to make a clean fix.
The right answer is to simply provide an implementation of
getBucketIndices for the CoordinateEncoder, but I honestly don't know
how to do it. This is a quick fix.

### DIFF
--- a/src/nupic/frameworks/opf/htm_prediction_model.py
+++ b/src/nupic/frameworks/opf/htm_prediction_model.py
@@ -40,7 +40,7 @@ import numpy
 from nupic.frameworks.opf.model import Model
 from nupic.data import SENTINEL_VALUE_FOR_MISSING_DATA
 from nupic.data.field_meta import FieldMetaSpecial, FieldMetaInfo
-from nupic.encoders import MultiEncoder, DeltaEncoder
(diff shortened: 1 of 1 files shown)


rhyolight:
Goal is to make it easier for users to construct models. At least one user wants to have more encoder options in the OPF. The OPF is already full of special cases like this, and we're not planning on fixing them. So I don't see the harm of adding CoordinateEncoder support. It was not hard to do, and I'll make it clear in our forums that this new functionality of the OPF does not include prediction.

rhyolight:
Ok, @scottpurdy just figured out what was wrong in the original report, and we don't actually have to do this.

PR #3784, opened 2017-12-28
Changed distribution keyword, casted some attributes to float, remove…

kyle-sorensen:
This fixes #3783, but there were other issues discovered underneath these fixes. 

1. The incoming object to be serialized has the structure `self._distribution['distribution']['mean']`, `self._distribution['distribution']['stdev']` etc, but the write method expects `self._distribution['distributionParams']['mean']` (anomaly_likelihood.py line 318-321)

2. The read method did not expect `distribution` or `distributionParams` (anomaly_likelihood.py line 274-279) thus when the serialized object was read back into memory it did not have a matching structure. 

3. The incoming values cannot be serialized by capnproto because they have the type `numpy.float64`, so we cast them to `float` (which is a double) (anomaly_likelihood.py line 318-321)

4. The AnomalyLikelihoodRegion.capnp file uses `Float32` (not a double) instead of `Float64`.

@rhyolight I'm one of the developers working on Grok, we need to use these serialization methods, if you want some time to talk on the phone about these changes I would be happy to do that. There is some confusion here as to how you guys are serializing datetimes, looks like you just save an integer in place of it to preserve order, (why not serialize a timestamp and recreate the datetime object?)
Please do look over this code though, because it throws a fatal KeyError as is. Greatly appreciate all the help given on the forums!

Issue #3783, opened 2017-12-27
Anomalylikelihood write method throws KeyError distributionParams vs distribution


kyle-sorensen:
```
Traceback (most recent call last):
  File "run.py", line 21, in <module>
    model.write(builder)
  File "/usr/local/lib/python2.7/site-packages/nupic/algorithms/anomaly_likelihood.py", line 321, in write
    proto.distribution.name = self._distribution["distributionParams"]["name"]
KeyError: 'distributionParams'
```

When I log this method, I can see the following:

```

{'distribution': {'mean': 0.66554064258358092,
                  'name': 'normal',
                  'stdev': 0.02710835571806828,
                  'variance': 0.00073486294973732511},
 'historicalLikelihoods': [0.08321225454772424,
                           0.42466239667745609,
                           0.41143882167909385,
                           0.38278323468141645,
                           0.097795022725065406,
                           0.42578440342970869,
                           0.47160308710543392,
                           0.22586380006966755,
                           0.43450092587880484,
                           0.29902835722605992],
 'movingAverage': {'historicalValues': [0.65061007991758246,
                                        0.92661753892022314,
                                        0.5091802474129109,
                                        0.52021695705560211,
                                        0.54556075732431397,
                                        0.91294922102966003,
                                        0.5928150752511081,
                                        0.5054987548601162,
                                        0.78161115646868851,
                                        0.56743353987350631],
                   'total': 6.5124933281137114,
                   'windowSize': 10}}
```

In the proto object the defined key is "distribution" too, not "distributionParams". 

The capnproto schema also uses Float32 (a float) but the incoming python "float" is actually a double. (Should this be Float64?) 

```
@0x8602f38429407eb0;

struct AnomalyLikelihoodRegionProto {
  iteration @0 :UInt64;
  historicalScores @1 :List(Score);
  distribution @2 :Distribution;
  probationaryPeriod @3 :UInt32;
  learningPeriod @4 :UInt32;
  reestimationPeriod @5 :UInt32;
  historicWindowSize @6 :UInt32;

  struct Score {
    value @0 :Float32;
    anomalyScore @1 :Float32;
  }

  struct Distribution {
    name @0 :Text;
    mean @1 :Float32;
    variance @2 :Float32;
    stdev @3 :Float32;
    movingAverage @4 :MovingAverage;
    historicalLikelihoods @5 :List(Float32);

    struct MovingAverage {
      windowSize @0 :UInt64;
      historicalValues @1 :List(Float32);
[...]

PR #3786, opened 2018-01-02
Issue #3783: Fixes test to compute pass the probation period

lscheinkman:
Fixes anomaly likelihood serialization test to include distribution calculation.
It should only be merged after #3784


rhyolight:
Ignore this validation failure:
- **continuous-integration/bamboo** because we know it is not running on PR, which is ok as long as the other two CI builds pass.

PR #3799, opened 2018-01-20
NUP-2487: Update category prediction example

### LINKED ISSUE #3790: investigate category prediction example problems
The example at `examples/prediction/category_prediction` doesn't seem to be running properly anymore. The generated `tokens.txt` file is dirty, and I don't understand what the results file is actually showing. I think this example needs to be rewritten.

lscheinkman:
Fixes #3790 
Replaces the `Word Prediction` example with a simple `Web site link prediction` example based on  [MSNBC.com Anonymous Web Data Data Set](https://archive.ics.uci.edu/ml/datasets/MSNBC.com+Anonymous+Web+Data)
 
The new example demonstrate how to use the `CategoryEncoder` and `HTMPrediction` model to predict the next page the user is more likely to click based on the page category.

@rhyolight Please review
CC: @subutai


rhyolight:
looking

subutai:
> it does not use the SP, which seems to make it an atypical use case

With category inputs, the SP does not add anything. It will actually make the results a little worse.

examples/prediction/category_prediction/category.py
```diff
@@ -0,0 +1,188 @@
...
+    "sensorParams": {
+      "verbosity": 0,
+      "encoders": {
+
+        # Category encoder used for web page categories
+        "category": {
+          "fieldname": u"category",
+          "name": u"category",
+          "type": "CategoryEncoder",
+          "categoryList": CATEGORIES.keys(),
```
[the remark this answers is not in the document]
lscheinkman: The dataset uses integers for categories. If I use the category names here then i would have to convert from integer to name before feeding to the model.  
See https://archive.ics.uci.edu/ml/machine-learning-databases/msnbc-mld/msnbc.data.html

subutai:
So I think no-SP should be typical for category examples, as long as there are no scalar inputs. Maybe we should explain that?

examples/prediction/category_prediction/category.py
```diff
@@ -0,0 +1,188 @@
...
+          "w": 21
+        },
+      },
+      "sensorAutoReset": None,
+    },
+
+    # No need to use SP for this dataset
+    "spEnable": False,
+    "spParams": {
+    },
```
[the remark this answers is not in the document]
lscheinkman: The category representation is already stable and there is no need for SP in this situation. I will add a better comment explaining that.

examples/prediction/category_prediction/category.py
```diff
@@ -0,0 +1,188 @@
...
+        # Learn distinct user sessions
+        if line in sessions:
+          continue
+
+        # Select user sessions with 2 or more pages
+        pages = line.split()
+        if len(pages) < 2:
+          continue
+
+        # Train model using the first 100 users sessions
```
subutai: Why only a hundred users? How many are there?

examples/prediction/category_prediction/category.py
```diff
@@ -0,0 +1,188 @@
...
+          break
+
+        # Simple progress status
+        sys.stdout.write("\r{} Sessions".format(count))
+        sys.stdout.flush()
+
+        # Keep track of learned user session
+        sessions.add(line)
+
+        # Present each user session 10 times
```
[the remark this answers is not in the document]
subutai: Yes, why are we feeding each user 10 times? Can we just feed in more users, but once per user?

examples/prediction/category_prediction/category.py
subutai: Do we need to even specify the categories ahead of time? I thought the encoder can add in categories incrementally.

examples/prediction/category_prediction/category.py
```diff
@@ -0,0 +1,188 @@
...
+
+# List of categories in the dataset
+CATEGORIES = {
+  "1": "frontpage", "2": "news", "3": "tech", "4": "local", "5": "opinion",
+  "6": "on-air", "7": "misc", "8": "weather", "9": "msn-news", "10": "health",
+  "11": "living", "12": "business", "13": "msn-sports", "14": "sports",
+  "15": "summary", "16": "bbs", "17": "travel"
+}
+
+# HTM Prediction Model Parameters
```
subutai: How did you get these parameters?

examples/prediction/category_prediction/category.py
```diff
@@ -0,0 +1,188 @@
...
+}
+
+
+
+def main():
+  # Create HTM Prediction model and enable inference on the "category" field
+  model = ModelFactory.create(MODEL_PARAMS)
+  model.enableInference({"predictedField": "category"})
+
+  # Assumes data file is in same directory as the script
```
subutai: I'm assuming the user has to download the dataset ahead of time. If the file is not present we should let them know in an error message.

lscheinkman: The datafile is in the repo. No need to download.

rhyolight:
I would be ok with this example if we: 
- exposes the category encoding more clearly (use strings, not ints, show the `"string" ==> 0000011111111000000000000` somewhere during the run so users can see how their strings are being encoded
- explained why the SP is not necessary
- removed the artificial looping over users during training (I always tell people we don't need training data, and this makes it seem like we do)
- explained how we came up with the model parameters and any category-specific parameters settings

lscheinkman:
@rhyolight Ready for review

examples/prediction/category_prediction/webdata.py
```diff
@@ -0,0 +1,298 @@
...
+
+
+
+def main():
+  # Create HTM prediction model and enable inference on the page field
+  model = ModelFactory.create(MODEL_PARAMS)
+  model.enableInference({"predictedField": "page"})
+
+  # Use the model encoder to display the encoded SDRs the model will learn
+  sdr_table = PrettyTable(field_names=["Page Category", "Encoded SDR"],
```
rhyolight: `"Encoded SDR (on bit indices)"`

(Just being blatantly obvious.)

examples/prediction/category_prediction/README.md
```diff
@@ -1,21 +1,121 @@
...
+```  
+python webdata.py
+```
 
-Use this simply to reset the generated files. Strongly recommended if you intend to use this code multiple times!
+### Sample output:
+```text
+The following table shows the encoded SDRs for every page category in the dataset
++---------------+-----------------------------------------------------------------------------+
+| Page Category | Encoded SDR                                                                 |
```
rhyolight: `"Encoded SDR (on bit indices)"`

examples/prediction/category_prediction/webdata.py
```diff
@@ -0,0 +1,298 @@
...
+    "spParams": {},
+
+    "tmEnable": True,
+    "tmParams": TM_PARAMS,
+
+    "clParams": CL_PARAMS,
+  },
+}
+
+TRAINING_RECORDS = 100000
```
rhyolight: I recommend taking this down to 10,000. When I ran with 10K I got:

```
 - Prediction Accuracy: 0.614173228346
 - Accuracy Predicting Top 3 Pages: 0.825196850394
```

This is close enough, I think, and it makes the example run so much faster. You might note that accuracy will increase if more records are used for learning. 

Also, instead of `TRAINING_RECORDS` I would change to this var name to something regarding learning. I don't like using the term "training" with HTM. Maybe `START_LEARNING_AFTER` or similar?

examples/prediction/category_prediction/README.md
```diff
@@ -1,21 +1,121 @@
...
+|    news   | ('news', 'frontpage', 'on-air', 'local', 'tech', 'weather', 'business', 'sports')       |
++-----------+-----------------------------------------------------------------------------------------+
+|  weather  | ('weather', 'frontpage', 'news', 'on-air', 'msn-news', 'msn-sports', 'local', 'sports') |
++-----------+-----------------------------------------------------------------------------------------+
+| frontpage | ('frontpage', 'news', 'business', 'sports', 'on-air', 'tech', 'local', 'misc')          |
++-----------+-----------------------------------------------------------------------------------------+
+
+Compute prediction accuracy by checking if the next page in the sequence is within the predicted pages calculated by the model:
+ - Prediction Accuracy: 0.698660714286
+ - Accuracy Predicting Top 3 Pages: 0.814732142857
```
rhyolight: Update these two lines per my comment below.

lscheinkman:
@rhyolight One more review

PR #3817, opened 2018-03-09
Optimization to spatial_pooler.py initialization. Fixes #1385

### LINKED ISSUE #1385: The _mapPotential method and the SparseBinaryArray, return a dense array

ctrl-z-9000-times:
I rewrote some of the SpatialPooler's initialization code to run faster by using Numpy instead of python "for" loops. 

This is related to issue #1385 which complains that the SP.init is slow. Issue #1385 is about using dense arrays to represent the potential synapses for each column.  The dense boolean arrays are iterated over by the function _initPermanence.  Issue #1385 calls for replacing these dense boolean arrays with a sparse array of indices.  Instead of directly fixing issue #1385 I instead made the surrounding code more capable of working with dense arrays.  This is accomplished primarily through the use of the function numpy.nonzero which converts the dense boolean arrays into sparse index arrays.

### DIFF
--- a/src/nupic/algorithms/spatial_pooler.py
+++ b/src/nupic/algorithms/spatial_pooler.py
@@ -1263,40 +1263,6 @@ class SpatialPooler(Serializable):
     self._connectedCounts[columnIndex] = newConnected.size
 
 
-  def _initPermConnected(self):
-    """
-    Returns a randomly generated permanence value for a synapses that is
-    initialized in a connected state. The basic idea here is to initialize
-    permanence values very close to synPermConnected so that a small number of
-    learning steps could make it disconnected or connected.
-
-    Note: experimentation was done a long time ago on the best way to initialize
-    permanence values, but the history for this particular scheme has been lost.
-    """
-    p = self._synPermConnected + (
-        self._synPermMax - self._synPermConnected)*self._random.getReal64()
-
-    # Ensure we don't have too much unnecessary precision. A full 64 bits of
-    # precision causes numerical stability issues across platforms and across
-    # implementations
-    p = int(p*100000) / 100000.0
-    return p
-
-
-  def _initPermNonConnected(self):
-    """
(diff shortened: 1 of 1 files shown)


rhyolight:
@ctrl-z-9000-times This repo is currently in maintenance mode and we are now accepting new features or enhancements. (All our engineers are currently doing research.) We use the C++ Spatial Pooler when we need speed. The python SP is really there for easy prototyping and readability. 

You might find that the [community fork](https://discourse.numenta.org/c/nupic/community-fork) will accept your contributions, however.

PR #3818, opened 2018-03-12
Fix softmax overflow

suzusuzu:
Please look this softmax formula.

If x_i is too big, it occur overflow. It is necessary to subtract constant m. Sometimes people set m to be the maximum of x.


suzusuzu:
Sorry, I didn't know the Contributor License.
Now, I have signed the Contributor License.

PR #3819, opened 2018-03-13
Fix softmax overflow

suzusuzu:
Please look this softmax formula.

If x_i is too big, it occur overflow. It is necessary to subtract constant m. Sometimes people set m to be the maximum of x.


rhyolight:
@scottpurdy Can you take a look at this?

scottpurdy:
This seems like a valid concern but I don't think the proposed change is right (wouldn't it make all the values negative?). I think the better approach is to:

1. Submit a test case that shows the overflow scenario
2. Address the issue when the weight matrix is updated to make sure that the weights stay in a reasonable range.

PR #3844, opened 2018-06-12
Added community link policy docs from forum

rhyolight:
See https://discourse.numenta.org/t/documentation-on-dimensions-links-linkpolicy-and-splitter-maps/3964/11?u=rhyolight


rhyolight:
I'm just waiting for the author of these new docs to give us the 👍 to use his words.

dkeeney:
Yes, please use my words if you would like.  I did not know you needed me to say so.
One change however....The Input buffer size in the destination's Spec does not seem to be used as I said.  The Output buffer sizes are checked but it does not appear that the Input buffer size in the Spec is used.

rhyolight:
@dkeeney Do you mind pointing to some exact changes to make because I don't understand the topic like you do, and I'm afraid I'm going to get it wrong.

dkeeney:
Matt

Change:
*Destination Input Buffer width:* The input buffer width can also be set in
the destination region’s Input Spec using the ‘count’ field. If the ‘count’
is 0 (and it most often is) then it uses the width of the source output
buffer to which it is connected.

to:
*Destination Input Buffer width:* The input buffer width is set using the
width of the source output buffer to which it is connected.  If more than
one link terminate on the same Input buffer, the input buffer width is set
using the sum of the widths of all source output buffers that are connected.

Does that help?
Dave
========

rhyolight:
@dkeeney Yup, perfect, thanks.

PR #3847, opened 2018-06-18
Close input source after swarm.

### LINKED ISSUE #3846: File descriptor leaking issue
When `swarm.OFPModelRunner.run()` finishes the loop and does `self._finalize()`, it doesn't release the file descriptor that it acquires at the beginning of loop, which caused file descriptor leaking issue.  It will be reported as `Not enough memory` error on Linux if running with large enough set of data and many enough processes. The processes will be killed consequently.

It is fixed by adding 
`  if self._inputSource:
        self._inputSource.close()
`
in `self._finalize`.

rhyolight:
Fixes #3846

### DIFF
--- a/src/nupic/swarming/ModelRunner.py
+++ b/src/nupic/swarming/ModelRunner.py
@@ -420,6 +420,11 @@ class OPFModelRunner(object):
     if self._predictionLogger:
       self._predictionLogger.close()
 
+    # =========================================================================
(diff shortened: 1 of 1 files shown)


rhyolight:
The CircleCI build keeps failing on MySQL errors. I don't think this is a real problem. I'm going to merge this.

PR #3880, opened 2019-08-14
Fix destruction of synapses when destroying segment

dan-mk:
When the **predictedSegmentDecrement** TM parameter is set to some value greater than 0, I was getting this error:

Captura de tela de 2019-08-07 02-42-00

Using the destroySynapse function prevents the error in my tests.

### DIFF
--- a/src/nupic/algorithms/connections.py
+++ b/src/nupic/algorithms/connections.py
@@ -290,9 +290,9 @@ class Connections(Serializable):
     :param segment: (:class:`Segment`) representing the segment to be destroyed.
     """
     # Remove the synapses from all data structures outside this Segment.
-    for synapse in segment._synapses:
-      self._removeSynapseFromPresynapticMap(synapse)
     self._numSynapses -= len(segment._synapses)
+    for synapse in list(segment._synapses):
+      self.destroySynapse(synapse)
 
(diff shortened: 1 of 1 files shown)


rhyolight:
@dan-mk Can you update the test suite to fail when `predictedSegmentDecrement` is non-zero? It would be best to expose this as an error in an existing test suite.

rhyolight:
BTW @dan-mk, this repository is in maintenance mode. There is no guarantee this contribution will be merged anytime soon. See the README for details. The currently active HTM C++/Python project is https://github.com/htm-community/htm.core

dan-mk:
@rhyolight I'm not sure about the right way to contribute to the test suite, so I'm attaching a modified "hello_tm" that should show the error before 60k iterations. If it does not, I've probably done something wrong

Captura de tela de 2019-08-21 03-30-30

[hello_tm.zip](https://github.com/numenta/nupic/files/3523838/hello_tm.zip)

I understand this repository is in maintenance mode and that it's not a critical bug (since it seems like I'm the first one having problems with this), so no problem if it's not merged

I'm also aware about the community repository, but I had the same error with its code, except that it gave me a segfault which is not as easy to debug haha

